PR #3, opened 2016-07-25
op-guide: Add deploy TiDB with docker document in Chinese.

shenli:
@iamxy @siddontang @qiuyesuifeng PTAL


shenli:
Move docs in Chines to pingcap/docs-cn repo.
See https://github.com/pingcap/docs-cn/pull/1

PR #64, opened 2016-11-15
Add the guide for history read

queenypingcap:


queenypingcap:
@coocood @shenli @siddontang @c4pt0r  PTAL.

c4pt0r:
LGTM

op-guide/history-read.md
```diff
@@ -0,0 +1,155 @@
...
++-----------------------+----------------------------+
+7 rows in set (0.00 sec)
+```
+
+Pay special attention to the following two rows:
+- `tikv_gc_life_time`: This row is to configure the retention time of the history version and its default value is 10m. You can use SQL statements to configure it. For example, if you want all the data within one day to be readable, set this row to 24h.
+
+  **Note:** If your data is updated very frequently, the following issues might occur if the value of the `tikv_gc_life_time` is set to be too large like in days or months：
+    - The more versions of the data, the more disk storage is occupied.
+    - A large amount of the history versions might slow down the query, especially the range queries like `select count(\*) from t`.
```
coocood: count(*)

op-guide/history-read.md
```diff
@@ -0,0 +1,155 @@
...
+| c    |
++------+
+|    1 |
+|    2 |
+|    3 |
++------+
+3 rows in set (0.00 sec)
+```
+
+7\. Set the  `tidb_snapshot` variable to be "" (Null) and you can read the data from the latest version:
```
coocood: (Null) -> (empty string)

op-guide/history-read.md
```diff
@@ -0,0 +1,155 @@
...
+- The variable is valid in the `Session` scope.
+- Its value can be modified using the `Set` statement. 
+- The data type for the variable is text. 
+- The variable is to record time in the following format: “2016-10-08 16:45:26.999”. Generally, the time can be set to seconds like in “2016-10-08 16:45:26”. 
+- When the variable is set, TiDB creates a Snapshot using its value as the timestamp, just for the data structure and there is no any overhead. After that, all the `Select` operations will read data from this Snapshot.
+
+**Note:** Because the timestamp in TiDB transactions is allocated by Placement Driver (PD), the version of the stored data is also marked based on the timestamp allocated by PD. When a Snapshot is created, the version number is based on the value of the `tidb_snapshot` variable. If there is a large difference between the local time of the TiDB server and the PD server, use the time of the PD server.
+
+After reading data from history versions, you can read data from the latest version by ending the current Session or using the `Set` statement to set the value of the `tidb_snapshot` variable to "" (Null).
```
coocood: (Null) -> (empty string)

op-guide/history-read.md
queenypingcap: "\" is used as the escape character, or "*" will be recognized as a markdown tag and won't be displayed correctly.

coocood:
LGTM

shenli:
LGTM

PR #71, opened 2016-11-29
Remove the reason about unsupported features

queenypingcap:
The previous reason is: 
because they cannot be implemented well in distributed scenario.
This is not entirely true. We are planning to support some of these features in the future. Therefore, changing the statement.
@shenli @ngaut PTAL.


qiuyesuifeng:
LGTM

shenli:
LGTM

PR #113, opened 2017-05-23
build_rocksdb.sh: check ldconfig 

yuanwhy:
1. Fix  error `./bin/tikv-server: error while loading shared libraries: librocksdb.so.5.2: cannot open shared object file: No such file or directory
`

     Actually,  rocksdb lib path is `/usr/local/lib`  ( it can be found in [rocksdb-readme](https://github.com/facebook/rocksdb/blob/master/INSTALL.md)  ) ,
     but  `/usr/local/lib`  is not default lib path  for command `ldconfig` . reference [ldconfig-manpage](http://man7.org/linux/man-pages/man8/ldconfig.8.html)

   So, we should check first , and then add `/usr/local/lib`  to `/etc/ld.so.conf` if necessary.
   

  2. Directory `deps` is temp directory ,  let's delete it

### DIFF
--- a/scripts/build_rocksdb.sh
+++ b/scripts/build_rocksdb.sh
@@ -31,6 +31,7 @@ function get_linux_platform {
 
 
 function ldconfig_for_rocksdb {
+    ldconfig
     if [ $(${SUDO} ldconfig -p | grep librocksdb | wc -l) -eq 0 ]; then
         ${SUDO} echo "/usr/local/lib" >> /etc/ld.so.conf
         ${SUDO} ldconfig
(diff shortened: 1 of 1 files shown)


yuanwhy:
PTAL @siddontang

andelf:
Changing important system config in such a build script is very unprofessional.
consider using LD_LIBRARY_PATH.

Never mind, this PR solves. 😆

andelf:
set LD_LIBRARY_PATH for current user may be acceptable.
It's a customized build process, so add `LD_LIBRARY_PATH` to start script is also acceptable.
For testing build, http://nixos.org/patchelf.html may help.
@jomenxiao @iamxy

andelf:
Since the recommend way of getting tikv is downloading from official website.
This PR will be closed. @yuanwhy

yuanwhy:
@andelf ok

PR #136, opened 2017-07-09
Should be availability?

huachaohuang:


huachaohuang:
@shenli How do you think?

PR #149, opened 2017-08-07
Add PROXY Protocol support command line parameter document.

blacktear23:
Add PROXY Protocol support command line parameter document.

### DIFF
--- a/op-guide/configuration.md
+++ b/op-guide/configuration.md
@@ -99,6 +99,11 @@ The default TiDB ports are 4000 for client requests and 10080 for status report.
 + default: "" (empty string)
 + The value can be IP address (192.168.1.50) or CIDR (192.168.1.0/24), if require more than one address use `,` to split. `*` means any IP addresses. Leaving it empty disable PROXY Protocol.
 
+### --proxy-protocol-header-timeout
++ PROXY Protocol header read timeout.
++ default: 5 (seconds)
++ The value set timeout for PROXY protocol header read. Unit is second. You should not set this value to 0.
+
 ## Placement Driver (PD)
 
 ### -L
--- a/op-guide/configuration.md
+++ b/op-guide/configuration.md
@@ -95,14 +95,14 @@ The default TiDB ports are 4000 for client requests and 10080 for status report.
 + The value can be (true) or (false). (true) is to enable join without any equal conditions. (false) is to disable it.
 
 ### --proxy-protocol-networks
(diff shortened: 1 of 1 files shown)


queenypingcap:
@shenli @hanfei1991  PTAL.

op-guide/configuration.md
```diff
@@ -94,6 +94,16 @@ The default TiDB ports are 4000 for client requests and 10080 for status report.
 + default: true
 + The value can be (true) or (false). (true) is to enable join without any equal conditions. (false) is to disable it.
 
+### --proxy-protocol-networks
++ PROXY Protocol allowed proxy server IP addresses or CIDRs (split by `,`).
```
queenypingcap: This sentence is ambiguous, does this mean:
The allowed proxy server IP addresses or CIDRs (split by `,`) that is allowed by the PROXY Protocol?

op-guide/configuration.md
```diff
@@ -94,6 +94,16 @@ The default TiDB ports are 4000 for client requests and 10080 for status report.
...
++ default: "" (empty string)
++ The value can be IP address (192.168.1.50) or CIDR (192.168.1.0/24), if require more than one address use `,` to split. `*` means any IP addresses. Leaving it empty disable PROXY Protocol.
```
queenypingcap: Please change it to:
The value can be an IP address (192.168.1.50) or CIDR (192.168.1.0/24). If more than one addresses are required, use `,` to separate. `*` means any IP addresses. Leaving it empty disables PROXY Protocol.

op-guide/configuration.md
```diff
@@ -94,6 +94,16 @@ The default TiDB ports are 4000 for client requests and 10080 for status report.
...
+
+### --proxy-protocol-header-timeout
++ PROXY Protocol header read timeout.
++ default: 5 (seconds)
++ The value set timeout for PROXY protocol header read. Unit is second. You should not set this value to 0.
```
queenypingcap: Can we change it to: The value sets the timeout for the PROXY protocol header read. The unit is second. You cannot set this value to 0.

blacktear23: Actuarially the timeout can be set to 0, and this sometimes may cause strange issue, and most time it can be work correctly. So `should not` is more fit here.

blacktear23:
@queenypingcap PR #3757 is not merged, so those parameter is not work now.

PR #153, opened 2017-08-09
reference configuration instead of list them all

BusyJay:
So we don't have to update both places when configuration is updated.


shenli:
LGTM

queenypingcap:
LGTM

PR #157, opened 2017-08-15
Update the sample table that cannot be migrated

queenypingcap:
Per https://github.com/pingcap/docs-cn/pull/224


queenypingcap:
@buggithubs @sunhao2017 PTAL.

PR #322, opened 2018-01-10
Add transaction isolation levels

lilin90:
Via https://github.com/pingcap/docs-cn/pull/470 and https://github.com/pingcap/docs-cn/pull/490.

### DIFF
--- a/sql/transaction-isolation.md
+++ b/sql/transaction-isolation.md
@@ -1,0 +1,75 @@
+---
+title: TiDB Transaction Isolation Levels
+category: user guide
+---
+
+# TiDB Transaction Isolation Levels
+
+Transaction isolation is one of the foundations of database transaction processing. Isolation is the I in the acronym ACID (Atomicity, Consistency, Isolation, Durability), which represents the isolation property of database transactions.
+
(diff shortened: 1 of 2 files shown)


sql/transaction-isolation.md
```diff
@@ -0,0 +1,75 @@
...
+
+Use the following command to set the level of transaction isolation:
+
+```
+SET SESSION TRANSACTION ISOLATION LEVEL [read committed|repeatable read]
+```
+
+## Repeatable Read
+
+Repeatable Read is the default transaction isolation level in TiDB. The Repeatable Read isolation level only sees data committed before the transaction began, and it never sees either uncommitted data or changes committed during transaction execution by concurrent transactions. However, the transaction statement does see the effects of previous updates executed within its own transaction, even though they are not yet committed.
```
CaitinChen: Repeatable Read is the default transaction isolation level in TiDB. The Repeatable Read isolation level only sees data committed before the transaction "begins"

lilin90:
@wentaoxu @CaitinChen PTAL. Please approve and add "LGTM" if no other problems. Thanks!

PR #387, opened 2018-03-22
Fix relative links in op-guide

ryanpbrewster:
I think the migration in https://github.com/pingcap/docs/pull/347 broke a few links. On master they currently 404.


lilin90:
Thanks for your contribution! 👍

PR #402, opened 2018-04-04
sql: add tidb memory control guide

CaitinChen:
via https://github.com/pingcap/docs-cn/pull/649
@lilin90  @zz-jason  PTAL


sql/tidb-memory-control
```diff
@@ -0,0 +1,44 @@
...
+```
+
+- If the configuration item above uses "log", when the memory usage of a single SQL query exceeds the threshold value which is controlled by the `tidb_mem_quota_query` variable, TiDB prints an entry of log. Then the SQL query continues to be executed. If OOM occurs, you can find the corresponding SQL query in the log.
+- If the configuration item above uses "cancel", when the memory usage of a single SQL query exceeds the threshold value, TiDB stops executing the SQL query immediately and returns an error to the client. The error information shows the memory usage of the execution operators that occupy ssmuch memory in detail.
+
+## How to configure the threshold value of memory usage
+
+You can control the memory usage of a query through the following session variables. Generally, you only need to configure `tidb_mem_quota_query`. Other variables are advanced items which most users do not need to care about.
+
+| Variable Name | Description | Unit | Default Value |
```
zz-jason: maybe we can refer to the code comments https://github.com/pingcap/tidb/blob/master/sessionctx/variable/tidb_vars.go#L116 ?
```go
    // "tidb_mem_quota_query":              control the memory quota of a query.
    // "tidb_mem_quota_hashjoin":           control the memory quota of "HashJoinExec".
    // "tidb_mem_quota_mergejoin":          control the memory quota of "MergeJoinExec".
    // "tidb_mem_quota_sort":               control the memory quota of "SortExec".
    // "tidb_mem_quota_topn":               control the memory quota of "TopNExec".
    // "tidb_mem_quota_indexlookupreader":  control the memory quota of "IndexLookUpExecutor".
    // "tidb_mem_quota_indexlookupjoin":    control the memory quota of "IndexLookUpJoin".
    // "tidb_mem_quota_nestedloopapply":    control the memory quota of "NestedLoopApplyExec".
```

CaitinChen:
@zz-jason

sql/tidb-memory-control.md
```diff
@@ -0,0 +1,44 @@
...
+
+```
+# Valid options: ["log", "cancel"]
+oom-action = "log"
+```
+
+- If the configuration item above uses "log", when the memory quota of a single SQL query exceeds the threshold value which is controlled by the `tidb_mem_quota_query` variable, TiDB prints an entry of log. Then the SQL query continues to be executed. If OOM occurs, you can find the corresponding SQL query in the log.
+- If the configuration item above uses "cancel", when the memory quota of a single SQL query exceeds the threshold value, TiDB stops executing the SQL query immediately and returns an error to the client. The error information shows the memory quota of the execution operators that occupy ssmuch memory in detail.
+
+## How to configure the threshold value of memory quota
```
lilin90: -> Configure the memory quota of a query

CaitinChen:
@lilin90

PR #406, opened 2018-04-08
removed the useless words 'ascending order'

elitecodegroovy:
The correct statement should like this "All the internal indexes are stored in ascending order". So , the original statement should remove the useless words.


lilin90:
Thanks for your contribution!

PR #419, opened 2018-04-12
sql: add STRAIGHT_JOIN and update comment syntax

lilin90:
@winoros @CaitinChen PTAL 
Via: https://github.com/pingcap/docs-cn/pull/663

### DIFF
--- a/sql/comment-syntax.md
+++ b/sql/comment-syntax.md
@@ -82,7 +82,7 @@ In TiDB, you can also use another version:
 SELECT STRAIGHT_JOIN col1 FROM table1,table2 WHERE ...
 ```
 
-If the server version number is specified in the comment, for example, `/*!50110 KEY_BLOCK_SIZE=1024 */`, in MySQL it means that the contents in this comment is processed only when the MySQL version is or higher than 5.1.10. But in TiDB, the version number does not work and all contents in the comment are processed.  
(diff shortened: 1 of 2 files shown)


sql/comment-syntax.md
```diff
@@ -93,25 +93,9 @@ SELECT /*+ hint */ FROM ...;
...
-  
-- TIDB_INLJ(t1, t2)
-
-  ```
-  SELECT /*+ TIDB_INLJ(t1, t2) */ * from t1，t2 where t1.id = t2.id
-  ```
-  
-  The Hint optimizer uses the Index Nested Loop Join algorithm. This algorithm is faster in some scenarios and consumes less system resources, while it may be slower in some other scenarios and consumes more system resources. For the scenarios that have a small result set (less than 10,000 lines) after the filtration of `WHERE` condition, you can try to use it. The parameter in `TIDB_INLJ()` is the candidate table of the driving table (outer table) when the query plan is created. In other words, `TIDB_INLJ(t1)` only uses `t1` as the driving table to create the query plan.
-  
+For details about the optimizer that TiDB supports, see [Optimizer hint](tidb-specific.md#optimizer-hint).
```
zz-jason: s/optimizer/optimizer hints/

CaitinChen:
LGTM

lilin90:
@zz-jason PTAL

zz-jason:
LGTM

PR #439, opened 2018-04-25
tools: add TiKV Control User Guide

lilin90:
@hicqu @CaitinChen PTAL
Via: https://github.com/pingcap/docs-cn/pull/691

### DIFF
--- a/tools/tikv-control.md
+++ b/tools/tikv-control.md
@@ -11,7 +11,7 @@ TiKV Control (`tikv-ctl`) is a command line tool of TiKV, used to manage the clu
 
 `tikv-ctl` provides two operation modes:
 
-- Remote mode: accept the service address of TiKV through the `--host` flag
+- Remote mode: use the `--host` option to accept the service address of TiKV as the argument
 
     For this mode, if SSL is started in TiKV, `tikv-ctl` also needs to specify the related certificate file. For example:
 
@@ -19,7 +19,7 @@ TiKV Control (`tikv-ctl`) is a command line tool of TiKV, used to manage the clu
     $ tikv-ctl --ca-path ca.pem --cert-path client.pem --key-path client-key.pem --host 127.0.0.1:21060 <subcommands>
     ```
 
-- Local mode: use the `--db` flag to specify the local TiKV data directory path
+- Local mode: use the `--db` option to specify the local TiKV data directory path
 
 Unless otherwise noted, all commands supports both the remote mode and the local mode.
 
@@ -101,7 +101,7 @@ To print the value of a key, use the `print` command.
 
 ### Compact data manually
 
-Use the `compact` command to manually compact TiKV data. If you specify the `--from` and `--to` options, then the flag is also in the form of escaped raw key. You can use the `--db` flag to specify the RocksDB that you need to compact. The optional flag values are `kv` and `raft`.
+Use the `compact` command to manually compact TiKV data. If you specify the `--from` and `--to` options, then their flags are also in the form of escaped raw key. You can use the `--db` option to specify the RocksDB that you need to compact. The optional values are `kv` and `raft`.
 
 ```bash
 $ tikv-ctl --db /path/to/tikv/db compact -d kv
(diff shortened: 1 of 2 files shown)


tools/tikv-control.md
```diff
@@ -0,0 +1,164 @@
...
+$ tikv-ctl --db /path/to/tikv/db size -r 2
+region id: 2
+cf default region size: 799.703 MB
+cf write region size: 41.250 MB
+cf lock region size: 27616
+```
+
+### Scan to view MVCC of a specific range
+
+The `--from` and `--to` options of the `scan` command accepts two escaped forms of raw key, and uses the `--show-cf` flag to specify the column families that you need to view.
```
CaitinChen: The `--from` and `--to` options of the `scan` command **accept** two escaped forms of raw key, and **use** the `--show-cf` flag to specify the column families that you need to view.

tools/tikv-control.md
```diff
@@ -0,0 +1,164 @@
...
+Use the `compact` command to manually compact TiKV data. If you specify the `--from` and `--to` options, then the flag is also in the form of escaped raw key. You can use the `--db` flag to specify the RocksDB that you need to compact. The optional flag values are `kv` and `raft`.
+
+```bash
+$ tikv-ctl --db /path/to/tikv/db compact -d kv
+success!
+```
+
+### Set a Region to tombstone
+
+The `tombstone` command is usually used in circumstances where the sync-log is not enabled, and some write data of the Raft state machine is lost caused by power down.
```
CaitinChen: The `tombstone` command is usually used in circumstances where the sync-log is not enabled, and **some data** **written in** the Raft state machine is lost caused by power down.

tools/tikv-control.md
```diff
@@ -0,0 +1,164 @@
...
+Use the `consistency-check` command to execute a consistency check among replicas in the corresponding Raft of a specific Region. If the check fails, TiKV itself panics. If the TiKV instance specified by `--host` is not the Region leader, an error is reported.
+
+```bash
+$ tikv-ctl --host 127.0.0.1:21060 consistency-check -r 2
+success!
+$ tikv-ctl --host 127.0.0.1:21061 consistency-check -r 2
+DebugClient::check_region_consistency: RpcFailure(RpcStatus { status: Unknown, details: Some("StringError(\"Leader is on store 1\")") })
+```
+
+> **Note:**
```
CaitinChen: I think maybe there is something wrong with the format since the color of some words and punctuation marks is red or blue.

lilin90:
@hicqu PTAL

tools/tikv-control.md
```diff
@@ -0,0 +1,164 @@
...
+
+TiKV Control (`tikv-ctl`) is a command line tool of TiKV, used to manage the cluster. When you compile TiKV, the `tikv-ctl` command is also compiled at the same time. If the cluster is deployed using Ansible, the binary file also exist in the corresponding `tidb-ansible/resources/bin` directory.
+
+## General options
+
+`tikv-ctl` provides two operation modes:
+
+- Remote mode: use the `--host` option to accept the service address of TiKV as the argument
+
+    For this mode, if SSL is started in TiKV, `tikv-ctl` also needs to specify the related certificate file. For example:
```
hicqu: Should be `SSL is enabled in TiKV`. The text in docs-cn is also wrong. I will fix the last one.

tools/tikv-control.md
```diff
@@ -0,0 +1,164 @@
...
+$ tikv-ctl --host 127.0.0.1:21061 consistency-check -r 2
+DebugClient::check_region_consistency: RpcFailure(RpcStatus { status: Unknown, details: Some("StringError(\"Leader is on store 1\")") })
+```
+
+> **Note:**
+>
+> - This command only supports the remote mode.
+> - Even if this command returns `success!`, you need to check whether TiKV panics. This is because this command is only a proposal that requests a consistency check for the leader, and you cannot know from the client whether the whole check process is successful or not.
+
+### Print the Regions where the Raft state machine reports an error
```
hicqu: `reports an error` should be `corrupt` or something similar.

hicqu:
rest LGTM.

PR #444, opened 2018-04-26
benchmark: add sysbench 2.0 vs. 1.0

CaitinChen:
via: https://github.com/pingcap/docs-cn/pull/699
@lilin90 @coocood PTAL


coocood:
LGTM

PR #468, opened 2018-05-16
Update the confusing statements

queenypingcap:


op-guide/ansible-deployment.md
```diff
@@ -318,6 +318,8 @@ To deploy TiDB using a normal user account, take the following steps:
 
 ## Test the cluster
 
+Note: Because TiDB is compatible with MySQL, you can use MySQL client to connect to TiDB directly. Before you start TiDB, make sure to install MySQL client first.
```
Hoverbear: Installing before starting TiDB is not necessary. I think the last sentence can be removed.

op-guide/binary-deployment.md
```diff
@@ -132,7 +132,9 @@ The key's randomart image is:
 
 TiDB provides the official binary installation package that supports Linux. For the operating system, it is recommended to use Redhat 7.3+, CentOS 7.3+ and higher versions.
 
-### Operating system: Linux (Redhat 7+, CentOS 7+)
+> Note: Because TiDB is compatible with MySQL, you can use MySQL client to connect to TiDB directly. Before you start TiDB, make sure to install MySQL client first.
```
Hoverbear: Installing before starting TiDB is not necessary. I think the last sentence can be removed.

op-guide/ansible-deployment.md
queenypingcap: If users don’t have MySQL client installed, how can they test their connection to TiDB? I believe TiDB doesn’t have a client yet...

Hoverbear: They need a MySQL client, yes! But they don't necessarily need one before installing TiDB. I think maybe `Because TiDB is compatible with MySQL, you must use MySQL client to connect to TiDB directly.` is appropriate?

PR #486, opened 2018-06-06
tikv, readme: add "Install and Deploy TiKV Using Docker"

lilin90:
@siddontang @queenypingcap PTAL

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@
   + Install and Deploy TiKV
     - [Prerequisites](op-guide/recommendation.md)
     - [Install and Deploy TiKV Using Docker Compose](tikv/deploy-tikv-docker-compose.md)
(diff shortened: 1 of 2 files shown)


gregwebs:
Do you mean service files for running with docker? I can create some if we don't have them yet.
That does point to a problem with the provided docker commands though. One should either be running them in systemd or if doing just docker adding the `--restart always` flag.

Hoverbear:
@gregwebs I'm more referring to the service files that a host VM might use to manage the services.

tennix:
@Hoverbear Lots of arguments needs to be configured in systemd service files. I don't know if it's appropriate to provide systemd service files

Hoverbear:
@tennix Okay. :+1:

PR #494, opened 2018-06-11
tools: add checking Region properties to TiKV Control

lilin90:
@huachaohuang @breeswish @CaitinChen PTAL
Via: https://github.com/pingcap/docs-cn/pull/754


breeswish:
LGTM. BTW @huachaohuang I think it would be better to change our all existing "local" into "offline mode"?

huachaohuang:
@breeswish I don't have a strong opinion about that, maybe you can discuss it on our tikv channel.

PR #502, opened 2018-06-14
FAQ: add processing time of DDL

lilin90:
@zimulala @CaitinChen PTAL
Via: https://github.com/pingcap/docs-cn/pull/752

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -388,13 +388,13 @@ The lease parameter (`--lease=60`) is set from the command line when starting a 
 
 #### What is the processing time of a DDL operation?
 
-Generally, the processing time of a DDL statement can be divided into three conditions:
+The processing time is different for different scenarios. Generally, you can consider the following three scenarios:
 
 1. The `Add Index` operation with a relatively small number of rows in the corresponding data table: about 3s
 2. The `Add Index` operation with a relatively large number of rows in the corresponding data table: the processing time depends on the specific number of rows and the QPS at that time (the `Add Index` operation has a lower priority than ordinary SQL operations)
 3. Other DDL operations: about 1s
 
-Besides, if the TiDB server that receives the DDL request is same TiDB server that the DDL owner is in, the first and third conditions above might cost dozens to hundreds of milliseconds.
+Besides, if the TiDB server instance that receives the DDL request is same TiDB server instance that the DDL owner is in, the first and third scenarios above might cost dozens to hundreds of milliseconds.
 
 #### Why it is very slow to run DDL statements sometimes?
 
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -386,13 +386,23 @@ The offline node usually indicates the TiKV node. You can determine whether the
 
 The lease parameter (`--lease=60`) is set from the command line when starting a TiDB server. The value of the lease parameter impacts the Database Schema Changes (DDL) speed of the current session. In the testing environments, you can set the value to 1s for to speed up the testing cycle. But in the production environments, it is recommended to set the value to minutes (for example, 60) to ensure the DDL safety.
 
+#### What is the processing time of a DDL operation?
+
+Generally, the processing time of a DDL statement can be divided into three conditions:
```
shenli: "divided into" is not appropriate.

shenli: In this answer, it means we should consider the following three scenarios. The processing time is different for each scenario.

FAQ.md
```diff
@@ -386,13 +386,23 @@ The offline node usually indicates the TiKV node. You can determine whether the
...
+
+1. The `Add Index` operation with a relatively small number of rows in the corresponding data table: about 3s
+2. The `Add Index` operation with a relatively large number of rows in the corresponding data table: the processing time depends on the specific number of rows and the QPS at that time (the `Add Index` operation has a lower priority than ordinary SQL operations)
+3. Other DDL operations: about 1s
+
+Besides, if the TiDB server that receives the DDL request is same TiDB server that the DDL owner is in, the first and third conditions above might cost dozens to hundreds of milliseconds.
```
shenli: TiDB server instance

FAQ.md
```diff
@@ -386,13 +386,23 @@ The offline node usually indicates the TiKV node. You can determine whether the
...
 - After you start the cluster successfully, the first DDL operation may take a longer time to run, usually around 30s. This is because the TiDB cluster is electing the leader that processes DDL statements.
-- In rolling updates or shutdown updates, the processing time of DDL statements in the first ten minutes after starting TiDB is affected by the server stop sequence (stopping PD -> TiDB), and the condition where TiDB does not clean up the registration data in time because TiDB is stopped using the `kill -9` command. When you run DDL statements during this period, for the state change of each DDL, you need to wait for 2 * lease (lease = 10s).
+- The processing time of DDL statements in the first ten minutes after starting TiDB is affected by two conditions: 1) TiDB cannot communicate with PD as usual when you are stopping TiDB (including the case of power failure); 2) TiDB fails to clean up the registration data from PD in time because TiDB is stopped by the `kill -9` command. If you run DDL statements during this period, for the state change of each DDL, you need to wait for 2 * lease (lease = 45s).
```
shenli: The processing time of DDL statements in the first ten minutes after starting TiDB would be much longer than the normal case if you meet the following conditions:

FAQ.md
```diff
@@ -386,13 +386,23 @@ The offline node usually indicates the TiKV node. You can determine whether the
...
 
+#### What is the processing time of a DDL operation?
+
+The processing time is different for different scenarios. Generally, you can consider the following three scenarios:
+
+1. The `Add Index` operation with a relatively small number of rows in the corresponding data table: about 3s
+2. The `Add Index` operation with a relatively large number of rows in the corresponding data table: the processing time depends on the specific number of rows and the QPS at that time (the `Add Index` operation has a lower priority than ordinary SQL operations)
+3. Other DDL operations: about 1s
+
+Besides, if the TiDB server instance that receives the DDL request is same TiDB server instance that the DDL owner is in, the first and third scenarios above might cost dozens to hundreds of milliseconds.
```
CaitinChen: Besides, if the TiDB server instance that receives the DDL request is **the** same TiDB server instance that the DDL owner is in, the first and third scenarios above might cost dozens to hundreds of milliseconds.

PR #510, opened 2018-06-25
op-guide: ansible deployment add dev_mode variable note

LinuxGit:


LinuxGit:
@lilin90 You could add it in  https://github.com/pingcap/docs/pull/510.

PR #539, opened 2018-07-14
correct gc duration error message

zhexuany:
"GC too early" is outdated. Filing this PR to correct it. @coocood @zz-jason PTAL


coocood:
LGTM

zz-jason:
cc @lilin90 @CaitinChen

PR #541, opened 2018-07-16
tools: add dynamically modifying RocksDB configuration

CaitinChen:
Via https://github.com/pingcap/docs-cn/pull/807

@lilin90  @Connor1996  PTAL. Thanks.

### DIFF
--- a/tools/tikv-control.md
+++ b/tools/tikv-control.md
@@ -179,3 +179,22 @@ If the command is successfully executed, it prints the above information. If the
     ```bash
     $ tikv-ctl --host 127.0.0.1:20160 region-properties -r 2
     ```
+
+### Modify the RocksDB configuration of TiKV dynamically
+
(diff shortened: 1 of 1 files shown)


tools/tikv-control.md
```diff
@@ -179,3 +179,22 @@ If the command is successfully executed, it prints the above information. If the
     ```bash
     $ tikv-ctl --host 127.0.0.1:20160 region-properties -r 2
     ```
+
+### Modify the RocksDB configuration of TiKV dynamically
+
+The `modify-tikv-config` command is used to dynamically modify the configuration arguments. Currently it only supports dynamically modifying RocksDB related arguments.
```
lilin90: - For the first sentence, this way sounds better and more user-centered: **You can use the** `modify-tikv-config` command **to** ...
- When we put "currently" at the beginning of a sentence, we usually add a comma after it and omit comma if it is in the middle.

CaitinChen:
@lilin90 Thanks. PTAL again.

PR #546, opened 2018-07-19
op-guide: add accessing Spark with Python or R

lilin90:
@tennix @CaitinChen PTAL
Via: https://github.com/pingcap/tidb-docker-compose/pull/27


tennix:
LGTM

PR #561, opened 2018-08-05
x509 certification error

ceohockey60:
x509 certification error prevents k8s operator tutorial from working. Also made minor typo and grammar edits.


gregwebs:
Thanks, will look at certs now.

PR #581, opened 2018-08-22
tools: add two commands in tikv-ctl

lilin90:
@Connor1996 @hicqu PTAL

### DIFF
--- a/tools/tikv-control.md
+++ b/tools/tikv-control.md
@@ -237,7 +237,7 @@ success!
 > **Note:**
 > 
 > - This command only supports the local mode. It prints `success!` when successfully run.
-> - You must run this command for all stores that have specified Regions' peers. If `--regions` is not set, all Regions are involved, and you need to run this command for all stores.
+> - You must run this command for all stores where specified Regions' peers locate. If `--regions` is not set, all Regions are involved, and you need to run this command for all stores.
 
 ### Recover from MVCC data corruption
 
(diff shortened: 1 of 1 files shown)


tools/tikv-control.md
```diff
@@ -103,6 +105,23 @@ key: zmDB:29\000\000\377\000\374\000\000\000\000\000\000\377\000H\000\000\000\00
 
 In this command, the key is also the escaped form of raw key.
 
+### Recover from MVCC data corruption
```
Connor1996: I prefer moving this with  `Force Region to recover the service from failure of multiple replicas` to the bottom

lilin90:
@Connor1996 PTAL again. Thanks!

tools/tikv-control.md
```diff
@@ -231,3 +222,36 @@ success!
...
+
+```bash
+$ tikv-ctl --db /path/to/tikv/db unsafe-recover remove-fail-stores --stores 3 --regions 1001,1002
+success!
+```
+
+> **Note:**
+> 
+> - This command only supports the local mode. It prints `success!` when successfully run.
+> - You must run this command for all stores that specify Regions' peers. If `--regions` is not set,all Regions are involved, and you need to run this command for all stores.
```
Connor1996: please add a space after`is not set,`

tools/tikv-control.md
```diff
@@ -231,3 +222,36 @@ success!
...
+```bash
+$ tikv-ctl --db /path/to/tikv/db recover-mvcc --regions 1001,1002 --pd 127.0.0.1:2379
+success!
+```
+
+> **Note**:
+> 
+> - This command only supports the local mode. It prints `success!` when successfully run.
+> - The argument of the `--pd/-p` option specifies the PD endpoints without the `http` prefix. Specifying the PD endpoints is to query whether the specified `region_id` is validated or not.
+> - You need to run this command for all stores that specify Regions' peers.
```
Connor1996: I don't quite get the meaning of `all stores that specify Regions' peers.`
IMO, it should be `all stores that have specified...` or `all stores where specified Regions' peers locate`

PR #584, opened 2018-08-23
tikv: complete the command to modify values.yaml

zheng1:


lilin90:
@zheng1 Thanks for your contribution. @tennix PTAL

tikv/deploy-tikv-docker-compose.md
```diff
@@ -44,7 +44,7 @@ Make sure you have installed the following items on your machine:
 
     ```bash
     cd tidb-docker-compose/compose
-    networkMode: host
+    sed -i "s/^networkMode: bridge/networkMode: host/g" values.yaml
```
tennix: This only modifies `networkMode`, but this step we should modify `networkMode` and comment the TiDB section out. So perhaps using following command would be better:
```
cd tidb-docker-compose
vim compose/vlaues.yaml
```
The description in this step is already clear to users how to modify values.yaml using vim.

lilin90:
@zheng1 Any update?

Issue #590, opened 2018-08-28
Differences in defaults between TiDB and MySQL 5.7


morgo:
Is it possible to create a manual page (or alternatively append to [this manual page](https://github.com/pingcap/docs/blob/master/sql/mysql-compatibility.md)) differences in defaults between TiDB and MySQL?  My short list:

* TiDB defaults to `utf8` (MySQL 5.7 is latin1 - this changes in 8.0 to utf8).
* TiDB defaults and only supports `lower_case_table_names=2`.
* TiDB's default collation is binary.

lilin90:
Related content is added to docs in this PR: https://github.com/pingcap/docs/pull/591

PR #591, opened 2018-08-29
sql: add default differences between TiDB and MySQL

lilin90:
Via this issue: https://github.com/pingcap/docs/issues/590

@tiancaiamao As for the character set, we say TiDB support UTF-8 or `utf8mb4`? Which one is more appropriate? Related links:

- https://github.com/pingcap/docs/blob/master/sql/character-set-configuration.md
- https://github.com/pingcap/docs/blob/master/sql/system-database.md#character_sets-table

### DIFF
--- a/sql/character-set-configuration.md
+++ b/sql/character-set-configuration.md
@@ -6,6 +6,6 @@ category: user guide
 
 # Character Set Configuration
 
-Currently, TiDB does not support configuring the character set. The default character set is UTF-8.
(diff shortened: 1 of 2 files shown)


lilin90:
@morgo PTAL

morgo:
LGTM.

@tiancaiamao See also https://mysqlserverteam.com/mysql-8-0-when-to-use-utf8mb3-over-utf8mb4/ (it is a regrettable legacy).

tiancaiamao:
LGTM

PR #604, opened 2018-09-07
sql: Added incompatibilities with MySQL

morgo:
events, sys do not exist
pfs exists but returns empty.

### DIFF
--- a/sql/mysql-compatibility.md
+++ b/sql/mysql-compatibility.md
@@ -150,7 +150,7 @@ The output of the query execution plan returned from the `EXPLAIN` command diffe
     - The default value in TiDB is `utf8` which is equivalent to `utf8mb4` in MySQL.
     - The default value in MySQL 5.7 is `latin1`, but changes in `utf8mb4` in MySQL 8.0.
 - Default collation: `latin1_swedish_ci` in MySQL 5.7, while `binary` in TiDB.
-- Default SQL Mode:
+- Default SQL mode:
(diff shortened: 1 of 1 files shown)


queenypingcap:
@shenli  PTAL.

morgo:
I've added more content for beginners (explaining that ENGINE=myisam is not creating a MyISAM table).

morgo:
Added large transactions and EXPLAIN (differs substantially).
Added Optimizer Trace (TiDB has query tracing, which is different from optimizer tracing).

sql/mysql-compatibility.md
```diff
@@ -83,12 +90,20 @@ TiDB implements the asynchronous schema changes algorithm in F1. The Data Manipu
...
 TiDB implements an optimistic transaction model. Unlike MySQL, which uses row-level locking to avoid write conflict, in TiDB, the write conflict is checked only in the `commit` process during the execution of the statements like `Update`, `Insert`, `Delete`, and so on.
 
 **Note:** On the business side, remember to check the returned results of `commit` because even there is no error in the execution, there might be errors in the `commit` process.
 
+### Large transactions
+
+Due to the distributed, 2-phase commit requirement of TiDB, large transactions that modify data can be particularly problematic. TiDB intentionally sets some limits on transaction sizes to reduce this impact:
+
+* Each Key-Value entry is no more than 6MB
+* The total number of Key-Value entries is no more than 300,000 rows
```
shenli: I suggest removing the `rows`. Because this limitation only considers kv pairs, not sql rows.

PR #605, opened 2018-09-07
sql: Improve Default Differences

morgo:
Improved clarity:

The current text says the default is utf8mb4 in TiDB, and in MySQL 8.0 becomes utf8.  It becomes utf8mb4 in 8.0, and technically is default utf8 in TiDB - it just works the same as utf8mb4.

Added Additional Difference:

The default SQL mode in TiDB differs from MySQL.


queenypingcap:
@shenli PTAL.

morgo:
I am going to close this pull requests in https://github.com/pingcap/docs/pull/604

The two PRs are starting to overlap as I document other differences.

PR #607, opened 2018-09-08
sql: Remove unsupported functions

morgo:
The following functions are not supported in TiDB, but are in the documentation.

I suggest removing them and add them back once implemented.


zz-jason:
@XuHuaiyu PTAL

PR #608, opened 2018-09-08
sql: YEAR(2) is not supported, M is display width

morgo:
Fixing minor errors.

Integer(M) is display width.

YEAR(2) is not supported in MySQL or TiDB:
```
MySQL [test]> create table tt1 ( a int not null primary key, b year(2));
ERROR 1105 (HY000): line 1 column 57 near ""Supports only YEAR or YEAR(4) column. (total length 57)
```


queenypingcap:
@lilin90 please update the same content in docs-cn.

PR #609, opened 2018-09-08
sql: Clarify default character-set is utf8

morgo:
Improved clarity of changing character set, linked back to mysql compatibility/differences in defaults so users can discover other differences.


shenli:
LGTM

queenypingcap:
@lilin90 Please add the content to the same docs in docs-cn. Thanks

PR #610, opened 2018-09-08
sql: Default tidb_ddl_reorg_priority is slow

morgo:
The default value for this variable is `PRIORITY_LOW` (not normal).  It can be verified with:

```
mysql> show global variables like 'tidb_ddl_reorg_priority';
+-------------------------+--------------+
| Variable_name           | Value        |
+-------------------------+--------------+
| tidb_ddl_reorg_priority | PRIORITY_LOW |
+-------------------------+--------------+
1 row in set (0.00 sec)
```

This PR also fixes minor formatting, where options are at the incorrect header level.


shenli:
LGTM

PR #611, opened 2018-09-08
*: split overview.md into sub documents

morgo:
The main navigation has anchor links into `overview.md` for introduction and architecture (but not features).    I think it makes sense to add an anchor link for 'Features', but in this PR I propose an alternative -- which is to break this document in three:

- `overview.md`
- `architecture.md`
- `features.md`

The advantage to breaking the document up, is that we can track their individual popularity in Google Analytics etc.  I suspect that `architecture.md` will be very popular :-)

Note: I've changed heading levels as part of this, but no other changes.  I will suggest other small changes after this merges.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,9 @@
 ## Documentation List
 
 + About TiDB
-  - [TiDB Introduction](overview.md#tidb-introduction)
-  - [TiDB Architecture](overview.md#tidb-architecture)
+  - [TiDB Introduction](overview.md)
+  - [TiDB Architecture](architecture.md)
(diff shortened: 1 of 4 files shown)


shenli:
LGTM
But I suggest adding more content in the features.md.

shenli:
@coocood PTAL

morgo:
@shenli I agree :-)  I will suggest more after this PR is merged (consider this refactor first).

shenli:
@morgo Yes, I agree with you. So you have already got my `LGTM`.

PR #614, opened 2018-09-11
sql: remove documentation about read committed

lamxTyler:
The read committed isolation level is removed in https://github.com/pingcap/tidb/pull/7280, we should remove it in the docs.

Via: https://github.com/pingcap/docs-cn/pull/871


shenli:
@lamxTyler Please add a description for this PR.

PR #620, opened 2018-09-13
Fix broken manual page

morgo:
The manual page is currently broken:
https://pingcap.com/docs/sql/mysql-compatibility/

It broke after push of https://github.com/pingcap/docs/pull/604 although it looks like it was a change in the markdown parser, since this line was not modified.

I investigated it, and think this might fix it - but it's difficult to tell since GitHub's markdown parser works fine.


gregwebs:
I doubt this could make it worse.

gregwebs:
Seems fixed now.

PR #627, opened 2018-09-26
pd-ctl: add topsize command and change the default value of limit to 16

liukun4515:
[docs-cn](https://github.com/pingcap/docs-cn/pull/894)


liukun4515:
PTAL @disksing

PR #632, opened 2018-09-26
sql: Add recent JSON functionality

morgo:


morgo:
Added json_length() - recently merged.

PR #634, opened 2018-09-26
Update tune-tikv.md

zhexuany:
https://github.com/tikv/tikv/blob/master/src/raftstore/store/config.rs#L60

### DIFF
--- a/op-guide/tune-tikv.md
+++ b/op-guide/tune-tikv.md
@@ -89,7 +89,7 @@ sync-log = true
 region-max-size = "384MB"
 # The threshold value of Region split
 region-split-size = "256MB"
-# When the data size in a Region is larger than the threshold value, TiKV checks whether this Region needs split.
+# When the data size change in a Region is larger than the threshold value, TiKV checks whether this Region needs split.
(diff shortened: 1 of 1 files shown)


zhexuany:
@CaitinChen  @breeswish PTAL

op-guide/tune-tikv.md
```diff
@@ -89,7 +89,7 @@ sync-log = true
 region-max-size = "384MB"
 # The threshold value of Region split
 region-split-size = "256MB"
-# When the data size in a Region is larger than the threshold value, TiKV checks whether this Region needs split.
+# When the data size change in a Region is larger than the threshold value, TiKV checks whether this Region needs split.
```
breeswish: I don't think the semantic is correct here. It isn't "change is larger...".

CaitinChen: How about
When the data size in a Region is changed to be larger than the threshold value, TiKV checks whether this Region needs split.

breeswish: Oh, my fault. This comment is for the config below. So it is indeed "change is larger".

PR #644, opened 2018-10-02
op-guide, tools: Recommend mydumper from enterprise tools

### LINKED ISSUE #630: Clarify which mydumper to use
The manual currently includes a link to `mydumper`, but doesn't mention we have a fork of it as well:
https://pingcap.com/docs/op-guide/backup-restore/

The upstream mydumper is actually unsafe for TiDB.  We have a PR request to fix it, and assuming it's merged we should update the advice to say the minimum version to use.

Or alternatively, we should point to our official `mydumper`

morgo:
The `mydumper` upstream is unsafe, since it relies on `ftwrl` instead of `tidb_snapshot`.  Soon the server will produce an error when attempting this unsafe statement: https://github.com/pingcap/tidb/pull/7712

The `mydumper` currently in enterprise tools is unsafe by default, but can be made safe by setting the `tidb_snapshot` as an option.

In https://github.com/morgo/mydumper/tree/tidb-head I have re-implimented the `tidb_snapshot` option to aut-set, so backups are safe by default.  It is also based on upstream's master branch, so it includes the last couple of years of bug fixes and support for generated columns.

`tidb-head` also includes all features that were in the master branch of `pingcap/mydumper`, so I am hoping we can switch it to be `pingcap/master`.  At which point the docs page in this pull request will be 100% accurate.

Fixes #630

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -691,7 +691,7 @@ TiDB is not suitable for tables of small size (such as below ten million level),
 
 #### How to back up data in TiDB?
 
-Currently, the preferred way of backing up is using the [TiDB fork of mydumper](tools/mydumper.md).Although the official MySQL tool `mysqldump` is also supported in TiDB to back up and restore data, its performance is poorer than [`mydumper`](tools/mydumper.md)/[`loader`](tools/loader.md) and it needs much more time to back up and restore large volumes of data. Therefore, it is not recommended to use `mysqldump`. 
+Currently, the preferred method for backup is using the [PingCAP fork of mydumper](tools/mydumper.md). Although the official MySQL tool `mysqldump` is also supported in TiDB to back up and restore data, its performance is poorer than [`mydumper`](tools/mydumper.md)/[`loader`](tools/loader.md) and it needs much more time to back up and restore large volumes of data.
 
 Keep the size of the data file exported from `mydumper` as small as possible. It is recommended to keep the size within 64M. You can set value of the `-F` parameter to 64.
 
--- a/FAQ.md
+++ b/FAQ.md
@@ -691,23 +691,11 @@ TiDB is not suitable for tables of small size (such as below ten million level),
 
 #### How to back up data in TiDB?
 
-Currently, the major way of backing up data in TiDB is using `mydumper`. For details, see [mydumper repository](https://github.com/maxbube/mydumper). Although the official MySQL tool `mysqldump` is also supported in TiDB to back up and restore data, its performance is poorer than `mydumper`/`loader` and it needs much more time to back up and restore large volumes of data. Therefore, it is not recommended to use `mysqldump`. 
+Currently, the preferred way of backing up is using the [TiDB fork of mydumper](tools/mydumper.md).Although the official MySQL tool `mysqldump` is also supported in TiDB to back up and restore data, its performance is poorer than [`mydumper`](tools/mydumper.md)/[`loader`](tools/loader.md) and it needs much more time to back up and restore large volumes of data. Therefore, it is not recommended to use `mysqldump`. 
 
 Keep the size of the data file exported from `mydumper` as small as possible. It is recommended to keep the size within 64M. You can set value of the `-F` parameter to 64.
 
 You can edit the `t` parameter of `loader` based on the number of TiKV instances and load status. For example, in scenarios of three TiKV instances, you can set its value to `3 * (1 ～ n)`. When the TiKV load is very high and `backoffer.maxSleep 15000ms is exceeded` displays a lot in `loader` and TiDB logs, you can adjust the parameter to a smaller value. When the TiKV load is not very high, you can adjust the parameter to a larger value accordingly.
-
-## Migrate the data and traffic
-
-### Full data export and import
-
-#### Mydumper
-
-See the [mydumper repository](https://github.com/maxbube/mydumper).
-
-#### Loader
-
-See [Loader Instructions](tools/loader.md).
 
 #### How to migrate an application running on MySQL to TiDB?
 
--- a/tools/mydumper.md
+++ b/tools/mydumper.md
@@ -1,7 +1,7 @@
 ---
 title: mydumper Instructions 
 summary: Use mydumper to export data from TiDB.
-category: advanced
+category: tools
 ---
 
 # mydumper Instructions
--- a/tools/mydumper.md
+++ b/tools/mydumper.md
@@ -8,7 +8,7 @@ category: tools
 
 ## What is mydumper?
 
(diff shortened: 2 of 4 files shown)


morgo:
@gregwebs I've added a new page under tools. PTAL.

gregwebs:
fantastic!

morgo:
I've found and fixed a further reference in FAQ.

morgo:
@kennytm @GregoryIan PTAL

tools/mydumper.md
```diff
@@ -0,0 +1,46 @@
...
++ Include the hidden `_tidb_rowid` column in exports when present
+
++ Allow for `tidb_snapshot` to be configurable (i.e. backup data as it appeared at an earlier point in time)
+
+
+### New parameter description
+
+```
+  -z, --tidb-snapshot: Set the tidb_snapshot to be used for the backup.
+                       Default: NOW()-INTERVAL 1 SECOND.
```
kennytm: This can be any datetime expression (e.g. `-z "'2018-10-10 22:22:22'"`)? Is there any link to the `tidb_snapshot` feature?

tools/mydumper.md
```diff
@@ -0,0 +1,46 @@
...
+
+`mydumper` is a fork of the [mydumper](https://github.com/maxbube/mydumper) project with additional functionality specific to TiDB. It is the recommended method to use for logical backups of TiDB.
+
+[Download the Binary](http://download.pingcap.org/tidb-enterprise-tools-latest-linux-amd64.tar.gz).
+
+## What enhancements does this contain over regular mydumper?
+
++ Use `tidb_snapshot` to provide backup consistency instead of `FLUSH TABLES WITH READ LOCK`
+
++ Include the hidden `_tidb_rowid` column in exports when present
```
kennytm: This affects both the CREATE TABLE schema and INSERT INTO data files?

```sql
CREATE TABLE tbl_name (
    _tidb_rowid BIGINT UNSIGNED AUTO_INCREMENT PRIMARY KEY,  -- there will be this?
    ...
) ...;

INSERT INTO tbl_name VALUES (1, ...), (2, ...), ...;
--                           ^ there will be this?
```

morgo: It is only present in INSERT, not CREATE TABLE.  I will make it clearer.

tools/mydumper.md
morgo: It is documented here: https://pingcap.com/docs/op-guide/history-read/#how-tidb-reads-data-from-history-versions

morgo: I've tried to make it clear.  links can't be used in code blocks, so I've added a link above when describing the feature.

tools/mydumper.md
```diff
@@ -0,0 +1,46 @@
...
+
+`mydumper` is a fork of the [mydumper](https://github.com/maxbube/mydumper) project with additional functionality specific to TiDB. It is the recommended method to use for logical backups of TiDB.
+
+[Download the Binary](http://download.pingcap.org/tidb-enterprise-tools-latest-linux-amd64.tar.gz).
+
+## What enhancements does this contain over regular mydumper?
+
++ Use `tidb_snapshot` to provide backup consistency instead of `FLUSH TABLES WITH READ LOCK`
+
++ Include the hidden `_tidb_rowid` column in `INSERT` statements when present
```
kennytm: I assume the INSERT statements would include the column names? I get a "Column count doesn't match value count" error otherwise. 

```sql
INSERT INTO `table` (_tidb_rowid, col1, col2) VALUES (1, 'x', 'y'), ...;
--                  ^~~~~~~~~~~~~~~~~~~~~~~~~
```

morgo: The relevant line is:
https://github.com/pingcap/mydumper/blob/master/mydumper.c#L3035

If it knows there is a generated col or hidden rowid, it must use the complete insert syntax that specifies column names.  See testcase:

https://gist.github.com/morgo/38b7562fcea4f9839433afd7503d8099

kennytm: I see, thanks :)

tools/mydumper.md
```diff
@@ -0,0 +1,46 @@
...
+# mydumper Instructions
+
+## What is mydumper?
+
+`mydumper` is a fork of the [mydumper](https://github.com/maxbube/mydumper) project with additional functionality specific to TiDB. It is the recommended method to use for logical backups of TiDB.
+
+[Download the Binary](http://download.pingcap.org/tidb-enterprise-tools-latest-linux-amd64.tar.gz).
+
+## What enhancements does this contain over regular mydumper?
+
```
CaitinChen: For the following sentences, we had better use "Uses", "Includes", and "Allows" instead of "Use", "Include" and "Allow", because these sentences are descriptive (describe something) rather than imperative (tell users to do something).
Reference: "These comments should be descriptive ("Opens the file") rather than imperative ("Open the file"); the comment describes the function, it does not tell the function what to do." on https://google.github.io/styleguide/cppguide.html

tools/mydumper.md
```diff
@@ -0,0 +1,46 @@
...
+                       Accepts either a TSO or valid datetime.  For example: -z "2016-10-08 16:45:26"
+```
+
+### Usage example
+
+Command line parameter:
+
+```
+./bin/mydumper -h 127.0.0.1 -u root -P 4000
+```
```
CaitinChen: Please add a line after it.
Note: For a code block enclosed with "```", we should add a line both before and after it.

FAQ.md
```diff
@@ -691,24 +691,12 @@ TiDB is not suitable for tables of small size (such as below ten million level),
...
+Currently, the preferred method for backup is using the [PingCAP fork of mydumper](tools/mydumper.md). Although the official MySQL tool `mysqldump` is also supported in TiDB to back up and restore data, its performance is poorer than [`mydumper`](tools/mydumper.md)/[`loader`](tools/loader.md) and it needs much more time to back up and restore large volumes of data.
 
 Keep the size of the data file exported from `mydumper` as small as possible. It is recommended to keep the size within 64M. You can set value of the `-F` parameter to 64.
 
 You can edit the `t` parameter of `loader` based on the number of TiKV instances and load status. For example, in scenarios of three TiKV instances, you can set its value to `3 * (1 ～ n)`. When the TiKV load is very high and `backoffer.maxSleep 15000ms is exceeded` displays a lot in `loader` and TiDB logs, you can adjust the parameter to a smaller value. When the TiKV load is not very high, you can adjust the parameter to a larger value accordingly.
 
-## Migrate the data and traffic
```
CaitinChen: I think we must keep "## Migrate the data and traffic" and "### Full data export and import". Otherwise, the following will have no corresponding title.

morgo: Sorry!  I misunderstood what this part meant.  I will submit a new PR to restore.

Issue #662, opened 2018-10-10
Add partitioning documentation


morgo:
Ideas:

- Add to mysql compatibility page (only range partitioning supported)
- Add to user-guide
- Add to optimizer somewhere (partition pruning)

lilin90:
The partitioning doc had been added in https://github.com/pingcap/docs/pull/1300, so I'll close this issue. You can also [view this doc on the PingCAP website](https://pingcap.com/docs/stable/reference/sql/partitioning/).

PR #667, opened 2018-10-14
sql: add partitioning to compatibility

morgo:
Relates to https://github.com/pingcap/docs/issues/662

### DIFF
--- a/sql/mysql-compatibility.md
+++ b/sql/mysql-compatibility.md
@@ -142,7 +142,7 @@ Architecturally, TiDB does support a similar storage engine abstraction to MySQL
 
 ### Partitioning
 
-TiDB supports only [`RANGE` Partitioning](https://dev.mysql.com/doc/refman/5.7/en/partitioning-range.html).  Support for `LIST`, `HASH`, `KEY` and subpartitioning is not supported at this time.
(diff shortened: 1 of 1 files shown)


lilin90:
@shenli PTAL

sql/mysql-compatibility.md
```diff
@@ -140,6 +140,10 @@ Create Table: CREATE TABLE `t1` (
 
 Architecturally, TiDB does support a similar storage engine abstraction to MySQL, and user tables are created in the engine specified by the [`--store`](server-command-option.md#--store) option used when you start tidb-server (typically `tikv`).
 
+### Partitioning
+
+TiDB supports only [`RANGE` Partitioning](https://dev.mysql.com/doc/refman/5.7/en/partitioning-range.html). Support for `LIST`, `HASH`, `KEY` and subpartitioning is not available at this time.
```
shenli: We should mention that partitioning is not enabled by default. See https://github.com/pingcap/tidb/blob/master/sessionctx/variable/tidb_vars.go#L110
/cc @tiancaiamao

tiancaiamao: I believe we can enable range partition by default on the master branch within this month.

morgo:
I will close this for now.  We can revisit with hash partitioning (coming soon).

PR #677, opened 2018-10-16
overview, readme: update ETL wording

lilin90:
@CaitinChen PTAL


gregwebs:
I actually prefer the previous version because the claim is weaker: we are just minimizing ETL to the extent that it is possible for a database to do. Whether it is minimal depends on the needs of the organization

* do they do ETL to insert data into the DB?
* do they need to transform their data to be more effective at analysis, or can they take the transactional schema as-is?

queenypingcap:
Ok, minimize it is

PR #684, opened 2018-10-22
op-guide: add document about `compatible-kill-query` in config

tiancaiamao:
It's introduced in https://github.com/pingcap/tidb/pull/7192

@lilin90 @zz-jason

### DIFF
--- a/op-guide/tidb-config-file.md
+++ b/op-guide/tidb-config-file.md
@@ -34,6 +34,12 @@ This document describes the options that are not involved in command line option
 - For details, you can see the [MySQL description](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_lower_case_table_names) of this variable
 - Currently, TiDB only supports setting the value of this option to 2. This means it is case-sensitive when you save a table name, but case-insensitive when you compare table names. The comparison is based on the lower case.
 
+### `compatible-kill-query`
+
++ Set "kill" statement to be MySQL compatible
++ Default: false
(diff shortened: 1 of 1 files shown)


op-guide/tidb-config-file.md
```diff
@@ -34,6 +34,12 @@ This document describes the options that are not involved in command line option
 - For details, you can see the [MySQL description](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_lower_case_table_names) of this variable
 - Currently, TiDB only supports setting the value of this option to 2. This means it is case-sensitive when you save a table name, but case-insensitive when you compare table names. The comparison is based on the lower case.
 
+### `compatible-kill-query`
+
++ Set "kill" statement to be MySQL compatible
```
lilin90: To keep it consistent with other similar items, please update as follows:

-> **To set the `kill`** statement to be MySQL compatible

op-guide/tidb-config-file.md
```diff
@@ -34,6 +34,12 @@ This document describes the options that are not involved in command line option
...
++ Default: false
++ In TiDB, the behavior of "kill xxx" is not compatible with MySQL. Only when you are executing "kill tidb xxx" will a query be killed. If `compatible-kill-query` is set to true, the "kill xxx" will be compatible with MySQL, so no additional "tidb" is need.
```
lilin90:
```suggestion
+ In TiDB, the behavior of "kill xxx" is not compatible with MySQL. A query is killed only when you are executing "kill tidb xxx". If `compatible-kill-query` is set to true, the "kill xxx" is compatible with MySQL, so no additional "tidb" is needed.
```

PR #691, opened 2018-10-26
sql: Improve description of i_s.engines

morgo:
It was improved to no longer show dummy data: https://github.com/pingcap/tidb/pull/7831


lilin90:
@shenli PTAL

PR #693, opened 2018-10-28
tools: update mydumper default tidb_snapshot

morgo:
In https://github.com/pingcap/mydumper/pull/8 our `mydumper` recently changed it's default from NOW()-1 second to the current TSO from `SHOW MASTER STATUS`.  This PR updates the docs.


morgo:
@kennytm PTAL

PR #694, opened 2018-10-29
pd-ctl: update the usage of region key

rleungx:
This PR is based on pingcap/pd#1298.

### DIFF
--- a/tools/pd-control.md
+++ b/tools/pd-control.md
@@ -420,7 +420,7 @@ Usage:
 
 ### `region key [--format=raw|encode] <key>`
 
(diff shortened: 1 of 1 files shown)


tools/pd-control.md
```diff
@@ -418,9 +418,9 @@ Usage:
 }
 ```
 
-### `region key [--format=raw|pb|proto|protobuf] <key>`
+### `region key [--format=raw|encode] <key>`
 
-Use this command to query the region that a specific key resides in. It supports the raw and protobuf formats.
+Use this command to query the region that a specific key resides in. It supports the raw and encoding formats. And it needs to use single quotes around the key when it is in the encoding format.
```
lilin90: it needs to use -> **you need** to use

disksing:
LGTM.

PR #695, opened 2018-10-29
sql: improve privileges docs

morgo:
I noticed a few small errors in the privileges page.  While few users use column-level privileges, I think it is still worth calling out on the compatibility page.

### DIFF
--- a/sql/privilege.md
+++ b/sql/privilege.md
@@ -22,19 +22,19 @@ TiDB user account names consist of a user name and a host name. The account name
 The `CREATE USER` statement creates new MySQL accounts.
 
 ```sql
-create user 'test'@'127.0.0.1' identified by 'xxx';
+CREATE USER 'test'@'127.0.0.1' IDENTIFIED BY 'xxx';
 ```
 
 If the host name is not specified, you can log in from any IP address. If the password is not specified, it is empty by default:
 
 ```sql
-create user 'test';
+CREATE USER 'test';
 ```
 
 Equals:
 
 ```sql
-create user 'test'@'%' identified by '';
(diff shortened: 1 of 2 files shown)


morgo:
PTAL @tiancaiamao

sql/privilege.md
```diff
@@ -289,7 +294,7 @@ When TiDB starts, some privilege-check tables are loaded into memory, and then t
 If an immediate effect is needed when you modify the `grant` table, you can run the following command:
 
 ```sql
-flush privileges
+flush privileges;
```
kennytm: Perhaps use ALL CAPS here

morgo: @lilin90 do we have a style for SQL commands?  I like to write keywords in caps as kenny suggests, but many of our existing examples are lower case.

kennytm: If the style is all-lowercase then the `DROP USER 'test';` statement above should be changed to `drop user 'test';`.

morgo: @lilin90 There are quite a few inconsistencies: `CREATE USER`, `REVOKE` are in caps.  I can go through and change once I have confirmation of the style.

lilin90: @morgo Thanks! Currently, we don't have a strict SQL style. But to improve the consistency, we can now establish one: "Capitalize all SQL keywords and reserved words." Later, I'll update that in docs-cn and tell other members to follow this style.

morgo: @lilin90 thanks, I've updated all keywords to caps.  PTAL

tiancaiamao:
LGTM

PR #711, opened 2018-11-06
tools: add Reparo User Guide

CaitinChen:
Via https://github.com/pingcap/docs-cn/pull/973
@WangXiangUSTC @lilin90 PTAL

### DIFF
--- a/tools/reparo.md
+++ b/tools/reparo.md
@@ -49,7 +49,7 @@ Usage of Reparo:
 ### Description of the configuration file
 
 ```
-# The storage directory for the binlog file in the protobuf format which Drainer outputs
+# The storage directory for the binlog file in the protobuf format that Drainer outputs
 data-dir = "./data.drainer"
 
 # Uses the index file to locate `ts`. Set this parameter if `start-ts` is set. The file 
@@ -110,7 +110,7 @@ password = ""
 
 ### Note
 
-* `data-dir` specifies the directory for the binlog file which Drainer outputs.
+* `data-dir` specifies the directory for the binlog file that Drainer outputs.
 * Both `start-datatime` and `start-tso` are used to specify the time point for starting recovery, but they are different in the time format. If they are not set, the recovery process starts from the earliest binlog file by default.
 * Both `stop-datetime` and `stop-tso` are used to specify the time point for finishing recovery, but they are different in the time format. If they are not set, the recovery process ends up with the last binlog file by default.
 * `dest-type` specifies the destination type. Its value can be "mysql" and "print." 
--- a/tools/reparo.md
+++ b/tools/reparo.md
@@ -1,11 +1,12 @@
 ---
 title: Reparo User Guide
+summary: Learn to use Reparo.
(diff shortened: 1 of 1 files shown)


tools/reparo.md
```diff
@@ -0,0 +1,121 @@
...
+## Reparo usage
+
+### Description of command line parameters
+
+```
+Usage of Reparo:
+-L string
+    The level of the output information of logs
+    Value: "debug"/"info"/"warn"/"error"/"fatal" ("info" by default)
+-V The print version
```
WangXiangUSTC: print the version?

tools/reparo.md
```diff
@@ -0,0 +1,121 @@
...
+# Value: "debug"/"info"/"warn"/"error"/"fatal" ("info" by default)
+log-level = "info"
+
+# Uses `start-datetime` and `stop-datetime` to specify the time range in which 
+# the binlog files are to be recovered.
+# Format: "2006-01-02 15:04:05"
+# start-datetime = ""
+# stop-datetime = ""
+
+# Correspond to `tso` of `start-datetime` and `stop-datetime` respectively.
```
lilin90:
```suggestion
# Correspond to `start-datetime` and `stop-datetime` respectively. 
```

tools/reparo.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+### Note
+
+* `data-dir` specifies the directory for the binlog file which Drainer outputs.
+* Both `start-datatime` and `start-tso` are used to specify the time point for starting recovery, but they are different in the time format. If they are not set, the recovery process starts from the earliest binlog file by default.
+* Both `stop-datetime` and `stop-tso` are used to specify the time point for finishing recovery, but they are different in the time format. If they are not set, the recovery process ends up with the last binlog file by default.
+* `dest-type` specifies the destination type. Its value can be "mysql" and "print." 
+
+    * When it is set to｀mysql｀, the data can be recovered to MySQL or TiDB that uses or is compatible with the MySQL protocol. In this case, you need to specify the database information in [dest-db] of the configuration information.
```
lilin90:
```suggestion
    * When it is set to｀mysql｀, the data can be recovered to MySQL or TiDB that uses or is compatible with the MySQL protocol. In this case, you need to specify the database information in `[dest-db]` of the configuration information.
```

tools/reparo.md
```diff
@@ -0,0 +1,121 @@
...
+    * When it is set to ｀print｀, only the binlog information is printed. It is generally used for debugging and checking the binlog information. In this case, there is no need to specify [dest-db].
```
lilin90:
```suggestion
    * When it is set to ｀print｀, only the binlog information is printed. It is generally used for debugging and checking the binlog information. In this case, there is no need to specify `[dest-db]`.
```

CaitinChen:
@lilin90 Thanks. PTAL again.

CaitinChen:
@lilin90 PTAL again~

PR #714, opened 2018-11-07
sql: Add small transactions to compatibility

morgo:
Add a use-case where TiDB may be slower.

### DIFF
--- a/sql/mysql-compatibility.md
+++ b/sql/mysql-compatibility.md
@@ -107,6 +107,24 @@ Due to the distributed, 2-phase commit requirement of TiDB, large transactions t
 * The total number of Key-Value entries is no more than 300,000
 * The total size of Key-Value entries is no more than 100MB
 
+### Small transactions
+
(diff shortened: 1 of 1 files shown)


morgo:
PTAL @shenli

shenli:
@jackysp @zhangjinpeng1987 PTAL

sql/mysql-compatibility.md
```diff
@@ -107,6 +107,28 @@ Due to the distributed, 2-phase commit requirement of TiDB, large transactions t
 * The total number of Key-Value entries is no more than 300,000
 * The total size of Key-Value entries is no more than 100MB
 
+### Small transactions
+
+Since each transaction in TiDB requires two round trips to the PD leader, small transactions may have higher latencies in TiDB than MySQL. As a hypothetical example, the following query could be improved by moving from `auto_commit` to using a small transaction:
```
kennytm: I find that this paragraph is confusing, as the first sentence claims small transaction is bad, and the second sentence claims small transaction is good. Maybe change to this?

```suggestion
Since each transaction in TiDB requires two round trips to the PD leader, small transactions may have higher latencies in TiDB than MySQL. As a hypothetical example, the following query could be improved by moving from `auto_commit` to using an explicit transaction:
```

PR #725, opened 2018-11-09
op-guide: clarify the location of the newly added node

CaitinChen:
@lilin90 PTAL  Via https://github.com/pingcap/docs-cn/pull/974


CaitinChen:
Via https://github.com/pingcap/docs-cn/pull/974

PR #727, opened 2018-11-11
sql: Include initial link to EKS tutorial

morgo:


morgo:
@YiniXu9506 is there a way to double check the display is correct?

Also: long-term I think we should maybe have each tab be a separate `.md` file?  I can see a use case for direct linking to each of the different platforms (not currently possible)

YiniXu9506:
@morgo Yes, you can check the display by merging your current development branch into the existed branch, which is preview-website. 

Since the file 'kubernetes.md' is a markdown, which only can read markdown template syntax and html, we cannot let it contain other markdown file so far.  But I will keep eyes on this issue and try to find a better way to solve it. 

Now, if you want to add a new tab and content, you have to strictly follow the html, which already exist. PS: the value of attribute 'value' inside element 'input' (` <input id="tabGoogle" type="radio" name="tabs" value="GoogleContent" checked>` ) has to be consistent with the value of attribute 'id' inside element 'section' ( `<section id="GoogleContent">`).

PR #728, opened 2018-11-11
sql: Add SQL mode details

### LINKED ISSUE #606: sql: Document which sql_modes are supported
In https://github.com/pingcap/tidb/issues/2738 it mentions that TiDB should support more sql_modes, but it is not clear which existing modes are supported.

Because TiDB does not produce an error when setting an unsupported sql_mode, it is important that this be documented on `sql/mysql-compatibility.md`

morgo:
Fixes #606.

Depends on `NO_AUTO_CREATE_USER` being merged, but expected soon:
https://github.com/pingcap/tidb/pull/8160

### DIFF
--- a/sql/mysql-compatibility.md
+++ b/sql/mysql-compatibility.md
@@ -168,7 +168,7 @@ Architecturally, TiDB does support a similar storage engine abstraction to MySQL
 
 ### SQL Modes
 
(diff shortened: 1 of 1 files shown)


morgo:
@zz-jason @shenli PTAL

shenli:
LGTM
@jackysp PTAL

PR #729, opened 2018-11-12
sql: replace "executor" with "operator"

CaitinChen:
Since "executor" and "operator" indicate the same thing, we need to keep the term consistent in this doc. zz-jason suggests using "operator" so I replace "executor" with "operator".
@morgo @zz-jason PTAL. Thanks.

### DIFF
--- a/sql/understanding-the-query-execution-plan.md
+++ b/sql/understanding-the-query-execution-plan.md
@@ -72,11 +72,11 @@ In the revisited `EXPLAIN` you can see the count of rows scanned has reduced via
 
 As an extension to `EXPLAIN`, `EXPLAIN ANALYZE` will execute the query and provide additional execution statistics in the `execution info` column as follows:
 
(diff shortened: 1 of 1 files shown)


zz-jason:
LGTM

morgo:
LGTM.

@zz-jason Can we keep referring to the whole execution system as "the executor" (terminology MySQL team uses) and an individual stage as "an operator"?  We need to also be careful because the Kubernetes deploy uses "an operator" too.

zz-jason:
> Can we keep referring to the whole execution system as "the executor"

The execution system contains not just the operator implementations, but also the task scheduler and memory controller, etc. Maybe "the execution engine" is more suitable?

CaitinChen:
@zz-jason @morgo Please help approve this PR so that it can be merged.

lilin90:
Currently, the expression "Execution Engine" is used in our Release Notes. [Ref link](https://github.com/pingcap/docs/blob/master/releases/21rc4.md) @morgo

morgo:
OK, I will use execution engine to refer to operators + memory controller + task scheduler.

PR #731, opened 2018-11-12
tools: add upgrade process and some links

CaitinChen:
Via https://github.com/pingcap/docs-cn/pull/980
@lilin90 @WangXiangUSTC PTAL


CaitinChen:
@lilin90 Thanks. PTAL again. @WangXiangUSTC PTAL

PR #737, opened 2018-11-15
lightning: increase the required version from 2.0.4 to 2.0.9

kennytm:
Recently we fixed a bug in RocksDB which would cause loss of data due to race condition (pingcap/rocksdb#59). So we need to increase the TiKV requirement to 2.0.9.


kennytm:
PTAL @lilin90 
cc @DorianZheng

PR #738, opened 2018-11-15
tools: add DM practice guide

lilin90:
@csuzhangxc @CaitinChen PTAL 
Related docs: 
- https://github.com/pingcap/tidb-tools/blob/docs/docs/dm/zh_CN/get-started.md
- https://github.com/pingcap/tidb-tools/blob/docs/docs/dm/zh_CN/maintenance/directory-structure.md


tools/data-migration-practice.md
```diff
@@ -0,0 +1,177 @@
...
+    dm-worker = "172.16.10.73:10081"
+    ```
+
+## Step 3: Configure the data synchronization task
+
+Assuming that you need to synchronize all the `test_table` table data in the `test_db` database of both the upstream MySQL-1 and MySQL-2 instances, to the downstream `test_table` table in the `test_db` database of TiDB, in the full data plus incremental data mode, you can edit the `task.yaml` task configuration file as follows:
+
+```yaml
+name: "test"                  # The task name. You need to use a different name for each
+# of the multiple tasks that run simultaneously.
```
csuzhangxc: for comments in a code block, do not break the line is better?

csuzhangxc:
rest LGTM

lilin90:
@csuzhangxc The English version of comment is much longer than that of the Chinese version, breaking a comment line is to make it more friendly to website users. So they don't have to scroll horizontally too much to read it. I put some of the comments in a new line for readability or consistency. If you have a better idea, I'll be glad to try it. :)

csuzhangxc:
@lilin90 oh, that's fine. LGTM

PR #769, opened 2018-11-26
sql: update memory control config

lilin90:
@zz-jason @CaitinChen PTAL
Via: https://github.com/pingcap/docs-cn/pull/1009


zz-jason:
LGTM

PR #782, opened 2018-11-29
sql: add two TiDB specific variables

lilin90:
@winkyao @CaitinChen PTAL
Via: https://github.com/pingcap/docs-cn/pull/1004


CaitinChen:
@winkyao PTAL

PR #783, opened 2018-11-29
sql: add concurrent execution of `ADD INDEX` and `GENERAL` 

CaitinChen:
Via https://github.com/pingcap/docs-cn/pull/1021
@lilin90 @zimulala PTAL

### DIFF
--- a/sql/ddl.md
+++ b/sql/ddl.md
@@ -8,6 +8,8 @@ category: user guide
 
 DDL (Data Definition Language) is used to define the database structure or schema, and to manage the database and statements of various objects in the database.
 
+Currently, TiDB has implemented parallel execution of the `ADD INDEX` operation and a general operation (namely a non-`ADD INDEX` DDL operation) across different tables. In this case, two workers handle the `ADD INDEX` operation and a general operation respectively; when the operation requests are for the same table, workers execute these operations in the order of receiving the DDL operation requests. This feature is to guarantee the `ADD INDEX` operation (its execution time is longer that others' in TiDB) will not block other DDL operations.
+
 ## CREATE DATABASE syntax
 
 ```sql
--- a/sql/ddl.md
+++ b/sql/ddl.md
@@ -8,7 +8,7 @@ category: user guide
 
 DDL (Data Definition Language) is used to define the database structure or schema, and to manage the database and statements of various objects in the database.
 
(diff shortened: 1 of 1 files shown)


sql/ddl.md
```diff
@@ -8,6 +8,8 @@ category: user guide
 
 DDL (Data Definition Language) is used to define the database structure or schema, and to manage the database and statements of various objects in the database.
 
+Currently, TiDB has implemented parallel execution of the `ADD INDEX` operation and a `GENERAL` operation (namely a non-`ADD INDEX` DDL operation) across different tables. In this case, two workers handle the `ADD INDEX` operation and a `GENERAL` operation respectively; when the operation requests are for the same table, workers execute these operations in the order of receiving the DDL operation requests. This feature is to guarantee the `ADD INDEX` operation (its execution time is longer that others' in TiDB) will not block other DDL operations.
```
morgo: I think this should be called "concurrent execution".

gregwebs: I agree that concurrent is more accurate.

shenli: Yes

CaitinChen: @morgo @gregwebs @shenli Thanks.

sql/ddl.md
```diff
@@ -8,6 +8,8 @@ category: user guide
 
 DDL (Data Definition Language) is used to define the database structure or schema, and to manage the database and statements of various objects in the database.
 
+Currently, TiDB has implemented concurrent execution of the `ADD INDEX` operation and a `GENERAL` operation (namely a non-`ADD INDEX` DDL operation) across different tables. In this case, two workers handle the `ADD INDEX` operation and a `GENERAL` operation respectively; when the operation requests are for the same table, workers execute these operations in the order of receiving the DDL operation requests. This feature is to guarantee the `ADD INDEX` operation (its execution time is longer that others' in TiDB) will not block other DDL operations.
```
lilin90:
```suggestion
Currently, TiDB has implemented concurrent execution of the `ADD INDEX` operation and the `GENERAL` operation (namely the non-`ADD INDEX` DDL operation) across different tables. In this case, two workers process the `ADD INDEX` operation and a `GENERAL` operation respectively. When the operation requests are on the same table, workers execute these operations in the order of receiving the DDL operation requests. This feature is to guarantee the `ADD INDEX` operation (its execution time is longer than that of other operations in TiDB) does not block other DDL operations.
```

CaitinChen:
@lilin90 PTAL

CaitinChen:
@zimulala PTAL

PR #801, opened 2018-12-05
sql, tools: remove "_tidb_rowid"

lilin90:
@GregoryIan @CaitinChen PTAL

Via: https://github.com/pingcap/docs-cn/pull/1037

### DIFF
--- a/sql/tidb-specific.md
+++ b/sql/tidb-specific.md
@@ -319,32 +319,6 @@ SELECT /*+ TIDB_HJ(t1, t2) */ * from t1, t2 where t1.id = t2.id
 
 This variable is used to remind the optimizer to use the `Hash Join` algorithm. This algorithm executes threads concurrently. It runs faster but takes up more memory.
 
-## _tidb_rowid
-
-This is a hidden column of TiDB, which represents the column name of the implicit ROW ID. It only exists on the tables with non-integer PK or without PK. You can execute the `SELECT`, `INSERT`, `UPDATE` and `DELETE` statements on this column, and the usage of these statements are as follows:
-
-- `SELECT`: `SELECT *, _tidb_rowid from t;`
-- `INSERT`: `INSERT t (c, _tidb_rowid) VALUES (1, 1);`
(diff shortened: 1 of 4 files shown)


sql/tidb-specific.md
```diff
@@ -319,32 +319,6 @@ SELECT /*+ TIDB_HJ(t1, t2) */ * from t1, t2 where t1.id = t2.id
...
-## _tidb_rowid
-
-This is a hidden column of TiDB, which represents the column name of the implicit ROW ID. It only exists on the tables with non-integer PK or without PK. You can execute the `SELECT`, `INSERT`, `UPDATE` and `DELETE` statements on this column, and the usage of these statements are as follows:
-
-- `SELECT`: `SELECT *, _tidb_rowid from t;`
-- `INSERT`: `INSERT t (c, _tidb_rowid) VALUES (1, 1);`
-- `UPDATE`: `UPDATE t SET c = c + 1 WHERE _tidb_rowid = 1;`
-- `DELETE`: `DELETE FROM t WHERE _tidb_rowid = 1;`
-
-## SHARD_ROW_ID_BITS
```
kennytm: I think this option should be kept. It could be reworded to avoid mentioning `_tidb_rowid`. 🤔

GregoryIan:
LGTM

sql/tidb-specific.md
lilin90: @kennytm Could you provide a version for the description?

kennytm: @lilin90 Deleting the first sentence "*You can use this TABLE OPTION to set the bit digit of the number of implicit `_tidb_rowid` shards.*" should be enough. The rest does not talk about `_tidb_rowid` at all.

lilin90: @GregoryIan What do you think about keeping the `SHARD_ROW_ID_BITS` section?

lilin90: > I think this option should be kept. It could be reworded to avoid mentioning `_tidb_rowid`. 🤔

@kennytm Thanks! I've updated them. By the way, do we need to keep `SHARD_ROW_ID_BITS` in 2.0 documentation?

PR #811, opened 2018-12-12
faq: add details of watching DDL job progress

lilin90:
@zimulala @CaitinChen PTAL
Via: https://github.com/pingcap/docs-cn/pull/1038


lilin90:
@zimulala PTAL

gregwebs:
I think @morgo was an original requester of this documentation.

PR #828, opened 2018-12-25
tools: update tikv-ctl unsafe recovery

CaitinChen:
Via https://github.com/pingcap/docs-cn/pull/1060
@lilin90 @hicqu PTAL

### DIFF
--- a/tools/tikv-control.md
+++ b/tools/tikv-control.md
@@ -227,7 +227,7 @@ success!
 
 Use the `unsafe-recover remove-fail-stores` command to remove the failed machines from the peer list of the specified Regions. This command has only one mode "local". Before running this command, you need to stop the service of the target TiKV to release the file lock.
 
-The `-s` option accepts multiple `store_id` separated by comma and uses the `-r` flag to specify involved Regions. To recover the service from failure of multiple replicas for all the Regions in one store, specify `store_id` as `--all-regions`.
+The `-s` option accepts multiple `store_id` separated by comma and uses the `-r` flag to specify involved Regions. To recover the service from the failure of multiple replicas for all the Regions in one store, specify `--all-regions`.
 
 ```bash
 $ tikv-ctl --db /path/to/tikv/db unsafe-recover remove-fail-stores -s 3 -r 1001,1002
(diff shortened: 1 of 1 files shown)


CaitinChen:
@lilin90 Thanks. PTAL again. @hicqu PTAL. Thanks.

tools/tikv-control.md
```diff
@@ -225,19 +225,24 @@ success!
 
 ### Force Region to recover the service from failure of multiple replicas
 
-Use the `unsafe-recover remove-fail-stores` command to remove the failed machines from the peer list of Regions. Then after you restart TiKV, these Regions can continue to provide services using the other healthy replicas. This command is usually used in circumstances where multiple TiKV stores are damaged or deleted.
+Use the `unsafe-recover remove-fail-stores` command to remove the failed machines from the peer list of the specified Regions. This command has only one mode "local". Before running this command, you need to stop the service of the target TiKV to release the file lock.
```
hicqu: I prefer `stop the target TiKV` instead of `stop the service of the target TiKV`.
Seems the Chinese version is also wordy here. What do you think?

tools/tikv-control.md
```diff
@@ -225,19 +225,24 @@ success!
...
 ```
 
+Then after you restart TiKV, these Regions can continue to provide services using the other healthy replicas. This command is usually used in circumstances where multiple TiKV stores are damaged or deleted.
+
 > **Note:**
 > 
 > - This command only supports the local mode. It prints `success!` when successfully run.
-> - You must run this command for all stores where specified Regions' peers locate. If `-r` is not set, all Regions are involved, and you need to run this command for all stores.
+> - Generally, you need to run this command for all stores where the peers of the specified Regions are located.
+> - If you specify `--all-regions`, you need to run this command for all the other healthy stores in the cluster.
```
hicqu: I prefer `you should run` instead of `you need to run`, the first is more sharp, I think.

CaitinChen:
@hicqu Thanks. PTAL again.

hicqu:
LGTM.

PR #832, opened 2018-12-28
scripts: update rustup.sh location

b41sh:
The location of rustup.sh has moved.
Run the following command to install from the new location:
    curl https://sh.rustup.rs -sSf | sh


lilin90:
@b41sh Thanks for your contribution ^_^

PR #835, opened 2019-01-03
Add FLUSH PRIVILEGES hint to user-account-management.md and FAQ.md

lastzero:
It looked like FLUSH PRIVILEGES is not implemented because of pingcap/tidb#674. user-account-management.md also didn't mention it. I've added to all relevant pages.

### DIFF
--- a/sql/user-account-management.md
+++ b/sql/user-account-management.md
@@ -43,6 +43,7 @@ mysql> GRANT ALL PRIVILEGES ON *.* TO 'finley'@'%' WITH GRANT OPTION;
 mysql> CREATE USER 'admin'@'localhost' IDENTIFIED BY 'admin_pass';
 mysql> GRANT RELOAD,PROCESS ON *.* TO 'admin'@'localhost';
 mysql> CREATE USER 'dummy'@'localhost';
+mysql> FLUSH PRIVILEGES;
 ```
 
 To see the privileges for an account, use `SHOW GRANTS`:
@@ -93,3 +94,13 @@ Or:
 ```sql
 ALTER USER 'jeffrey'@'localhost' IDENTIFIED BY 'mypass';
 ```
+
+## Flush privileges 
+
+Please run the following command to apply changed permissions immediately:
+
+```sql
+FLUSH PRIVILEGES;
+```
+
+See [Privilege Management](privilege.md) for details.
--- a/FAQ.md
+++ b/FAQ.md
@@ -104,6 +104,7 @@ The display content of TiDB `show processlist` is almost the same as that of MyS
 #### How to modify the user password and privilege?
 
 To modify the user password in TiDB, it is recommended to use `set password for 'root'@'%' = '0101001';` or `alter`, not `update mysql.user` which might lead to the condition that the password in other nodes is not refreshed timely.
+Run `FLUSH PRIVILEGES` to apply changed permissions immediately.
 
 It is recommended to use the official standard statements when modifying the user password and privilege. For details, see [TiDB user account management](sql/user-account-management.md).
 


morgo:
May I request you make it clearer:
- `FLUSH PRIVILEGES` is only required in the case of updates directly to the tables.
- If you use `GRANT`, `SET PASSWORD`, `DROP USER`, `CREATE USER` etc, `FLUSH PRIVILEGES`  is run automatically.

lastzero:
If you are absolutely sure and it counts as a bug if it doesn't work immediately, I'll invest more time to debug the issue with unit tests 👍

lonng:
@morgo https://github.com/pingcap/tidb/pull/8886

> May I request you make it clearer:
> 
> * `FLUSH PRIVILEGES` is only required in the case of updates directly to the tables.
> * If you use `GRANT`, `SET PASSWORD`, `DROP USER`, `CREATE USER` etc, `FLUSH PRIVILEGES`  is run automatically.

Should execute `FLUSH PRIVILEGES` manually in the current release.

lastzero:
> Should execute `FLUSH PRIVILEGES` manually in the current release. If I am wrong, please point them out.

This is also what your CTO told me. IMHO the docs should mention that, independent of what future releases might do. Does running `FLUSH PRIVILEGE` cause any harm I am not aware of?

morgo:
> If you are absolutely sure and it counts as a bug if it doesn't work immediately, I'll invest more time to debug the issue with unit tests 👍

Yes, I worked on this code recently: https://github.com/pingcap/tidb/pull/8886

When you run `FLUSH PRIVILEGES`, it sends a notification to etcd to update privileges.  The bug I fixed was so that if there is no etcd (i.e. the test suite) it also runs `FLUSH PRIVILEGES` synchronously.

This is MySQL compatible behavior.

lastzero:
Ah, fixed 5 days ago? No wonder it doesn't work for me.

What about users that run older versions?

morgo:
> Ah, fixed 5 days ago? No wonder it doesn't work for me.

I fixed it for the case of the test suite 5 days ago, if you have etcd, the notification channel should still work.  If it doesn't, that is a new bug.

> What about users that run older versions?

The docs are internally versioned (although currently the website only shows 'master' - that is something that is being worked on).  /cc @queenypingcap

lastzero:
We use the local version for our software. I know, special use case. We are a test suite for new ideas.

lastzero:
Also it seems there are ready-to-use Docker images out there, that use the local version. These people will waste a lot of time figuring out what's wrong, like I did. At least half a day.

morgo:
> We use the local version for our software. I know, special use case. We are a test suite for new ideas.

Do you mean tidb w/mocktikv (no tikv or pd)?  If so, that was my motivation for fixing that bug :-)  As well as this one https://github.com/pingcap/tidb/pull/8836.  I would love to chat more if you don't mind me emailing you.

> Also it seems there are ready-to-use Docker images out there, that use the local version. These people will waste a lot of time figuring out what's wrong, like I did. At least half a day.

If it is documented as it is though, that means it is intended behavior.  It is not intended, or MySQL compatible.  We could change the wording and apply to non-master docs, or we could backport the fix to existing GA releases.  I am fine with either approach.

morgo:
> We could change the wording and apply to non-master docs, or we could backport the fix to existing GA releases. I am fine with either approach.

I had a chat with @shenli.  I will prepare a patch to backport to the 2.1 branch.

lastzero:
> I would love to chat more if you don't mind me emailing you.

Sounds good! Let's have a chat, here are my contact details: https://blog.liquidbytes.net/contact/

morgo:
Changes LGTM.  Thanks for editing!

sql/user-account-management.md
```diff
@@ -93,3 +93,13 @@ Or:
 ```sql
 ALTER USER 'jeffrey'@'localhost' IDENTIFIED BY 'mypass';
 ```
+
+## Flush privileges 
+
+If you modified the grant tables directly, please run the following command to apply changes immediately:
```
CaitinChen:
```suggestion
If you modified the grant tables directly, run the following command to apply changes immediately:
```

sql/user-account-management.md
```diff
@@ -93,3 +93,13 @@ Or:
...
+
+```sql
+FLUSH PRIVILEGES;
+```
+
+See [Privilege Management](privilege.md) for details.
```
CaitinChen:
```suggestion
For details, see [Privilege Management](../sql/privilege.md).
```

sql/user-account-management.md
CaitinChen: We usually do not use "please" in technical documents.

CaitinChen:
@lastzero Thanks for your contribution!

sql/user-account-management.md
lastzero: Is `../sql/` a mistake? Both documents are in sql, so ../sql is the same directory.

sql/user-account-management.md
```diff
@@ -93,3 +93,13 @@ Or:
 ```sql
 ALTER USER 'jeffrey'@'localhost' IDENTIFIED BY 'mypass';
 ```
+
+## Flush privileges
```
CaitinChen:
```suggestion
## `FLUSH PRIVILEGES` 
```

sql/user-account-management.md
CaitinChen: @lastzero No, `../sql/` is not a mistake. Your link path is correct on GitHub, but on the PingCAP website it will be a broken link. 
Our link rule: Any relative path that links to another file must start from the level of the root directory.
We updated the link path rule for SEO. Thanks for your cooperation :)

lastzero: Interesting. You use pandoc, right? In my own projects, MkDocs automatically creates the right links in HTML as needed.

CaitinChen: We do use pandoc to convert HTML to PDF, but it has nothing to do with our links in the documents.
We have our own link rule because of the complex hierarchical relationship in our website.

PR #838, opened 2019-01-04
sql: update the type change of `null`

CaitinChen:
Via https://github.com/pingcap/docs-cn/pull/1070
@zimulala @lilin90 PTAL


caocao-dba:
What version of this feature is updated? 2.1.6 is not available yet.

PR #851, opened 2019-01-14
sql: add statement limit

morgo:
I believe the limit is `"stmt-count-limit": 5000` (in configuration).

PTAL @shenli


c4pt0r:
LGTM

PR #877, opened 2019-01-28
releases: pluralize common words

morgo:
It is more common to refer to it as "views", "window functions" etc.


CaitinChen:
@morgo Thanks for your PR, but please note that:
- The release notes is described in the present tense because it is published on the release day.
- The TiDB-Ansible version information should be added in the release notes for the convenience of users.
Later I will file a PR for these issues.

PR #908, opened 2019-02-19
*: add TiDB 2.1.3 sysbench test result

TomShawn:


lilin90:
The test result has been updated a lot and will be soon added in another PR.

PR #917, opened 2019-02-23
sql: update explicit defaults for timestamps

morgo:
Better document a behavior difference from MySQL.


morgo:
@jackysp PTAL

jackysp:
PTAL @CaitinChen

PR #938, opened 2019-03-07
FAQ: update some FAQ items

dcalvin:
updated some FAQ sentences for accuracy

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -646,7 +646,7 @@ WAL belongs to ordered writing, and currently, we do not apply a unique configur
 
 Generally, enabling `sync-log` reduces about 30% of the performance. For write performance when `sync-log` is set to `false`, see [Performance test result for TiDB using Sysbench](benchmark/sysbench.md).
 
-#### Can Raft + multiple replicas in the TiKV architecture achieve absolute data security? Is it necessary to apply the most strict mode (`sync-log = true) to a standalone storage?
+#### Can Raft + multiple replicas in the TiKV architecture achieve absolute data security? Is it necessary to apply the most strict mode (`sync-log = true`) to a standalone storage?
 
 Data is redundantly copied between TiKV nodes using the [Raft consensus algorithm](https://raft.github.io/) to ensure recoverability should a node failure occur. Only when the data has been written into more than 50% of the nodes, will the application return ACK (two out of three nodes). However, theoretically, two nodes might crash. Therefore, for scenarios with a strict requirement on data security, for example, the financial industry, you need to enable the `sync-log` mode.
 
--- a/FAQ.md
+++ b/FAQ.md
@@ -642,13 +642,13 @@ WAL belongs to ordered writing, and currently, we do not apply a unique configur
 - NUMA: no specific suggestion; for memory allocation strategy, you can use `interleave = all`
 - File system: ext4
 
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -642,13 +642,13 @@ WAL belongs to ordered writing, and currently, we do not apply a unique configur
...
 
-#### Can the Raft + multiple replicas in the upper layer implement complete data security? Is it required to apply the most strict mode to standalone storage?
+#### Can Raft + multiple replicas in the TiKV architecture achieve absolute data security? Is it necessary to apply the most strict mode (`sync-log = true) to a standalone storage?
 
-Raft uses strong consistency, and only when the data has been written into more than 50% of the nodes, the application returns ACK (two out of three nodes). In this case, data consistency is guaranteed. However, theoretically, two nodes might crash. Therefore, for scenarios that have a strict requirement on data security, such as scenarios in financial industry, you need to enable the `sync-log`.
+Raft replicates data with strong consistency. Only when the data has been written into more than 50% of the nodes, will the application return ACK (two out of three nodes). In this case, data consistency is guaranteed. However, theoretically, two nodes might crash. Therefore, for scenarios with a strict requirement on data security, for example, the financial industry, you need to enable the `sync-log` mode.
```
dcalvin: "In this case, data consistency is guaranteed."  Does this equal to "security is guaranteed"?

morgo: I think it is important to be careful using the word consistency with raft.  It does not provide the snapshot isolation consistency for example, since it operates at a much lower level.

```suggestion
Data is redundantly copied between TiKV nodes using the [Raft consensus algorithm](https://raft.github.io/) to ensure recoverability should a node failure occur. Only when the data has been written into more than 50% of the nodes, will the application return ACK (two out of three nodes). However, theoretically, two nodes might crash. Therefore, for scenarios with a strict requirement on data security, for example, the financial industry, you need to enable the `sync-log` mode.

As an alternative to using `sync-log`, you may also consider having five nodes instead of three in your Raft group. This would allow for the failure of two nodes, while still providing data safety.
```

FAQ.md
```diff
@@ -642,13 +642,15 @@ WAL belongs to ordered writing, and currently, we do not apply a unique configur
...
 - File system: ext4
 
-#### How is the write performance in the most strict data available mode of `sync-log = true`?
+#### How is the write performance in the most strict data available mode (`sync-log = true`)?
 
-Generally, enabling `sync-log` reduces about 30% of the performance. For the test about `sync-log = false`, see [Performance test result for TiDB using Sysbench](benchmark/sysbench.md).
+Generally, enabling `sync-log` reduces about 30% of the performance. For write performance when `sync-log` is set to `false`, see [Performance test result for TiDB using Sysbench](benchmark/sysbench.md).
 
-#### Can the Raft + multiple replicas in the upper layer implement complete data security? Is it required to apply the most strict mode to standalone storage?
+#### Can Raft + multiple replicas in the TiKV architecture achieve absolute data security? Is it necessary to apply the most strict mode (`sync-log = true) to a standalone storage?
```
kennytm:
```suggestion
#### Can Raft + multiple replicas in the TiKV architecture achieve absolute data security? Is it necessary to apply the most strict mode (`sync-log = true`) to a standalone storage?
```

Missing an end backquote

PR #948, opened 2019-03-12
*: update SQL_NO_CACHE description

lilin90:
@crazycs520 PTAL  Via: https://github.com/pingcap/docs-cn/pull/1138

I updated this description in all docs versions. If it is inappropriate to update it in all versions, please let me know which versions need to be updated. Thanks!


morgo:
LGTM

PR #966, opened 2019-03-21
op-guide, tools/lightning: remove Tab spaces

CaitinChen:
@lilin90 PTAL
@kennytm For the lightning.md file, PTAL

### DIFF
--- a/tools/lightning/csv.md
+++ b/tools/lightning/csv.md
@@ -192,11 +192,11 @@ trim-last-separator = false
 Example content:
 
 ```
-ID	Region	Count
-1	East	32
-2	South	NULL
-3	West	10
-4	North	39
+ID  Region  Count
+1   East    32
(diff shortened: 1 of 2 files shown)


tools/lightning/csv.md
```diff
@@ -192,11 +192,11 @@ trim-last-separator = false
 Example content:
 
 ```
-ID	Region	Count
-1	East	32
-2	South	NULL
-3	West	10
-4	North	39
+ID    Region    Count
```
kennytm: These tabs here are intentional though.

CaitinChen: @kennytm Tabs in our documents cause CI failure, so we should replace them with blanks.

kennytm: But it is not showing "tab" separated values format then.

PR #971, opened 2019-03-23
TOC: Update releases to be per version

morgo:
(There are a few very safe changes in the DITA re-org we can make now.  This pre-work helps reduce the scope of the conversion.)

This PR changes release notes to be organized by release version.


morgo:
@queenypingcap @lilin90 PTAL

shenli:
LGTM

PR #972, opened 2019-03-23
SQL: Add constraints

morgo:

This PR adds some initial documentation for constraints.


gregwebs:
This looks great! My only concern is that for someone that is familiar with the expected behavior, they sometimes have to go through the example to find important information about differences that is listed at the very end.

PR #973, opened 2019-03-23
sql: Organize security topics

morgo:

This PR slightly reorganizes some security topics, adds mysql compatibility.


morgo:
@kolbe @tiancaiamao PTAL

PR #974, opened 2019-03-24
report-issue, contribute, support: new documents

morgo:

This adds new content which will be used in the DITA version of the manual


morgo:
@lilin90 @queenypingcap PTAL

ceohockey60:
LGTM

PR #981, opened 2019-03-27
sql, TOC: Reorganize Compatibility for DITA

morgo:

This PR slightly reorganizes how we talk about compatibility by splitting it into two pages, and creating an explicitly "Transactions" menu item.

It does not change any of the content (other than the general compatibility intro).  The general intro also now links to the sub-topics.


morgo:
@ceohockey60 @kolbe PTAL

ceohockey60:
LGTM

kolbe:
LGTM

PR #982, opened 2019-03-27
toc, sql: remove character set configuration

morgo:
I checked with @jackysp and we are going to say we support utf8 + latin1, so the page is no longer accurate.  Under DITA the other character set page will move under "configuration" and handle the use case of this page.


morgo:
@kolbe @jackysp PTAL

PR #983, opened 2019-03-27
overview: Update overview with key features

morgo:


morgo:
@ceohockey60 @kolbe PTAL

PR #995, opened 2019-03-29
TOC, op-guide: spit configuration by component

morgo:
(This change is used by the DITA toc. I have only made minor edits to the content to accommodate the split.)


morgo:
@ceohockey60 @kolbe PTAL

kolbe:
LGTM

PR #997, opened 2019-03-29
sql: move optimizer hints to separate file

morgo:
In the DITA version, optimizer hints will go under a specific section.


morgo:
@kolbe @ceohockey60 PTAL

kolbe:
LGTM

PR #998, opened 2019-03-29
Copy overview.md to readme.md

morgo:
These two files are duplicated.  In future README.md should have instructions on how to contribute, but for now lets keep them in sync so the content is not outdated.


morgo:
@ceohockey60 @kolbe PTAL

kolbe:
LGTM

PR #1000, opened 2019-03-29
sql: update constraints to show lazy checking occurs on primary key too

morgo:
Important to document it is both types of constraints.


morgo:
@ceohockey60 @kolbe PTAL

ceohockey60:
LGTM

PR #1016, opened 2019-04-03
FAQ: update outdated info

dcalvin:
Updated outdated FAQ info regarding PD and time synchronization error.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -458,7 +458,7 @@ This is because the `--initial-cluster` in the PD startup parameter contains a m
 
 #### What's the maximum tolerance for time synchronization error of PD?
 
-Theoretically, the smaller the time synchronization error, the better. PD can tolerate any synchronization error, but a larger error value means a longer period of service stop during the leader change. If the clock goes back during a leader change, the change process will not proceed until it catches up with the previous leader.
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -458,7 +458,7 @@ This is because the `--initial-cluster` in the PD startup parameter contains a m
 
 #### What's the maximum tolerance for time synchronization error of PD?
 
-Theoretically, the smaller the time synchronization error, the better. PD can tolerate any synchronization error, but a larger error value means a longer period of service stop during the leader change. If the clock goes back during a leader change, the change process will not proceed until it catches up with the previous leader.
+Theoretically, the smaller the time synchronization error, the better. PD can tolerate any synchronization error, but a larger error value means a larger gap between the timestamp allocated by the PD and the physical time, which will affect functions such as read of historical versions.
```
morgo: Maybe we can remove this sentence: "Theoretically, the smaller the time synchronization error, the better."

Since I don't think this is just theoretically and gets to the key answer quicker (pd can tolerate time sync error).

dcalvin: Makes sense to me

dcalvin:
```suggestion
PD can tolerate any synchronization error, but a larger error value means a larger gap between the timestamp allocated by the PD and the physical time, which will affect functions such as read of historical versions.
```

PR #1026, opened 2019-04-08
op-guide: Create kubernetes-local.md

morgo:
I won't link to this from the existing TOC, but this is used by the DITA toc.


morgo:
PTAL @kolbe @ceohockey60

kolbe:
LGTM

PR #1029, opened 2019-04-09
op-guide: differentiate requirement vs. recommend

morgo:
Some parts of the requirements are really recommendations. We should make the language clearer, so someone does not think it won't work without this.


morgo:
PTAL @kolbe @ceohockey60

PR #1032, opened 2019-04-10
sql: add alias for deleted file

morgo:
This was removed in https://github.com/pingcap/docs/pull/982

I just learned how to do aliases, so this redirects from the removed location to the closest match.


morgo:
PTAL @CaitinChen

PR #1037, opened 2019-04-11
op-guide: split migration guide

morgo:
The split here is used by the DITA TOC to make the content flow easier.


morgo:
PTAL @kolbe @ceohockey60

ceohockey60:
LGTM

PR #1039, opened 2019-04-12
sql: update EXPLAIN related document

huangxiuyan:
via: https://github.com/pingcap/docs-cn/pull/1227 @CaitinChen PTAL


sql/mysql-compatibility.md
```diff
@@ -66,6 +66,10 @@ In TiDB, auto-increment columns are only guaranteed to be incremental and unique
 
 Performance schema tables return empty results in TiDB. TiDB uses a combination of [Prometheus and Grafana](https://pingcap.com/docs/op-guide/monitor/#use-prometheus-and-grafana) for performance metrics instead.
 
+### Query Execution Plan
+
+The output format of Query Execution Plan ( `EXPLAIN` / `EXPLAIN FOR` ) in TiDB is greatly different from that in MySQL. Besides, the output content and the privileges setting of `EXPLAIN FOR` are not the same as those of MySQL. See [Understand the Query Execution Plan](/sql/understanding-the-query-execution-plan.md) for more details.
```
morgo: This is already mentioned under "Database Administration" in the same document. It originally had its own section for "EXPLAIN", but I moved it under Database administration when the document was getting too long.  We then split this document into Security, Transactions and SQL... If you would like to move it back here you can.

benchmark/sysbench-v4.md
```diff
@@ -13,8 +13,10 @@ This test aims to test the performance of TiDB 3.0 Beta on NVMe SSD.
 
 ## Test version, time, and place
 
-TiDB version: 3.0 Beta
+TiDB version: 3.0 Beta  
+
 Time: February, 2019
+
```
morgo: Do we need this file in the commit?  It looks unrelated..

sql/understanding-the-query-execution-plan.md
```diff
@@ -14,7 +14,7 @@ The result of the `EXPLAIN` statement provides information about how TiDB execut
 
 - `EXPLAIN` works together with `SELECT`, `DELETE`, `INSERT`, `REPLACE`, and `UPDATE`.
 - When you run the `EXPLAIN` statement, TiDB returns the final physical execution plan which is optimized by the SQL statement of `EXPLAIN`. In other words, `EXPLAIN` displays the complete information about how TiDB executes the SQL statement, such as in which order, how tables are joined, and what the expression tree looks like. For more information, see [`EXPLAIN` output format](#explain-output-format).
-- TiDB does not support `EXPLAIN [options] FOR CONNECTION connection_id` currently. We'll do it in the future. For more information, see [#4351](https://github.com/pingcap/tidb/issues/4351).
+- TiDB now supports `EXPLAIN [options] FOR CONNECTION connection_id` which works differently than `EXPLAIN FOR` in MySQL. See [`EXPLAIN FOR CONNECTION`](#explain-for-connection)。
```
morgo:
```suggestion
- TiDB also supports `EXPLAIN [options] FOR CONNECTION connection_id`, but with minor differences from MySQL. See [`EXPLAIN FOR CONNECTION`](#explain-for-connection).
```

- I don't think users need the history that TiDB did not support EXPLAIN FOR CONNECTION.
- I think the semantic differences are only minor.  We should not scare users it is completely different, given that we have already explained the output format of EXPLAIN is different.

sql/understanding-the-query-execution-plan.md
```diff
@@ -92,6 +92,12 @@ mysql> EXPLAIN ANALYZE SELECT count(*) FROM trips WHERE start_date BETWEEN '2017
 +------------------------+---------+------+--------------------------------------------------------------------------------------------------+-----------------------------------+
 4 rows in set (0.08 sec)
 ```
+## `EXPLAIN FOR CONNECTION`
+
+`EXPLAIN FOR CONNECTION` is used for getting the last query execution plan in one connection. The output format is totally the same as `EXPLAIN`. But the execution in TiDB is different from that in MySQL. Except for the output format, there are other differences as followed:
```
morgo:
```suggestion
`EXPLAIN FOR CONNECTION` provides the last query execution plan used by a given connection. The output format is the same as `EXPLAIN` in TiDB. The usage has the following semantic differences from MySQL: 
```

sql/understanding-the-query-execution-plan.md
```diff
@@ -92,6 +92,12 @@ mysql> EXPLAIN ANALYZE SELECT count(*) FROM trips WHERE start_date BETWEEN '2017
 +------------------------+---------+------+--------------------------------------------------------------------------------------------------+-----------------------------------+
 4 rows in set (0.08 sec)
 ```
+## `EXPLAIN FOR CONNECTION`
+
+`EXPLAIN FOR CONNECTION` is used for getting the last query execution plan in one connection. The output format is totally the same as `EXPLAIN`. But the execution in TiDB is different from that in MySQL. Except for the output format, there are other differences as followed: 
+
+- **EXECUTE** query plan is returned in MySQL while **LAST EXECUTED** is returned in TiDB.
```
morgo:
```suggestion
- MySQL returns the plan for the currently **executing query** while TiDB returns the execution plan for the **last executed** query.
```

sql/understanding-the-query-execution-plan.md
```diff
@@ -92,6 +92,12 @@ mysql> EXPLAIN ANALYZE SELECT count(*) FROM trips WHERE start_date BETWEEN '2017
...
+- In MySQL document, the connection of the user account should be the same as that of query or the user account has the `PROCESS` privileges. For TiDB, the user account should have the same connection as the query or has the `SUPER` privileges.
```
morgo:
```suggestion
- In MySQL, executing `EXPLAIN FOR CONNECTION` for connections owned by other users requires the `PROCESS` privilege. In TiDB, this statement requires the `SUPER` privilege.
```

sql/mysql-compatibility.md
huangxiuyan: Yeah, you're right. I ask kenan for confirmation and he suggested deleting the part in "Data Administration" because the part in "Query Execution Plan" is more specific.

morgo: No problem - I think because the document split, it is fine to have its own section.  I have changed "Database Administration" to be "Analyze table", because it doesn't make sense to have this heading for only one point any more.

CaitinChen:
```suggestion
The output format of Query Execution Plan (`EXPLAIN`/`EXPLAIN FOR`) in TiDB is greatly different from that in MySQL. Besides, the output content and the privileges setting of `EXPLAIN FOR` are not the same as those of MySQL. See [Understand the Query Execution Plan](/sql/understanding-the-query-execution-plan.md) for more details.
```

CaitinChen:
@eurekaka PTAL

eurekaka:
@erjiaqing PTAL

huangxiuyan:
@erjiaqing PTAL, many thanks!

PR #1040, opened 2019-04-12
sql: Update transaction-model.md subheading

morgo:
This page was previously called "transactions: mysql compatibility".  Because that was dropped the differences heading makes less sense.. differences from what?

Updating the sub-heading to be differences from MySQL.


morgo:
@dcalvin PTAL

PR #1047, opened 2019-04-15
TOC: fix a broken link

CaitinChen:
@lilin90 PTAL


lilin90:
@morgo FYI. The `(Enterprise)` causes a broken link, so it is removed.

PR #1049, opened 2019-04-15
overview: add binary deploy

morgo:
Add binary install guides.

(Edit: this PR also fixed a bug, but @dcalvin had a better fix in https://github.com/pingcap/docs/pull/1051 )


morgo:
PTAL @CaitinChen

kolbe:
Can I suggest that you move the Binary Deployment option to the bottom of the list, since we'd probably prefer that people use Ansible or other deployment methods?

morgo:
@kolbe done

PR #1054, opened 2019-04-16
toc, config flags: fix inconsistencies in headings

dcalvin:


op-guide/tidb-configuration.md
```diff
@@ -4,54 +4,54 @@ summary: Learn some configuration flags for TiDB
 category: operations
 ---
```
lilin90: Please help update the `title` metadata: TIDB -> **TiDB**

morgo:
@dcalvin This renames a file without an alias to redirect. Can we please change it back, since it will be moved again once the ToC merges.

dcalvin:
> @dcalvin This renames a file without an alias to redirect. Can we please change it back, since it will be moved again once the ToC merges.
@morgo Will revert the change for now, and change the file name and alias after the URL and directories are migrated.

PR #1062, opened 2019-04-18
v1.0, v2.0, v2.1: remove older versions

morgo:
This PR removes the older versions from the docs.  Let me explain the rationale:

1. The DITA TOC will not be backported to v1.0 or v2.0 (we should put a PDF up for download for these users).
2. To create the DITA TOC for 2.1, I plan to start with dev/ and then remove the content related to 3.0. This is easier than bringing the current v2.1 forward to using the new TOC.

The reason I am proposing we do this now, is some contributors may look to backport their changes to v2.1, but they will be lost, as I plan to abandon this directory for the strategy mentioned in (2).

PTAL @queenypingcap @lilin90 If that sounds good to you, I will convert this from DRAFT to PR.


queenypingcap:
I think it would be okay. @lilin90 Any potential risks？

lilin90:
@morgo @queenypingcap **I suggest keeping the docs/v1.0 and docs/v2.0 directories** because some TiDB users are still using an earlier version like 1.0 or 2.0.

I also checked with @liubo0127, he also suggested keeping 1.0 and 2.0 docs, in case that users turn to us for support of 1.0/2.0.

Several days ago, the tools team told me that different versions of TiDB-Binlog are all still being used. They also hope those documents can be kept.

Let's keep a docs version for earlier TiDB versions at pingcap/docs, but we won't apply the new TOC for them. We are still maintaining them, though nor often, like typos or some inappropriate descriptions.

PR #1065, opened 2019-04-19
tools: update the loader pprof-addr default port

ericsyh:
This PR updates the pprof-addr default port in Loader documents.

According to the https://internal.pingcap.net/confluence/pages/viewpage.action?pageId=21273346, Loader pprof-addr default port has changed from 10084 to 8272 while the public documents is out of date.

Loader code for reference:
https://github.com/pingcap/tidb-enterprise-tools/blob/master/loader/config.go
fs.StringVar(&cfg.PprofAddr, "pprof-addr", ":8272", "Loader pprof addr")

And the same PR in docs-cn： https://github.com/pingcap/docs-cn/pull/1256

### DIFF
--- a/tools/loader.md
+++ b/tools/loader.md
@@ -60,7 +60,7 @@ Since tools like mysqldump will take us days to migrate massive amounts of data,
 
   -p string: the account and password of TiDB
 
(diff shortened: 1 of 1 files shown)


lilin90:
@GregoryIan @amyangfei PTAL Thanks!

PR #1095, opened 2019-04-30
Removing unnecessary warning message

ceohockey60:
Removing unnecessary warning message about Operator, which currently gets rendered on website


gregwebs:
@tennix PTAL

PR #1097, opened 2019-05-04
sql/mysql-compatibility: Improve DDL description

morgo:
The rest of this page shows what TiDB is not capable of. The DDL section is written a bit weird, since it is saying what TiDB is capable of, then pointing out exceptions.

I think we should just assume if it is not mentioned, it is supported. This helps with readability.


winkyao:
@zimulala Please help Morgan to complete what we are not supported.

sql/mysql-compatibility.md
```diff
@@ -78,38 +78,23 @@ TiDB supports most of the MySQL built-in functions, but not all. See [TiDB SQL G
...
-        >
-        > The changing/modifying column operation cannot make the length of the original type become shorter and it cannot change the unsigned/charset/collate attributes of the column.
-
-    - Supports changing the following type definitions: `default value`, `comment`, `null`, `not null` and `OnUpdate`.
-    - Supports parsing the `LOCK [=] {DEFAULT|NONE|SHARED|EXCLUSIVE}` syntax, but there is no actual operation.
+    - Does not support setting a column as the `PRIMARY KEY`, or creating a unique index, or specifying `auto_increment` while adding it.
++ Drop Column: Does not support dropping the `PRIMARY KEY` column or index column.
++ Change/Modify Column:
+    - Does not support lossy changes, such as from `BIGINT` to `INTEGER` or `VARCHAR(255)` to `VARCHAR(10)`.
+    - Does not support changing the `UNSIGNED`, `CHARACTER SET` or `COLLATE` attributes.
```
zimulala: About changing `CHARACTER SET`,  now we only support for `CharsetUTF8` -> `CharsetUTF8MB4`.

sql/mysql-compatibility.md
```diff
@@ -78,38 +78,23 @@ TiDB supports most of the MySQL built-in functions, but not all. See [TiDB SQL G
...
 ### DDL
 
-TiDB implements the asynchronous schema changes algorithm in F1. The Data Manipulation Language (DML) operations cannot be blocked during DDL the execution. Currently, the supported DDL includes:
+In TiDB DDL does not block reads or writes to tables while in operation. However, some restrictions currently apply to DDL changes:
 
-+ Create Database
-+ Drop Database
-+ Create Table
-+ Drop Table
 + Add Index: Does not support creating multiple indexes at the same time.
```
zimulala: Don't support some operations about the generated column, like doing `add index` to a generated column.
Ref https://github.com/pingcap/docs/blob/master/v2.1/sql/generated-columns.md#limitations

zimulala:
I am sorry for late reply.

morgo:
> About `rebase AUTO_INCREMENT`:
> Assume there are 2 TiDB servers: TiDB-A with allocator range of 0 ~ 30000; TiDB-B with allocator range of 30001 ~ 60000.
> If the user sends SQL `alter table t1 auto_increment = 6` to a TiDB, the one TiDB autoIncID will start from 60001, another will start from 90001.

There is a section on auto-increment in the page. It kind of describes this behavior.

> Besides, do we need to consider `create view` and `table partition`？

Yes, but I will add them later. We are going to split the manual into multiple versions next week, and I will use the current 'master' for 2.1 as well.

morgo:
@zimulala thank you for the suggestions, I've added a note about generated columns and charset. PTAL again.

zimulala:
OK.

dev/reference/mysql-compatibility.md
```diff
@@ -79,38 +79,26 @@ TiDB supports most of the MySQL built-in functions, but not all. See [TiDB SQL G
...
+  - Adding an index on a generated column via `ALTER TABLE` is not supported.
++ Add Column:
+  - Does not support creating multiple columns at the same time.
+  - Does not support setting a column as the `PRIMARY KEY`, or creating a unique index, or specifying `auto_increment` while adding it.
++ Drop Column: Does not support dropping the `PRIMARY KEY` column or index column.
++ Change/Modify Column:
+  - Does not support lossy changes, such as from `BIGINT` to `INTEGER` or `VARCHAR(255)` to `VARCHAR(10)`.
+  - Does not support changing the `UNSIGNED` attribute.
+  - Only supports changing the `CHARACTER SET` attribute from `utf8` to `utf8mb4`.
++ `LOCK [=] {DEFAULT|NONE|SHARED|EXCLUSIVE}`: the syntax is supported, but is not applicable to TiDB. All DDL changes that are supported do not lock the table.
```
zimulala: About the  syntax of "ALGORITHM [=] {DEFAULT|INSTANT|INPLACE|COPY}" is supported like `LOCK`. But we do some checks for it. Ref: https://github.com/pingcap/tidb/pull/9367

morgo: @zimulala Thank you. PTAL again.

PR #1111, opened 2019-05-08
Update remaining how-to pages to use new URL structure

morgo:
This is the same branch as https://github.com/pingcap/docs/pull/1081 which was reverted.

It was tested again under https://github.com/pingcap/docs/pull/1094 and should be ready to merge.


morgo:
PTAL @lilin90 @dcalvin

morgo:
I have also moved roadmap/contributing/tech support to top-level TOC items. These were changes proposed in https://github.com/pingcap/docs-cn/pull/1279

morgo:
DNM, I will wait for https://github.com/pingcap/docs/pull/1113 to merge first.

morgo:
This PR is now ready for review again. PTAL @lilin90

TOC.md
```diff
@@ -50,38 +50,38 @@
...
   + Maintain
-    - [Backup and Restore](op-guide/backup-restore.md)
+    - [Backup and Restore](dev/how-to/maintain/backup-and-restore.md)
     + Scale
-      - [Scale a TiDB Cluster](op-guide/horizontal-scale.md)
-      - [Scale Using Ansible](op-guide/ansible-deployment-scale.md)
-    - [Identify Slow Queries](sql/slow-query.md)
+      - [Scale a TiDB Cluster](dev/how-to/maintain/scale/horizontally.md)
+      - [Scale using Ansible](dev/how-to/maintain/scale/with-ansible.md)
+    - [Identify slow queries](dev/how-to/maintain/identify-slow-queries.md)
```
dcalvin: Shouldn't this be all initial-capped for top-level headers? (also applied to other similar changes)
```suggestion
    - [Identify Slow Queries](dev/how-to/maintain/identify-slow-queries.md)
```

dcalvin: This may not be an issue - but too deep a hierarchy is making both the file naming and the corresponding URL look a little bit weird and illogical

morgo: If we do this, it would be nice to have a convention, such as including the first categorization (how-to) and the last (scale) while omitting the others (maintain). I like how-to/scale/ .. but I don't necessarily want to change the menu indentation to match, because it creates too many items at that leve.

morgo: I chatted with @kolbe on this for a second opinion. He brought up the point that for the index pages (the 'cards design' for section overview), it makes sense for the hierarchy to be preserved in the url. So deleting the last part of a url lands on an index page.. it is a small feature, but nice to have.

PR #1114, opened 2019-05-09
op-guide: update docs for shared block cache

yiwu-arbug:
Update docs to reflect that in tikv3.0 block cache will be shared among column families by default.

### DIFF
--- a/op-guide/dashboard-tikv-info.md
+++ b/op-guide/dashboard-tikv-info.md
@@ -248,7 +248,7 @@ RocksDB - kv | SST read duration | The time consumed when reading SST files |
 RocksDB - kv | Write stall duration | The time which is caused by write stall | It should be `0` in normal case.
 RocksDB - kv | Memtable size | The memtable size of each column family |
 RocksDB - kv | Memtable hit | The hit rate of memtable |
-RocksDB - kv | Block cache size | The block cache size of each column family |
+RocksDB - kv | Block cache size | The block cache size. Breakdown by column family if shared block cache is disabled. |
 RocksDB - kv | Block cache hit | The hit rate of block cache |
 RocksDB - kv | Block cache flow | The flow of different kinds of block cache operations |
 RocksDB - kv | Block cache operations | The count of different kinds of block cache operations |
@@ -281,7 +281,7 @@ RocksDB - raft | SST read duration | The time consumed when reading SST files |
 RocksDB - raft | Write stall duration | The time which is caused by write stall | It should be `0` in normal case.
 RocksDB - raft | Memtable size | The memtable size of each column family |
 RocksDB - raft | Memtable hit | The hit rate of memtable |
-RocksDB - raft | Block cache size | The block cache size of each column family |
+RocksDB - raft | Block cache size | The block cache size. Breakdown by column family if shared block cache is disabled. |
 RocksDB - raft | Block cache hit | The hit rate of block cache |
 RocksDB - raft | Block cache flow | The flow of different kinds of block cache operations |
 RocksDB - raft | Block cache operations | The count of different kinds of block cache operations |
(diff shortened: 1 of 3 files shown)


op-guide/dashboard-tikv-info.md
```diff
@@ -248,7 +248,7 @@ RocksDB - kv | SST read duration | The time consumed when reading SST files |
 RocksDB - kv | Write stall duration | The time which is caused by write stall | It should be `0` in normal case.
 RocksDB - kv | Memtable size | The memtable size of each column family |
 RocksDB - kv | Memtable hit | The hit rate of memtable |
-RocksDB - kv | Block cache size | The block cache size of each column family |
+RocksDB - kv | Block cache size | The block cache size. Breakdown by column family if shared block cache is disabled. |
```
lilin90: The "Breakdown by" here is confusing for me. Could you explain more about it? A noun here looks weird.

op-guide/tune-tikv.md
```diff
@@ -23,7 +23,7 @@ TiKV implements `Column Families` (CF) from RocksDB.
     
     - The `default` CF stores the Raft log. The corresponding parameters are in `[raftdb.defaultcf]`.
 
-Each CF has a separate `block cache` to cache data blocks to accelerate the data reading speed in RocksDB. You can configure the size of the `block cache` by setting the `block-cache-size` parameter. The bigger the `block-cache-size`, the more hot data can be cached, and the easier to read data, in the meantime, the more system memory will be occupied.
+By default, all CFs share one block cache instance. You can configure the size of the cache by setting `capacity` parameter under `[storage.block-cache]`. The bigger the block cache, the more hot data can be cached, and the easier to read data, in the meantime, the more system memory will be occupied. To use separate block cache instance for each CF, set `shared=false` under `[storage.block-cache]`, and configure individual block cache size for each CF. For example, you can configure the size of `write` CF by setting `block-cache-size` parameter under `[rocksdb.writecf]`.
```
lilin90:
```suggestion
By default, all CFs share one block cache instance. You can configure the size of the cache by setting the `capacity` parameter under `[storage.block-cache]`. The bigger the block cache, the more hot data can be cached, and the easier to read data, in the meantime, the more system memory is occupied. To use a separate block cache instance for each CF, set `shared=false` under `[storage.block-cache]`, and configure individual block cache size for each CF. For example, you can configure the size of `write` CF by setting the `block-cache-size` parameter under `[rocksdb.writecf]`.
```

op-guide/tune-tikv.md
```diff
@@ -62,6 +62,30 @@ log-level = "info"
 # `scheduler-worker-pool-size` higher and increase the number of write threads.
 # scheduler-worker-pool-size = 4
 
+[storage.block-cache]
+## Whether to create a shared block cache for all RocksDB column families.
+##
+## Block cache is used by RocksDB to cache uncompressed blocks. Big block cache can speed up read.
+## It is recommended to turn on shared block cache. Since only the total cache size need to be
+## set, it is easier to config. In most cases it should be able to auto-balance cache usage
```
lilin90:
```suggestion
## set, it is easier to configure. In most cases, it should be able to auto-balance cache usage
```

op-guide/dashboard-tikv-info.md
yiwu-arbug: @lilin90 thanks for helping! For the "breakdown", let me give you an example. When shared cache is used, Grafana will show a single cache size, label with "all":
Screen Shot 2019-05-08 at 4 41 29 PM
when shared cache is disabled, Grafana will show individual cache size for each CFs:
Screen Shot 2019-05-09 at 10 46 50 AM

yiwu-arbug:
@lilin90 Do you have more comments? And btw, are you able to help with the Chinese version? :)

op-guide/dashboard-tikv-info.md
lilin90: @yiwu-arbug Got it. Thanks for your explanation!

lilin90:
> @lilin90 Do you have more comments? And btw, are you able to help with the Chinese version? :)

@yiwu-arbug Sure. We'll update it accordingly in docs-cn.

zhangjinpeng1987:
Seems we should mention that this change is for 3.0 version, for the 2.1 version we should keep the old way.

PR #1115, opened 2019-05-09
[DNM] tools/lightning: fix download URL for 2.1.9

kennytm:
It has been split into two tarballs.

DNM: Awaiting decision if we should fix the docs or fix the tarballs.


kennytm:
PTAL @CaitinChen @GregoryIan

kennytm:
tidb-lightning has been put back into `tidb-v2.1.9-linux-amd64.tar.gz` and for all future 2.1 releases. Closing.

PR #1123, opened 2019-05-10
*: Update reference to new URL structure

morgo:
Renames reference urls to new locations. Still need to update internal links within documents + update toc


morgo:
@dcalvin @lilin90 PTAL

I have done something different with this PR, where the TOC and internal links point to the old locations and then use redirections to find the new locations. I did this to reduce merge conflicts during the review (it has been quite difficult resolving them).

I will update the URLs to go to the new locations in a followup PR after this one.

PR #1136, opened 2019-05-14
sql: redirect orphans to new locations

morgo:
This PR removes some orphans, and adds redirects to the closest match in the TOC:

* sql/ddl.md: was replaced by the statement reference organization, and now redirects to `CREATE TABLE` (closest match). `CREATE TABLE` links to all the other pages in 'see also', so it is hopefully reasonable for those redirected here.
* sql/dml.md: was replaced by the statement reference organization, and now redirects to `SELECT` (this is a non-perfect match, but again.. there are see also links).
* sql/prepare.md: was replaced by the statement reference organization, and now redirects to `PREPARE` (this is a good match).
* sql/server-command-option.md: this was redundant with 3 different files commenting about configuration. It now redirects to dev/reference/configuration/tidb-server/configuration.md (good match).
* sql/tidb-server.md: this is an overview file. It could make sense as an index page when we support them, but only provides incremental value, so I've removed it. The index pages will also likely follow a different format with 'cards'.
* sql/user-manual.md: this was only a draft and never shown in public. It was a type of index page.
* sql/util.md: this is made obsolete by statement reference as well, and maps closest to EXPLAIN docs. The EXPLAIN page now includes the additional examples here, and a see also for TRACE.

There are now three remaining files in `sql/*`:
- error.md: due to an issue with redirects (need designers to investigate)
- date-and-time-types.md: it is an orphan, but has good content. Need to figure out how to incorporate.
- admin.md: not yet in the statement reference, will be added in a second push.


morgo:
@dcalvin incorporated changes, PTAL again

PR #1138, opened 2019-05-14
reference: improve set statement

morgo:
This PR includes separate `SET` pages for password and character set, since the `SET [GLOBAL|SESSION]` pages do not cover them accurately.


morgo:
@jackysp PTAL (covers charset and privilege system commands)

PR #1144, opened 2019-05-15
reference: split data types to separate pages

morgo:
This PR splits datatypes into a page per data type category + a separate overview and a page for default values. The format is also changed in the page to show an anchor point per type so that it can easily be added in the TOC.

The motivation for doing this, is that it was planned in the DITA conversion (similar to per-statement guide: this makes it easier as a reference). The motivation for doing this now, is there was an orphan document in `sql/date-and-time-types.md` which duplicated the previous data types page (but was longer). To be able to delete it would have meant merging in content to `data-types.md`, which was already too long.


dev/reference/sql/data-types/string.md
```diff
@@ -0,0 +1,161 @@
...
+
+`VARCHAR` is a string of variable-length. M represents the maximum column length in characters (not bytes). The range of M is 0 to 65,535, but the effective maximum-length will be shorter since the total size of all columns must not exceed 65,535 bytes (the maximum row-size in TIDB).
+
+```sql
+[NATIONAL] VARCHAR(M) [CHARACTER SET charset_name] [COLLATE collation_name]
+```
+
+### `TINYTEXT` type
+
+`TINYTEXT` is a string of variable-length. The length (M) is optional, with the maximum length being restricted to 255 characters:
```
dcalvin:
```suggestion
`TINYTEXT` is a string of variable-length. The length (M) is optional, with a maximum length of 255 characters:
```

dev/reference/sql/data-types/string.md
```diff
@@ -0,0 +1,161 @@
...
+
+```sql
+TINYTEXT[(M)] [CHARACTER SET charset_name] [COLLATE collation_name]
+```
+
+### `TEXT` type
+
+`TEXT` is a string of variable-length. M represents the maximum column length ranging from 0 to 65,535. The range of M is 0 to 65,535, but the effective maximum-length will be shorter since the total size of all columns must not exceed 65,535 bytes (the maximum row-size in TIDB):
```
dcalvin:
```suggestion
`TEXT` is a string of variable-length. M represents the maximum column length ranging from 0 to 65,535. The effective maximum-length will be shorter since the total size of all columns must not exceed 65,535 bytes (the maximum row-size in TIDB):
```

dev/reference/sql/data-types/string.md
```diff
@@ -0,0 +1,161 @@
...
+
+The `TINYBLOB` type is similar to the `TINYTEXT` type, but stores binary byte strings rather than nonbinary character strings:
+
+```sql
+TINYBLOB
+```
+
+### `BLOB` type
+
+The `BLOB` type is similar to the `TEXT` type, but stores binary byte strings rather than nonbinary character strings. M represents the maximum column length, up to 65,535 bytes. The effective maximum-length will be shorter since the total size of all columns must not exceed 65,535 bytes (the maximum row-size in TIDB):
```
dcalvin:
```suggestion
The `BLOB` type is similar to the `TEXT` type, but stores binary byte strings rather than nonbinary character strings. M represents the maximum column length ranging from 0 to 65,535 bytes. The effective maximum-length will be shorter since the total size of all columns must not exceed 65,535 bytes (the maximum row-size in TIDB):
```

dev/reference/sql/data-types/string.md
```diff
@@ -0,0 +1,161 @@
...
+
+The `BINARY` type is similar to the `CHAR` type, but stores binary byte strings rather than nonbinary character strings. M represents the maximum length in bytes:
+
+```sql
+BINARY(M)
+```
+
+### `VARBINARY` type
+
+The `VARBINARY` type is similar to the `VARCHAR` type, but stores binary byte strings rather than nonbinary character strings. M represents the maximum length, up to 65,535 bytes:
```
dcalvin:
```suggestion
The `VARBINARY` type is similar to the `VARCHAR` type, but stores binary byte strings rather than nonbinary character strings. M represents the maximum length, ranging from 0 to 65,535 bytes:
```

morgo:
I have removed the section 'Storage and range' from numeric types. The range is covered in each of the data types, and the storage size is a mysql-ism and incorrect. TiKV works differently.

dev/reference/sql/data-types/date-and-time.md
```diff
@@ -230,3 +191,71 @@ When numeral `00` is inserted to `YEAR(4)`, the result is 0000 rather than 2000.
...
+TIMESTAMP[(fsp)]
+```
+
+#### Timezone Handling
+
+When `TIMESTAMP` is to be stored, TiDB converts the `TIMESTAMP` value from the current time zone to UTC time zone. When `TIMESTAMP`  is to be retrieved, TiDB converts the stored `TIMESTAMP` value from UTC time zone to the current time zone (Note: `DATETIME` is not handled in this way). The default time zone for each connection is the server's local time zone, which can be modified by the environment variable `time_zone`.
+
+> **Warning:**
+> 
+> As in MySQL, the `TIMESTAMP` data type suffers from the [Year 2038 Problem](https://en.wikipedia.org/wiki/Year_2038_problem). For storing values that may span beyond 2038, please consider using the `DATETIME` type instead.
```
winkyao: as far as know, TiDB doesn't use 32-bit integer to store timestamp, so there is no `Year 2038 problem` in TiDB.

morgo: It uses 64-bit, but there are 3 type systems in TiKV (one for schema, one for comparisons and one for storage). The schema type system enforces range, and prevents values beyond 2038 from being stored:

```
mysql> CREATE TABLE t1 (a int not null primary key auto_increment, b timestamp not null);
Query OK, 0 rows affected (0.08 sec)

mysql> INSERT INTO t1 (b) VALUES ('2039-01-01 10:00:00');
ERROR 1292 (22007): invalid time format: '{2039 1 1 10 0 0 0}'
mysql> INSERT INTO t1 (b) VALUES ('2030-01-01 10:00:00');
Query OK, 1 row affected (0.01 sec)
```

morgo:
In the future we should add some detail about TiKV's type systems, but I have left it out for now.

PR #1146, opened 2019-05-15
sql: Move orphan admin command page

morgo:
This moves the orphan 'admin.md' to the sql statement reference for admin (best fit). Much of the other content on this admin page is already in the statement reference under other commands.

Long term I would like to document each of the admin commands under statement reference, but it is quite a lot of work, so out of scope for now.


morgo:
@dcalvin PTAL

PR #1155, opened 2019-05-17
dev/reference/sql: update `explain` statement

huangxiuyan:
Ref: https://github.com/pingcap/docs-cn/pull/1324
PTAL @lilin90 @morgo


dev/reference/sql/statements/explain.md
```diff
@@ -91,7 +91,7 @@ mysql> EXPLAIN DELETE FROM t1 WHERE c1=3;
 +---------------------+----------+------+-------------------------------------------------------------+
 3 rows in set (0.00 sec)
 ```
-In addition to the MySQL standard result format, TiDB also supports DotGraph and you need to specify `FORMAT = "dot"` as in the following example:
+In addition to the MySQL standard result format, TiDB also supports DotGraph and you need to specify `FORMAT = "dot"` or `FORMAT = "row"` as in the following example:
```
lilin90: This is not what the docs-cn PR means. Please check the meaning again.

dev/reference/sql/statements/explain.md
```diff
@@ -91,7 +91,9 @@ mysql> EXPLAIN DELETE FROM t1 WHERE c1=3;
 +---------------------+----------+------+-------------------------------------------------------------+
 3 rows in set (0.00 sec)
 ```
-In addition to the MySQL standard result format, TiDB also supports DotGraph and you need to specify `FORMAT = "dot"` or `FORMAT = "row"` as in the following example:
+If you do not specify the `FORMAT`, or specify `FORMAT = "row"`, `EXPLAIN` statement will ouput the results in MySQL standard result format. But the field is different with that of MySQL. See [Understand the Query Execution Plan](https://pingcap.com/docs/dev/reference/performance/understanding-the-query-execution-plan/) for more information.
```
morgo: This is actually not true. The format from TiDB is very different from MySQL's tabular explain.

huangxiuyan: Could you please provide a better explanation of the output format. I think the point is "how different", since the doc-cn believe it is a little bit different(only the field). Thanks!
@jackysp Could you please take a look at the output format difference? Thanks!

morgo: @huachaohuang It is completely different.  Try a simple example:
```
In TiDB:
mysql> CREATE TABLE t1 (a int primary key not null);

mysql> insert into t1 values (1),(2),(3);
Query OK, 3 rows affected (0.01 sec)
Records: 3  Duplicates: 0  Warnings: 0

mysql> explain select * from t1 where a = 1;
+-------------+-------+------+--------------------+
| id          | count | task | operator info      |
+-------------+-------+------+--------------------+
| Point_Get_1 | 1.00  | root | table:t1, handle:1 |
+-------------+-------+------+--------------------+
1 row in set (0.00 sec)

In MySQL:
mysql [localhost:8015] {msandbox} (test) > CREATE TABLE t1 (a int primary key not null);
Query OK, 0 rows affected (0.05 sec)

mysql [localhost:8015] {msandbox} (test) > insert into t1 values (1),(2),(3);
Query OK, 3 rows affected (0.02 sec)
Records: 3  Duplicates: 0  Warnings: 0

mysql [localhost:8015] {msandbox} (test) > explain select * from t1 where a = 1;
+----+-------------+-------+------------+-------+---------------+---------+---------+-------+------+----------+-------------+
| id | select_type | table | partitions | type  | possible_keys | key     | key_len | ref   | rows | filtered | Extra       |
+----+-------------+-------+------------+-------+---------------+---------+---------+-------+------+----------+-------------+
|  1 | SIMPLE      | t1    | NULL       | const | PRIMARY       | PRIMARY | 4       | const |    1 |   100.00 | Using index |
+----+-------------+-------+------------+-------+---------------+---------+---------+-------+------+----------+-------------+
1 row in set, 1 warning (0.00 sec)
```

morgo: In terms of how to describe it, just say it is different as in this page: https://pingcap.com/docs/dev/reference/mysql-compatibility/

dev/reference/sql/statements/explain.md
```diff
@@ -91,6 +91,8 @@ mysql> EXPLAIN DELETE FROM t1 WHERE c1=3;
 +---------------------+----------+------+-------------------------------------------------------------+
 3 rows in set (0.00 sec)
 ```
+If you do not specify the `FORMAT`, or specify `FORMAT = "row"`, `EXPLAIN` statement will ouput the results in MySQL standard result format. But the field is different with that of MySQL. See [Understand the Query Execution Plan](https://pingcap.com/docs/dev/reference/performance/understanding-the-query-execution-plan/) for more information.
```
morgo:
```suggestion
If you do not specify the `FORMAT`, or specify `FORMAT = "row"`, `EXPLAIN` statement will output the results in a tabular format. See [Understand the Query Execution Plan](https://pingcap.com/docs/dev/reference/performance/understanding-the-query-execution-plan/) for more information. 
```

huangxiuyan:
@jackysp PTAL

dev/reference/sql/statements/explain.md
huangxiuyan: Addressed.

PR #1158, opened 2019-05-20
dev/reference: add date and time compatibility notes

huangxiuyan:
Via: https://github.com/pingcap/docs-cn/pull/1304
PTAL @lilin90

### DIFF
--- a/dev/reference/mysql-compatibility.md
+++ b/dev/reference/mysql-compatibility.md
@@ -2,7 +2,7 @@
 title: Compatibility with MySQL
 summary: Learn about the compatibility of TiDB with MySQL, and the unsupported and different features.
 category: reference
-aliases: ['/docs/sql/mysql-compatibility/'] 
+aliases: ['/docs/sql/mysql-compatibility/']
 ---
 
 # Compatibility with MySQL
@@ -51,15 +51,15 @@ In TiDB, auto-increment columns are only guaranteed to be incremental and unique
 > **Warning:**
 >
 > If you use auto-increment IDs in a cluster with multiple tidb-server instances, do not mix default values and custom values, otherwise an error will occur in the following situation:
-> 
+>
 > Assume that you have a table with the auto-increment ID:
-> 
+>
 > `create table t(id int unique key auto_increment, c int);`
-> 
+>
 > The principle of the auto-increment ID in TiDB is that each tidb-server instance caches a section of ID values (currently 30000 IDs are cached) for allocation and fetches the next section after this section is used up.
 >
 > Assume that the cluster contains two tidb-server instances, namely Instance A and Instance B. Instance A caches the auto-increment ID of [1, 30000], while Instance B caches the auto-increment ID of [30001, 60000].
-> 
+>
 > The operations are executed as follows:
 >
 > 1. The client issues the `insert into t values (1, 1)` statement to Instance B which sets the `id` to 1 and the statement is executed successfully.
@@ -100,7 +100,7 @@ For more information, see [Online Schema Changes](/key-features.md#online-schema
 ### Analyze table
 
 [`ANALYZE TABLE`](/dev/reference/performance/statistics.md#manual-collection) works differently in TiDB than in MySQL, in that it is a relatively lightweight and short-lived operation in MySQL/InnoDB, while in TiDB it completely rebuilds the statistics for a table and can take much longer to complete.
-    
+
 ### Storage engines
 
 For compatibility reasons, TiDB supports the syntax to create tables with alternative storage engines. Metadata commands describe tables as being of engine InnoDB:
@@ -126,7 +126,7 @@ TiDB supports **all of the SQL modes** from MySQL 5.7 with minor exceptions:
 
 - The compatibility modes deprecated in MySQL 5.7 and removed in MySQL 8.0 are not supported (such as `ORACLE`, `POSTGRESQL` etc).
 - The mode `ONLY_FULL_GROUP_BY` has minor [semantic differences](/dev/reference/sql/functions-and-operators/aggregate-group-by-functions.md#differences-from-mysql) to MySQL 5.7, which we plan to address in the future.
-- The SQL modes `NO_DIR_IN_CREATE` and `NO_ENGINE_SUBSTITUTION` are supported for compatibility, but are not applicable to TiDB. 
+- The SQL modes `NO_DIR_IN_CREATE` and `NO_ENGINE_SUBSTITUTION` are supported for compatibility, but are not applicable to TiDB.
 
 ### Default differences
 
@@ -148,7 +148,7 @@ TiDB supports **all of the SQL modes** from MySQL 5.7 with minor exceptions:
     - The default value in MySQL:
         - For MySQL 5.7: `OFF`
         - For MySQL 8.0: `ON`
-    
+
 ### Date and Time
 
 #### Named timezone
@@ -159,12 +159,12 @@ Because they are built-in, named time zones in TiDB might behave slightly differ
 
 > **Note:**
 >
-> TiKV calculates time-related expressions that can be pushed down to it. This calculation uses the built-in time zone rule and does not depend on the time zone rule installed in the system. If the time zone rule installed in the system does not match the version of the built-in time zone rule in TiKV, the time data that can be inserted cannot be read in a few cases.
-> 
-> For example, if the tzdata 2018a time zone rule is installed in the system, the time `1988-04-17 02:00:00` can be inserted into TiDB of the 3.0.0-rc.1 version when the time zone is set to Asia/Shanghai or the time zone is set to the local time zone and the local time zone is Asia/Shanghai. But this record can no longer be read out by specific types of SQL statements because this time does not exist in the Asia/Shanghai time zone according to the tzdata 2018i time zone rule used by TiKV 3.0.0-rc.1. Daylight saving time is one hour late.
-> 
-> The named timezone rules in TiKV of two versions are as follows: 
-> 
+> TiKV calculates time-related expressions that can be pushed down to it. This calculation uses the built-in time zone rule and does not depend on the time zone rule installed in the system. If the time zone rule installed in the system does not match the version of the built-in time zone rule in TiKV, the time data that can be inserted might result in a statement error in a few cases.
+>
+> For example, if the tzdata 2018a time zone rule is installed in the system, the time `1988-04-17 02:00:00` can be inserted into TiDB of the 3.0.0-rc.1 version when the time zone is set to Asia/Shanghai or the time zone is set to the local time zone and the local time zone is Asia/Shanghai. But reading this record might result in a statement error because this time does not exist in the Asia/Shanghai time zone according to the tzdata 2018i time zone rule used by TiKV 3.0.0-rc.1. Daylight saving time is one hour late.
+>
+> The named timezone rules in TiKV of two versions are as follows:
+>
 > - 3.0.0 RC.1 and later: [tzdata 2018i](https://github.com/eggert/tz/tree/2018i)
 > - 2.1.0 RC.1 and later: [tzdata 2018e](https://github.com/eggert/tz/tree/2018e)
 
--- a/dev/reference/mysql-compatibility.md
+++ b/dev/reference/mysql-compatibility.md
@@ -170,4 +170,4 @@ Because they are built-in, named time zones in TiDB might behave slightly differ
 
 #### Zero month and zero day
 
-It is not recommended to unset the SQL modes `NO_ZERO_DATE` or `NO_ZERO_IN_DATE`, which are enabled by default in TiDB as in MySQL. While TiDB supports operating with these modes disabled, the TiKV coprocessor does not. Executing certain statements that push down date and time processing functions to TiKV might result in a statement error.
+It is not recommended to unset the `NO_ZERO_DATE` and `NO_ZERO_IN_DATE` SQL modes, which are enabled by default in TiDB as in MySQL. While TiDB supports operating with these modes disabled, the TiKV coprocessor does not. Executing certain statements that push down date and time processing functions to TiKV might result in a statement error.
--- a/dev/reference/mysql-compatibility.md
+++ b/dev/reference/mysql-compatibility.md
@@ -168,4 +168,4 @@ Because they are built-in, named time zones in TiDB may behave slightly differen
 
 #### Zero month and zero day
 
-TiDB does not accept that the date value of month and day is 0. This kind of date data can be normally inserted but might not be read by certain types of SQL statements. 
+It is not recommended to unset the SQL modes `NO_ZERO_DATE` or `NO_ZERO_IN_DATE`, which are enabled by default in TiDB as in MySQL. While TiDB supports operating with these modes disabled, the TiKV coprocessor does not. Executing certain statements that push down date and time processing functions to TiKV may result in a statement error.
--- a/dev/reference/mysql-compatibility.md
+++ b/dev/reference/mysql-compatibility.md
@@ -149,3 +149,23 @@ TiDB supports **all of the SQL modes** from MySQL 5.7 with minor exceptions:
         - For MySQL 5.7: `OFF`
         - For MySQL 8.0: `ON`
     
+### Date and Time
+
+#### Time zone
+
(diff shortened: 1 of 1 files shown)


dev/reference/mysql-compatibility.md
```diff
@@ -149,3 +149,23 @@ TiDB supports **all of the SQL modes** from MySQL 5.7 with minor exceptions:
         - For MySQL 5.7: `OFF`
         - For MySQL 8.0: `ON`
     
+### Date and Time
+
+#### Time zone
+
+By default, MySQL time zone is set to the server's time zone and calculated with the system time zone rules such as the beginning of daylight saving time. Befor loading [time zone information table](https://dev.mysql.com/doc/refman/8.0/en/time-zone-support.html#time-zone-installation), you cannot specify the time zone with the time zone name.
```
morgo: I think we could make it clearer by referring to them as "Named timezones".
```suggestion
TiDB supports named timezones such as `America/Los_Angeles` without having to load the [time zone information tables](https://dev.mysql.com/doc/refman/8.0/en/time-zone-support.html#time-zone-installation) as in MySQL.
```

dev/reference/mysql-compatibility.md
```diff
@@ -149,3 +149,23 @@ TiDB supports **all of the SQL modes** from MySQL 5.7 with minor exceptions:
         - For MySQL 5.7: `OFF`
         - For MySQL 8.0: `ON`
     
+### Date and Time
+
+#### Time zone
+
+By default, MySQL time zone is set to the server's time zone and calculated with the system time zone rules such as the beginning of daylight saving time. Befor loading [time zone information table](https://dev.mysql.com/doc/refman/8.0/en/time-zone-support.html#time-zone-installation), you cannot specify the time zone with the time zone name. 
+
+All the time zone names are available in TiDB without loading the time zone information table. The system calculates based on its own time zone rules which is generally the `tzdata` package. The calculation rules cannot be modified by loading time zone information table.
```
morgo:
```suggestion
Because they are built-in to TiDB, named time zones in TiDB may behave slightly differently to MySQL, and cannot be modified. For example, in TiDB the names are case-sensitive [#8087](https://github.com/pingcap/tidb/issues/8087).
```

dev/reference/mysql-compatibility.md
```diff
@@ -149,3 +149,23 @@ TiDB supports **all of the SQL modes** from MySQL 5.7 with minor exceptions:
...
+> 
+> The time zone rules in TiKV of two versions are as follows: 
+> 
+> - 3.0.0 RC.1 and later: [tzdata 2018i](https://github.com/eggert/tz/tree/2018i)
+> - 2.1.0 RC.1 and later: [tzdata 2018e](https://github.com/eggert/tz/tree/2018e)
+
+#### Zero month and zero day
+
+TiDB does not accept that the date value of month and day is 0. This kind of date data can be normally inserted but might not be read by certain types of SQL statements.
```
morgo: Do we have a bug report or issue to describe this? It sounds scary.

huangxiuyan: @breeswish Could you please provide a bug report for a better explanation? Thanks!

breeswish: @morgo Actually in normal SQL modes zero month and zero day are also not accepted. Additionally TiDB fully supports zero month and zero day but TiKV does not. So when this kind of date is processed by TiKV there will be errors. This is a bug that can be fixed but due to very low usage and very large work to do we decided to write it out in the documentation currently to let user know it (that we don't have plans to fix it and just treat it as a MySQL compatibility issue that you need to avoid it for now).

dev/reference/mysql-compatibility.md
breeswish: .. without the need ..

breeswish: TiDB does the date time calculation according to the system time zone rules ... which will not be modified even if time zone information table is manually loaded in the MySQL way.

dev/reference/mysql-compatibility.md
```diff
@@ -149,3 +149,23 @@ TiDB supports **all of the SQL modes** from MySQL 5.7 with minor exceptions:
...
+
+All the time zone names are available in TiDB without loading the time zone information table. The system calculates based on its own time zone rules which is generally the `tzdata` package. The calculation rules cannot be modified by loading time zone information table. 
+
+> **Warning**:
+>
+> TiKV calculates the time-related expressions received with its built-in time zone calculation rules instead of the system time zone rules. If these two kinds of time zone rules do not match with each other, it may occur that the inserted time data cannot be read again. In addition, if you install the tzdata 2018a time zone rules, when you set the time zone to Asia/Shanghai or to the server's time zone which is Asia/Shanghia, TiDB 3.0 RC.1 can be inserted to `1988-04-17 02:00:00`. But this record can not be read again by certain types of SQL statements. The reason is that the time does not exist in Asia/Shanghai time zone according to the tzdata 2018i rules on which the TiKV 3.0 based. (For example, the daylight saving time is shifted back by an hour.)
```
breeswish: In addition -> For example

breeswish: `(For example, the daylight saving time is shifted back by an hour.)` -> `(That is, the time does not exist because the DST shifts forward time by an hour.)`

dev/reference/mysql-compatibility.md
morgo: @breeswish so another way to put this, is we really, really do not recommend you unset the `NO_ZERO_IN_DATE` sql mode which is enabled by default? If you can write it but not read it correctly, that is weird.  Maybe we should add validation and refuse to unset it?

breeswish: We will respond an error explicitly that such date cannot be processed by TiKV, so user don't need to worry about hidden data integrity issues. I think we don't need to change Sql Mode because this issue only occurs in rare cases. Most date time functions are not processed by TiKV which will be fine. Even if user really meets the issue we can provide some manual workarounds.

huangxiuyan:
@morgo @breeswish  PTAL, thanks!

dev/reference/mysql-compatibility.md
morgo: (We chatted offline, and suggest that tidb add functionality to disable coprocessor execution for specific functions or statements. The tidb team already has such functionality planned.)

dev/reference/mysql-compatibility.md
```diff
@@ -149,3 +149,23 @@ TiDB supports **all of the SQL modes** from MySQL 5.7 with minor exceptions:
...
+> 
+> The named timezone rules in TiKV of two versions are as follows: 
+> 
+> - 3.0.0 RC.1 and later: [tzdata 2018i](https://github.com/eggert/tz/tree/2018i)
+> - 2.1.0 RC.1 and later: [tzdata 2018e](https://github.com/eggert/tz/tree/2018e)
+
+#### Zero month and zero day
+
+TiDB does not accept that the date value of month and day is 0. This kind of date data can be normally inserted but might not be read by certain types of SQL statements.
```
morgo: I think we should tone down this message. I still think it is important, since some users rely on a zero-day to emulate a 'month' data type, but a user can not do this with the default SQL mode.

Perhaps:

```suggestion
It is not recommended to unset the SQL modes `NO_ZERO_DATE` or `NO_ZERO_IN_DATE`, which are enabled by default in TiDB as in MySQL. While TiDB supports operating with these modes disabled, the TiKV coprocessor does not.  While inserting zero-values for month and day is supported, executing certain statements that push down date and time processing functions to TiKV may result in a statement error.
```
PTAL @breeswish

huangxiuyan:
@breeswish PTAL about the morgan's suggestion, thanks!

dev/reference/mysql-compatibility.md
breeswish: The modification of morgo Looks good to me.

dev/reference/mysql-compatibility.md
```diff
@@ -149,3 +149,23 @@ TiDB supports **all of the SQL modes** from MySQL 5.7 with minor exceptions:
         - For MySQL 5.7: `OFF`
         - For MySQL 8.0: `ON`
     
+### Date and Time
+
+#### Named timezone
+
+TiDB supports named timezones such as `America/Los_Angeles` without having to load the [time zone information tables](https://dev.mysql.com/doc/refman/8.0/en/time-zone-support.html#time-zone-installation) as in MySQL.
+
+Because they are built-in, named time zones in TiDB may behave slightly differently to MySQL, and cannot be modified. For example, in TiDB the names are case-sensitive [#8087](https://github.com/pingcap/tidb/issues/8087).
```
lilin90: Let's use "might" instead of "may" in the documentation, because "may" has multiple meanings.

```suggestion
Because they are built-in, named time zones in TiDB might behave slightly differently to MySQL, and cannot be modified. For example, in TiDB the names are case-sensitive [#8087](https://github.com/pingcap/tidb/issues/8087).
```

dev/reference/mysql-compatibility.md
```diff
@@ -149,3 +149,23 @@ TiDB supports **all of the SQL modes** from MySQL 5.7 with minor exceptions:
...
+
+> **Note:**
+>
+> TiKV calculates the time-related expressions received with its built-in named timezone calculation rules instead of the system named timezone rules. If these two kinds of named timezone rules do not match with each other, it may occur that the inserted time data cannot be read again. For example, if you install the tzdata 2018a named timezone rules, when you set the named timezone to Asia/Shanghai or to the server's timezone which is Asia/Shanghia, TiDB 3.0 RC.1 can be inserted to `1988-04-17 02:00:00`. But this record can not be read again by certain types of SQL statements. The reason is that the time does not exist in `Asia/Shanghai` named timezone according to the tzdata 2018i rules on which the TiKV 3.0 based. That is, the time does not exist because the DST shifts time forward by an hour.
```
lilin90: - may -> might
- Shanghia (2nd one) -> Shanghai (Pay attention to typos.)

```suggestion
> TiKV calculates the time-related expressions received with its built-in named timezone calculation rules instead of the system named timezone rules. If these two kinds of named timezone rules do not match with each other, it might occur that the inserted time data cannot be read again. For example, if you install the tzdata 2018a named timezone rules, when you set the named timezone to Asia/Shanghai or to the server's timezone which is Asia/Shanghai, TiDB 3.0 RC.1 can be inserted to `1988-04-17 02:00:00`. But this record cannot be read again by certain types of SQL statements. The reason is that the time does not exist in `Asia/Shanghai` named timezone according to the tzdata 2018i rules on which the TiKV 3.0 based. That is, the time does not exist because the DST shifts time forward by an hour.
```

dev/reference/mysql-compatibility.md
```diff
@@ -149,3 +149,23 @@ TiDB supports **all of the SQL modes** from MySQL 5.7 with minor exceptions:
...
+> 
+> The named timezone rules in TiKV of two versions are as follows: 
+> 
+> - 3.0.0 RC.1 and later: [tzdata 2018i](https://github.com/eggert/tz/tree/2018i)
+> - 2.1.0 RC.1 and later: [tzdata 2018e](https://github.com/eggert/tz/tree/2018e)
+
+#### Zero month and zero day
+
+It is not recommended to unset the SQL modes `NO_ZERO_DATE` or `NO_ZERO_IN_DATE`, which are enabled by default in TiDB as in MySQL. While TiDB supports operating with these modes disabled, the TiKV coprocessor does not. Executing certain statements that push down date and time processing functions to TiKV may result in a statement error.
```
lilin90:
```suggestion
It is not recommended to unset the SQL modes `NO_ZERO_DATE` or `NO_ZERO_IN_DATE`, which are enabled by default in TiDB as in MySQL. While TiDB supports operating with these modes disabled, the TiKV coprocessor does not. Executing certain statements that push down date and time processing functions to TiKV might result in a statement error.
```

huangxiuyan:
PTAL @lilin90 thanks!

lilin90:
@huangxiuyan Please pay attention to my comments. One is missed: the `can not -> cannot` in "But this record can not be read again by"

dev/reference/mysql-compatibility.md
```diff
@@ -149,3 +149,23 @@ TiDB supports **all of the SQL modes** from MySQL 5.7 with minor exceptions:
...
+
+Because they are built-in, named time zones in TiDB might behave slightly differently to MySQL, and cannot be modified. For example, in TiDB the names are case-sensitive [#8087](https://github.com/pingcap/tidb/issues/8087).
+
+> **Note:**
+>
+> TiKV calculates the time-related expressions received with its built-in named timezone calculation rules instead of the system named timezone rules. If these two kinds of named timezone rules do not match with each other, it might occur that the inserted time data cannot be read again. For example, if you install the tzdata 2018a named timezone rules, when you set the named timezone to Asia/Shanghai or to the server's timezone which is Asia/Shanghai, TiDB 3.0 RC.1 can be inserted to `1988-04-17 02:00:00`. But this record can not be read again by certain types of SQL statements. The reason is that the time does not exist in `Asia/Shanghai` named timezone according to the tzdata 2018i rules on which the TiKV 3.0 based. That is, the time does not exist because the DST shifts time forward by an hour.
```
lilin90: Do you think this is reasonable? "TiDB 3.0 RC.1 can be inserted to `1988-04-17 02:00:00`." 

A **version** cannot be inserted into `1988-04-17 02:00:00`. We need to first understand the meaning when translating it.

lilin90: The translation of this paragraph has many problems. I cannot see the meaning conveyed in the Chinese version. Please check.

```suggestion
> TiKV calculates time-related expressions that can be pushed down to it. This calculation uses the built-in time zone rule and does not depend on the time zone rule installed in the system. If the time zone rule installed in the system does not match the version of the built-in time zone rules in TiKV, the time data that can be inserted cannot be read in a few cases.
>
> For example, if the tzdata 2018a time zone rule is installed in the system, the time `1988-04-17 02:00:00` can be inserted into TiDB 3.0 RC.1 when the time zone is set to Asia/Shanghai or the time zone is set to the local time zone and the local time zone is Asia/Shanghai. But this record can no longer be read out by specific types of SQL statements because this time does not exist in the Asia/Shanghai time zone according to the tzdata 2018i time zone rule used by TiKV 3.0 RC.1. Daylight saving time is one hour late.
```

lilin90:
@breeswish PTAL again~ Thanks!

dev/reference/mysql-compatibility.md
```diff
@@ -149,3 +149,25 @@ TiDB supports **all of the SQL modes** from MySQL 5.7 with minor exceptions:
...
+
+#### Named timezone
+
+TiDB supports named timezones such as `America/Los_Angeles` without having to load the [time zone information tables](https://dev.mysql.com/doc/refman/8.0/en/time-zone-support.html#time-zone-installation) as in MySQL.
+
+Because they are built-in, named time zones in TiDB might behave slightly differently to MySQL, and cannot be modified. For example, in TiDB the names are case-sensitive [#8087](https://github.com/pingcap/tidb/issues/8087).
+
+> **Note:**
+>
+> TiKV calculates time-related expressions that can be pushed down to it. This calculation uses the built-in time zone rule and does not depend on the time zone rule installed in the system. If the time zone rule installed in the system does not match the version of the built-in time zone rule in TiKV, the time data that can be inserted cannot be read in a few cases.
```
breeswish: To be more precise and align with "Zero month and zero day", we can change all occurrences of "cannot be read" to "result in a statement error".

breeswish: The rest is cool!

dev/reference/mysql-compatibility.md
```diff
@@ -149,3 +149,25 @@ TiDB supports **all of the SQL modes** from MySQL 5.7 with minor exceptions:
...
+> 
+> For example, if the tzdata 2018a time zone rule is installed in the system, the time `1988-04-17 02:00:00` can be inserted into TiDB of the 3.0.0-rc.1 version when the time zone is set to Asia/Shanghai or the time zone is set to the local time zone and the local time zone is Asia/Shanghai. But this record can no longer be read out by specific types of SQL statements because this time does not exist in the Asia/Shanghai time zone according to the tzdata 2018i time zone rule used by TiKV 3.0.0-rc.1. Daylight saving time is one hour late.
```
breeswish: "But this record can no longer be read out..." -> "But reading this record will result in a statement error..."

PR #1161, opened 2019-05-21
dev/reference: update transaction related documents

lilin90:
Via: https://github.com/pingcap/docs-cn/pull/1340

### DIFF
--- a/dev/reference/configuration/tidb-server/tidb-specific-variables.md
+++ b/dev/reference/configuration/tidb-server/tidb-specific-variables.md
@@ -256,11 +256,13 @@ set @@global.tidb_distsql_scan_concurrency = 10
 
 - Scope: SESSION | GLOBAL
 - Default: 1
-- This variable is used to set whether to disable automatic retry of explicit transactions. If you set this variable to 1, the transaction does not retry automatically. If there is a transactional conflict, the transaction needs to be retried at the application layer.
+- This variable is used to set whether to disable automatic retry of explicit transactions. The default value of 1 means that transactions will not automatically retry in TiDB and `COMMIT` statements might return errors that need to be handled in the application layer.
+
+    Setting the value to 0 means that TiDB will automatically retry transactions, resulting in fewer errors from `COMMIT` statements. Be careful when making this change, because it might result in lost updates.
 
     This variable does not affect automatically committed implicit transactions and internally executed transactions in TiDB. The maximum retry count of these transactions is determined by the value of `tidb_retry_limit`.
 
-    To decide whether you need to disable automatic retry, see [description of optimistic transactions](/dev/reference/transactions/transaction-isolation.md#description-of-optimistic-transactions).
+    To decide whether you can enable automatic retry, see [description of optimistic transactions](/dev/reference/transactions/transaction-isolation.md#description-of-optimistic-transactions).
 
 ### tidb_back_off_weight
 
--- a/dev/reference/configuration/tidb-server/tidb-specific-variables.md
+++ b/dev/reference/configuration/tidb-server/tidb-specific-variables.md
@@ -250,13 +250,27 @@ set @@global.tidb_distsql_scan_concurrency = 10
 
 - Scope: SESSION | GLOBAL
 - Default value: 10
-- When a transaction encounters retriable errors, such as transaction conflicts and TiKV busy, this transaction can be re-executed. This variable is used to set the maximum number of the retries.
+- When a transaction encounters retriable errors (such as transaction conflicts, over slow transaction commit, or table schema changes), this transaction can be re-executed. This variable is used to set the maximum number of the retries.
 
 ### tidb_disable_txn_auto_retry
 
 - Scope: SESSION | GLOBAL
-- Default: 0
-- This variable is used to set whether to disable automatic retry of explicit transactions. If you set this variable to 1, the transaction does not retry automatically. If there is a transactional conflict, the transaction needs to be retried at the application layer. To decide whether you need to disable automatic retry, see [description of optimistic transactions](/dev/reference/transactions/transaction-isolation.md#description-of-optimistic-transactions).
+- Default: 1
+- This variable is used to set whether to disable automatic retry of explicit transactions. If you set this variable to 1, the transaction does not retry automatically. If there is a transactional conflict, the transaction needs to be retried at the application layer.
+
+    This variable does not affect automatically committed implicit transactions and internally executed transactions in TiDB. The maximum retry count of these transactions is determined by the value of `tidb_retry_limit`.
+
+    To decide whether you need to disable automatic retry, see [description of optimistic transactions](/dev/reference/transactions/transaction-isolation.md#description-of-optimistic-transactions).
+
+### tidb_back_off_weight
+
+- Scope: SESSION | GLOBAL
+- Default value: 2
+- This variable is used to increase the weight of the maximum time of TiDB `back-off`, that is, the maximum retry time for sending a retry request when an internal network or other component (TiKV, PD) failure is encountered. This variable can be used to adjust the maximum retry time and the minimum value is 1.
+
+    For example, the base timeout for TiDB to take TSO from PD is 15 seconds. When `tidb_back_off_weight = 2`, the maximum timeout for taking TSO is: *base time * 2 = 30 seconds*.
+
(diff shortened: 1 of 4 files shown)


dev/reference/transactions/overview.md
```diff
@@ -66,17 +66,18 @@ This statement is used to roll back the current transaction and cancels all the
 
 TiDB supports explicit transactions (`BEGIN/COMMIT`) and implicit transactions (`SET autocommit = 1`).
 
-If you set the value of `autocommit` to 1 and start a new transaction through `BEGIN`, the autocommit is disabled before `COMMIT`/`ROLLBACK` which makes the transaction becomes explicit.
+If you set the value of `autocommit` to 1 and start a new transaction through the `BEGIN` statement, the autocommit is disabled before `COMMIT`/`ROLLBACK` which makes the transaction becomes explicit.
```
morgo:
```suggestion
If you set the value of `autocommit` to 1 and start a new transaction through the `[BEGIN|START TRANSACTION]` statement, the autocommit is disabled before `COMMIT`/`ROLLBACK` which makes the transaction becomes explicit.
```

dev/reference/transactions/transaction-isolation.md
```diff
@@ -57,18 +57,13 @@ The MySQL Repeatable Read isolation level is not the snapshot isolation level. T
 
 ## Transaction retry
 
-Transactions that fail may automatically be retried by TiDB, which may lead to lost updates. This feature can be disabled by setting `tidb_retry_limit = 0`.
+Failed transactions might be automatically retried by TiDB, which can lead to update loss. You can enable **this feature** by setting `tidb_disable_txn_auto_retry = 0`.
```
morgo:
```suggestion
By default TiDB will not retry transactions since this may lead to lost updates. If your application can tolerate lost updates, and does not require Snapshot Isolation consistency, you can enable **this feature** by setting `tidb_disable_txn_auto_retry = 0`. This has the benefit of fewer `COMMIT` statements generating errors.
```

dev/reference/transactions/transaction-isolation.md
```diff
@@ -57,18 +57,13 @@ The MySQL Repeatable Read isolation level is not the snapshot isolation level. T
...
-...
-# The maximum number of retries when commit a transaction.
-retry-limit = 10
-```
+The value of `tidb_retry_limit` cannot be 0. Otherwise, the automatic retry is also disabled.
 
 ## Transactional anomalies caused by automatic retries
 
-Because TiDB automatically retries transactions [by default](#transaction-retry), the final result might not be as expected if the transactions created by the explicit `BEGIN` statement automatically retry after meeting a conflict.
+TiDB does not automatically retry transactions [by default](#transaction-retry). If automatic transaction retry is enabled, the final result might not be as expected if the transaction created by the explicit `BEGIN` statement automatically retries after encountering a conflict.
```
morgo: This is another case of BEGIN|START TRANSACTION. Even `set autocommit=0` will also explicitly create a transaction, so I suggest we remove it here. If it is important to the original author, we could also say "non-autocommit transactions", which would be more accurate.

```suggestion
TiDB does not automatically retry transactions [by default](#transaction-retry). If automatic transaction retry is enabled, the final result might not be as expected if the transaction automatically retries after encountering a conflict.
```

dev/reference/transactions/transaction-model.md
```diff
@@ -13,13 +13,13 @@ Similarly, functions such as `GET_LOCK()` and `RELEASE_LOCK()` and statements su
...
 >
-> On the business side, remember to check the returned results of `commit` because even there is no error in the execution, there might be errors in the `commit` process.
+> On the application side, remember to check the returned results of `COMMIT` because even there is no error in the execution, there might be errors in the `COMMIT` process.
 
 ## Differences from MySQL
 
 ### Transaction retry
 
-By default, transactions that fail may automatically be retried by TiDB, which may lead to lost updates. This feature can be disabled by setting `tidb_retry_limit = 0`.
+TiDB does **not** automatically retry failed transactions by default, because the retry can cause update loss. You can enable this feature by setting `tidb_disable_txn_auto_retry = 0`.
```
morgo:
```suggestion
While not enabled by default, TiDB can automatically retry failed transactions when `tidb_disable_txn_auto_retry = 0`. This feature is now disabled by default because retry may lead to lost updates.
```

dev/reference/configuration/tidb-server/tidb-specific-variables.md
```diff
@@ -250,13 +250,27 @@ set @@global.tidb_distsql_scan_concurrency = 10
...
-- Default: 0
-- This variable is used to set whether to disable automatic retry of explicit transactions. If you set this variable to 1, the transaction does not retry automatically. If there is a transactional conflict, the transaction needs to be retried at the application layer. To decide whether you need to disable automatic retry, see [description of optimistic transactions](/dev/reference/transactions/transaction-isolation.md#description-of-optimistic-transactions).
+- Default: 1
+- This variable is used to set whether to disable automatic retry of explicit transactions. If you set this variable to 1, the transaction does not retry automatically. If there is a transactional conflict, the transaction needs to be retried at the application layer.
+
+    This variable does not affect automatically committed implicit transactions and internally executed transactions in TiDB. The maximum retry count of these transactions is determined by the value of `tidb_retry_limit`.
+
+    To decide whether you need to disable automatic retry, see [description of optimistic transactions](/dev/reference/transactions/transaction-isolation.md#description-of-optimistic-transactions).
```
morgo:
```suggestion
    To decide whether you can enable automatic retry, see [description of optimistic transactions](/dev/reference/transactions/transaction-isolation.md#description-of-optimistic-transactions).
```

dev/reference/configuration/tidb-server/tidb-specific-variables.md
```diff
@@ -250,13 +250,27 @@ set @@global.tidb_distsql_scan_concurrency = 10
...
+- When a transaction encounters retriable errors (such as transaction conflicts, over slow transaction commit, or table schema changes), this transaction can be re-executed. This variable is used to set the maximum number of the retries.
 
 ### tidb_disable_txn_auto_retry
 
 - Scope: SESSION | GLOBAL
-- Default: 0
-- This variable is used to set whether to disable automatic retry of explicit transactions. If you set this variable to 1, the transaction does not retry automatically. If there is a transactional conflict, the transaction needs to be retried at the application layer. To decide whether you need to disable automatic retry, see [description of optimistic transactions](/dev/reference/transactions/transaction-isolation.md#description-of-optimistic-transactions).
+- Default: 1
+- This variable is used to set whether to disable automatic retry of explicit transactions. If you set this variable to 1, the transaction does not retry automatically. If there is a transactional conflict, the transaction needs to be retried at the application layer.
```
morgo:
```suggestion
- This variable is used to set whether to disable automatic retry of explicit transactions. The default value of 1 means that transactions will not automatically retry in TiDB and `COMMIT` statements may return errors that will need to be handled in the application layer.

Setting the value to 0 means that TiDB will automatically retry transactions, resulting in fewer errors from `COMMIT` statements. Be careful when making this change, since it may result in lost updates.
```

dev/reference/transactions/transaction-model.md
```diff
@@ -68,4 +68,8 @@ Due to its distributed nature, workloads that are single-threaded or latency-sen
...
+    When TiDB executes the `LOAD DATA` operation, a record with 20,000 rows of data is seen as a transaction for persistent storage by default. If one `LOAD DATA` operation inserts more than 20,000 rows, it is divided into multiple transactions to commit. If an error occurs in one transaction, this transaction in process is not committed. However, transactions before that are committed successfully. In this case, a part of the `LOAD DATA` operation is successfully inserted, and the rest of the data insertion fails. But MySQL treats a `LOAD DATA` operation as one transaction, one error leads to the failure of the entire `LOAD DATA` operation.
+
+    > **Warning:**
+    >
+    > The `LOAD DATA` operation in TiDB does not guarantee the atomicity of transactions, so it is not recommended to use it in production environment.
```
morgo: You can use it in production for loading data. You can't regularly use it once data is loaded.
```suggestion
    > The `LOAD DATA` operation in TiDB does not guarantee the atomicity of transactions, so it recommended only for initial batch loading data, and not during regular usage for production environment(s).
```

lilin90:
@morgo Thanks for your detailed suggestions! I've updated them all. PTAL again~

PR #1165, opened 2019-05-22
dev/how-to/migrate, TOC: add a doc of migrating from MariaDB

CaitinChen:
@wangkedba @lilin90 @kolbe PTAL

### DIFF
--- a/dev/how-to/migrate/from-mariadb-galera-cluster.md
+++ b/dev/how-to/migrate/from-mariadb-galera-cluster.md
@@ -63,6 +63,10 @@ This document introduces how to migrate data from MariaDB Galera Cluster to TiDB
 
 - Configure the MariaDB yum repository.
 
+    > **Note:**
+    >
+    > Currently, Data Migration supports only MariaDB 10.1.2 and later.
+
     ```
     Create mariadb.repo.
     # vi /etc/yum.repo.d/mariadb.repo
@@ -406,6 +410,10 @@ You need to install the following on all the nodes in the database.
         ```
 
     2. Edit `inventory.ini`.
+    
+        > **Note:**
+        >
+        > Add `flavor="mariadb"` when writing `invintory.ini`, because the binlog formats between MariaDB and MySQL are different.
 
         ```
         ## DM modules
@@ -561,8 +569,3 @@ You need to install the following on all the nodes in the database.
         ```
  
 For more information about dmctl basic usage, see [Manage the Data Synchronization Task](https://pingcap.com/docs/dev/reference/tools/data-migration/manage-tasks/).
- 
-## Note
-
-- Add `flavor="mariadb"` when writing `invintory.ini`, because the binlog formats between MariaDB and MySQL are different.
-- Currently, Data Migration supports only MariaDB 10.1.2 and later.
--- a/dev/how-to/migrate/from-mariadb-galera-cluster.md
+++ b/dev/how-to/migrate/from-mariadb-galera-cluster.md
@@ -80,7 +80,7 @@ This document introduces how to migrate data from MariaDB Galera Cluster to TiDB
 
     [mariadb-maxscale]
     # To use the latest stable release of MaxScale, use "latest" as the version.
-    # To use the latest beta (or stable if no current beta) release of MaxScale, use "beta" as the version.
+    # To use the latest beta (or stable if there is no beta currently) release of MaxScale, use "beta" as the version.
     name = MariaDB MaxScale
     baseurl = https://downloads.mariadb.com/MaxScale/2.3/centos/$releasever/$basearch
     gpgkey = file:///etc/pki/rpm-gpg/MariaDB-MaxScale-GPG-KEY
--- a/dev/how-to/migrate/from-mariadb-galera-cluster.md
+++ b/dev/how-to/migrate/from-mariadb-galera-cluster.md
@@ -1,0 +1,568 @@
+---
+title: Migrate Data from MariaDB Galera Cluster to TiDB
+summary: Learn to migrate data from MariaDB Galera Cluster to TiDB.
+category: how-to
+---
+
+# Migrate Data from MariaDB Galera Cluster to TiDB
+
+This document introduces how to migrate data from MariaDB Galera Cluster to TiDB. 
+
+## Hardware and system requirements
+
+### Hardware requirements
+
+- NIC bonding or teaming
+- PCIe SSDs (not SATA SSDs)
+- CPU power saving management disabled in BIOS
+- NUMA disabled in BIOS (`flush_caches` is optimized in Percona Server 5.7.10. For details, see [Improved NUMA support](https://www.percona.com/doc/percona-server/5.7/performance/innodb_numa_support.html))
+
+### Operating system requirements
+
+- Use CentOS 6.9 64-bit or CentOS 7.3-1611 64-bit as the operating system.
+
+- If you use a virtual machine, do not deploy a cluster on a single physical machine.
+
+- Use XFS as the file system.
+
+- Disable SWAP space by setting `vm.swappiness` to 0.
+
+- If the memory size exceeds 64 GB, enable huge pages.
+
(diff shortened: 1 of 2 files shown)


queenypingcap:
@kolbe PTAL. Thanks!

dev/how-to/migrate/from-mariadb-galera-cluster.md
```diff
@@ -0,0 +1,571 @@
...
+### Hardware requirements
+
+- NIC bonding or teaming
+- PCIe SSDs (not SATA SSDs)
+- CPU power saving management disabled in BIOS
+- NUMA disabled in BIOS (`flush_caches` is optimized in Percona Server 5.7.10. For details, see [Improved NUMA support](https://www.percona.com/doc/percona-server/5.7/performance/innodb_numa_support.html))
+
+### Operating system requirements
+
+- Use CentOS 6.9 64-bit or CentOS 7.3-1611 64-bit as the operating system.
```
kolbe: Why do we refer to these specific release of CentOS 6 and CentOS 7? Surely it would be best if people used the most up-to-date releases of these OSes. (Current CentOS 7 is 7.6, for example.)

dev/how-to/migrate/from-mariadb-galera-cluster.md
```diff
@@ -0,0 +1,571 @@
...
+
+- Use XFS as the file system.
+
+- Disable SWAP space by setting `vm.swappiness` to 0.
+
+- If the memory size exceeds 64 GB, enable huge pages.
+
+- If the memory exceeds 32 GB, disable NUMA in the operating system.
+
+- Disable IPv6.
```
kolbe: Is it really necessary to disable IPv6? Why?

dev/how-to/migrate/from-mariadb-galera-cluster.md
```diff
@@ -0,0 +1,571 @@
...
+    ```
+    # vim /etc/sysctl.d/99-sysctl.conf
+    net.ipv6.conf.all.disable_ipv6 = 1
+    net.ipv6.conf.default.disable_ipv6 = 1
+    ```
+
+- Enable port forwarding.
+
+    ```
+    # vim /etc/sysctl.d/99-sysctl.conf
```
kolbe: This edits the same file as the previous step. To make things easier for users, we should combine steps when possible.

dev/how-to/migrate/from-mariadb-galera-cluster.md
```diff
@@ -0,0 +1,571 @@
...
+    ```
+
+- Modify `/etc/pam.d/login`.
+
+    ```
+    # vim /etc/pam.d/login 
+    session required /lib64/security/pam_limits.so
+    ```
+
+- Configure the MariaDB yum repository.
```
kolbe: Please change this to use the mariadb_repo_setup script instead:

```
curl -sS https://downloads.mariadb.com/MariaDB/mariadb_repo_setup | sudo bash
```

dev/how-to/migrate/from-mariadb-galera-cluster.md
```diff
@@ -0,0 +1,571 @@
...
+1. Install MariaDB.
+
+    ```
+    # yum install MariaDB-server MariaDB-client galera mariabackup socat –y
+    ```
+
+2. Install `percona-toolkit`. It is a collection of advanced command-line tools, used for performance tuning, and it is an essential tool for DBAs.
+
+    ```
+    # yum install percona-toolkit –y
```
kolbe: Notice that there are 3 `yum install` commands in a row here. Why not combine them together into a single command so a user doesn't have to keep copying/pasting this stuff and waiting for yum to do its thing?

dev/how-to/migrate/from-mariadb-galera-cluster.md
```diff
@@ -0,0 +1,571 @@
...
+    ```
+
+3. Configure parameters according to your actual demand.
```
kolbe: I think we should probably remove this entire section about installing and configuring MariaDB. We should assume that someone migrating *from* MariaDB/Galera already has those components set up and running. We are not in a position to help guide someone through setup and configuration of this product.

dev/how-to/migrate/from-mariadb-galera-cluster.md
```diff
@@ -0,0 +1,571 @@
...
+    - `wsrep_local_state_commment` is `synced`.
+    - `wsrep_local_state` is 4.
+    - `swrep_cluster_status` is `primary`.
+    - Values of `wsrep_local_index`s are different.
+
+4. Log in to different nodes respectively, create testing tables, insert data, and observe the data replication status.
+
+## Migrate data to TiDB using Data Migration
+
+1. Deploy Data Migration.
```
kolbe: I don't think it makes sense to have such detailed instructions for deploying MariaDB/Galera and basically zero instructions for deploying TiDB. Presumably a user coming *from* MariaDB will already be using MariaDB but will *not* be using TiDB?

queenypingcap:
Thanks @wangkedba for the detailed steps and @kolbe for the instant and constructive feedback!
If the migration process is not so different from MySQL/MariaDB Galera to TiDB, we might just need to add a single document that gives a good overview of the different sources (Galera, Aurora) supported by DM and the differences among them. 
What do you think? @wangkedba and @GregoryIan ?

CaitinChen:
@ericsyh Please share your thoughts. Thanks. :-)

wangkedba:
If you only need to synchronize data to TiDB use DM, you can use the existing DM document, but you need to specify flavor="mariadb". The rest of the configuration is the same as mysql

ericsyh:
I think it's no need to introduce how to deploy the MariaDB cluster here, description of upstream DB arch is enough

kolbe:
It’s outside the scope of this PR, but DM should simply auto-detect the “flavor” instead of requiring the user to manually configure it.

dev/how-to/migrate/from-mariadb-galera-cluster.md
```diff
@@ -0,0 +1,571 @@
...
+    gpgcheck = 1
+    enabled = 1
+    ```
+
+- Install the following software.
+
+    ```
+    # yum install iotop sysstat -y
+    ```
+
```
ericsyh: ### MariaDB requirements
MariaDB >= 10.1.2

dev/how-to/migrate/from-mariadb-galera-cluster.md
```diff
@@ -0,0 +1,571 @@
...
+2. Configure the data replication task.
+
+    1. Configure the `task.yaml` configuration file.
+
+        Copy and edit `{ansible deploy}/conf/task.yaml.example`, and generate the following task configuration file `task.yaml`:
+
+        ```
+        ---
+        name: "test"                    # The task name. The names of different tasks that are running simultaneously cannot be the same.
+        task-mode: "incremental"        # Full + incremental (all) replication modes.
```
ericsyh: If the mode is incremental, i think it's better to be mention in the doc title.

ericsyh:
> It’s outside the scope of this PR, but DM should simply auto-detect the “flavor” instead of requiring the user to manually configure it.
> […](#)
> On May 23, 2019, at 1:46 AM, wangkedba ***@***.***> wrote: If you only need to synchronize data to TiDB use DM, you can use the existing DM document, but you need to specify flavor="mariadb". The rest of the configuration is the same as mysql — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or mute the thread.

Sounds reasonable, i will check with DM dev to remove the "flavor" config by auto-detecting the upstream db.

CaitinChen:
I think we can use the existing DM Tutorial doc and add a note about specifying flavor="mariadb".

kolbe:
I agree. I’ll work some additional text into my recently-published DM tutorial to add a couple notes about the “flavor” configuration option.

Kolbe

kolbe:
https://github.com/pingcap/docs/pull/1169 adds information about the "flavor" setting to the TiDB-DM Tutorial.

PR #1169, opened 2019-05-23
Added information about "flavor" setting to data-migration.md

kolbe:


dev/how-to/get-started/data-migration.md
```diff
@@ -227,6 +227,8 @@ password = ""
 port = 3307
 ```
 
+The `flavor` option should be set to `"mysql"` (the default) if migrating from MySQL Server, Percona Server, Percona XtraDB Cluster, or Amazon Aurora or RDS. If migrating from MariaDB Server or MariaDB (Galera) Cluster, use `flavor = "mariadb"` (only MariaDB versions greater than 10.1.2 are supported).
```
queenypingcap: How about:

- If you migrate from MySQL Server, Percona Server, Percona XtraDB Cluster, or Amazon Aurora or Amazon RDS to TiDB, use `flavor = "mysql"` , which is the default option.
- If you migrate from MariaDB Server or MariaDB Galera Cluster to TiDB, use `flavor = "mariadb"`. 
 ** Note: ** DM supports MariaDB versions later than 10.1.2 only.

kolbe: I think that's needlessly verbose and confusing, actually. Do you really object so strenuously to the single-paragraph style I used?

PR #1170, opened 2019-05-23
Changed instructions for modifying MySQL settings

kolbe:
The instructions in the syncer docs told users to dynamically change settings in MySQL that would not persist after a restart of the service. That's a bad practice, so I updated the doc to provide better instructions.

### DIFF
--- a/dev/reference/tools/syncer.md
+++ b/dev/reference/tools/syncer.md
@@ -389,52 +389,28 @@ Before replicating data using Syncer, check the following items:
 
         If the result is `log_bin = OFF`, you need to enable the binlog. See the [document about enabling the binlog](https://dev.mysql.com/doc/refman/5.7/en/replication-howto-masterbaseconfig.html).
 
-    2. Check whether the binlog format in MySQL is `ROW`.
+    2. The binlog format must be `ROW` and the binary log must be written with `FULL` row images. Check both of these variables:
 
         Check the binlog format using the following command:
 
         ```sql
-        mysql> show global variables like 'binlog_format';
-        +--------------------+----------+
-        | Variable_name      | Value    |
-        +--------------------+----------+
-        | binlog_format      | ROW      |
-        +--------------------+----------+
-        1 row in set (0.00 sec)
+        > select variable_name, variable_value from information_schema.global_variables where variable_name in ('binlog_format','binlog_row_image');
+        +------------------+----------------+
+        | variable_name    | variable_value |
+        +------------------+----------------+
+        | BINLOG_FORMAT    | ROW            |
+        | BINLOG_ROW_IMAGE | FULL           |
+        +------------------+----------------+
+        2 rows in set (0.001 sec)
         ```
 
-        - If the binlog format is not `ROW`, set it to `ROW` using the following command:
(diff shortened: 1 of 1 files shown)


dev/reference/tools/syncer.md
```diff
@@ -389,50 +389,22 @@ Before replicating data using Syncer, check the following items:
...
-
-        If the result of `binlog_row_image` is not `FULL`, set it to `FULL` using the following command:
-
-        ```sql
-        mysql> set global binlog_row_image = FULL;
-        Query OK, 0 rows affected (0.01 sec)
-        ```
+        - If one of the settings is not correct, you should change the configuration file on disk and restart the MySQL service.
+        - It's critical to persist any configuration changes to disk, so that they're reflected if the MySQL service restarts.
+        - Because existing connections will keep old values of global variables, you should *not* use the `SET` statement to dynamically chnage these settings.
```
morgo:
```suggestion
        - Because existing connections will keep old values of global variables, you should *not* use the `SET` statement to dynamically change these settings.
```

dev/reference/tools/syncer.md
```diff
@@ -389,50 +389,22 @@ Before replicating data using Syncer, check the following items:
...
-        ```
-
-        If the result of `binlog_row_image` is not `FULL`, set it to `FULL` using the following command:
-
-        ```sql
-        mysql> set global binlog_row_image = FULL;
-        Query OK, 0 rows affected (0.01 sec)
-        ```
+        - If one of the settings is not correct, you should change the configuration file on disk and restart the MySQL service.
+        - It's critical to persist any configuration changes to disk, so that they're reflected if the MySQL service restarts.
```
morgo: It's important, I don't think it's critical. Critical is something like remembering to keep copies of the keys to your encrypted database.

kolbe: What will happen if binlog_format or binlog_row_image are changed back after syncer is already running? Will it simply stop if it gets an unexpected type of data, or is there some chance the data on the replica could drift?

morgo: I am not sure, but I would consider this a loss of HA (degraded service). Critical should be saved for loss of service. You could also say 'required' if you prefer.

kolbe: I'll go with "important", thanks for the heads up!

dev/reference/tools/syncer.md
```diff
@@ -389,50 +389,22 @@ Before replicating data using Syncer, check the following items:
 
         If the result is `log_bin = OFF`, you need to enable the binlog. See the [document about enabling the binlog](https://dev.mysql.com/doc/refman/5.7/en/replication-howto-masterbaseconfig.html).
 
-    2. Check whether the binlog format in MySQL is `ROW`.
-
-        Check the binlog format using the following command:
+    2. The binlog format must be `ROW` and the binary log must be written with `FULL` row images. Check both of these variables:
```
morgo: There is some duplication, because the sentence above says:
>  If the result is `log_bin = OFF`, you need to enable the binlog.

Do you mind cleaning this up too to flow more naturally?

morgo: Nevermind, discussed via chat. Looks good.

PR #1194, opened 2019-06-05
scripts: append go/bin to $PATH

YKG:
1. After `golang.tar.gz` was unzipped, we can append a command `export PATH=...` to profile settings.

2. We can add a tip to let the developers know that they need to update env variables if they don't have Rust or Go installed before.

@siddontang PTAL


lilin90:
@aylei Could you please help review this PR? Thanks!

aylei:
Please resolve the merge conflicts

PR #1201, opened 2019-06-06
Changed drainer.toml to drainer-cluster.toml to match new filename

kolbe:
Commit fd8534d8ded60053a56ef2aeef0e999ef90ed249 to pingcap/tidb-ansible split drainer.toml into drainer-kafka.toml and drainer-cluster.toml. This PR modifies the tidb-ansible documentation to refer to drainer-cluster.toml instead of drainer.toml.


morgo:
@kolbe can you take another look at this. We version split our documents overnight, so you will have to update dev/ v3.0/ and v2.1/ as it applies.

PR #1215, opened 2019-06-12
sql/statement: add syntax description for SELECT FOR UPDATE

dcalvin:
This PR added back the Syntax description table for it contains information that is relevant in Jepsen test.  https://github.com/pingcap/docs-cn/pull/1389  PTAL @lilin90


dev/reference/sql/statements/select.md
```diff
@@ -59,6 +59,27 @@ The `SELECT` statement is used to read data from TiDB.
...
+|`FOR UPDATE` | The `SELECT FOR UPDATE` clause read-locks all the data in the result sets to detect concurrent updates from other transactions. Data that match the query conditions but do not exist in the result sets are not read-locked, such as data written by other transactions after the current transaction is started. TiDB uses the [Optimistic Transaction Model](mysql-compatibility.md#transaction). The transaction conflicts are not detected in the statement execution phase. Therefore, the current transaction will not block other transactions from executing `UPDATE`, `DELETE` or `SELECT FOR UPDATE` like databases such as PostgreSQL would do. In the committing phase, the rows read by `SELECT FOR UPDATE` will be committed in 2 phases, which means they can also join the conflict detection. If write conflicts occur, commit fails for all transactions that include the `SELECT FOR UPDATE` clause. If no conflict is detected, commit succeeds. And a new version will be generated for the locked rows, so that write conflicts can be detected when other uncommitted transactions are to be committed.|
```
lilin90: 1. Can `read-lock` be used as a verb?
2. Update the link
3. Update wording

```suggestion
| `FOR UPDATE` | The `SELECT FOR UPDATE` clause locks all the data in the result sets to detect concurrent updates from other transactions. Data that match the query conditions but do not exist in the result sets are not read-locked, such as the row data written by other transactions after the current transaction is started. TiDB uses the [Optimistic Transaction Model](/dev/reference/transactions/transaction-model.md). The transaction conflicts are not detected in the statement execution phase. Therefore, the current transaction does not block other transactions from executing `UPDATE`, `DELETE` or `SELECT FOR UPDATE` like other databases such as PostgreSQL. In the committing phase, the rows read by `SELECT FOR UPDATE` are committed in two phases, which means they can also join the conflict detection. If write conflicts occur, the commit fails for all transactions that include the `SELECT FOR UPDATE` clause. If no conflict is detected, the commit succeeds. And a new version is generated for the locked rows, so that write conflicts can be detected when other uncommitted transactions are being committed later.|
```

PR #1218, opened 2019-06-13
dev/reference: add transactions overview

morgo:
The file exists, but was not in the TOC by accident.


lilin90:
@CaitinChen FYI. Please help update it in docs-cn. Thanks!

PR #1229, opened 2019-06-17
dev/reference/security: update description about "grant" to a non-exist user

IzabelWang:
1. Update description about "grant" to a non-exist user in privilege-system.md. Via: [#1414](https://github.com/pingcap/docs-cn/pull/1414/files)
2. Fix a mistake.

PTAL @lilin90 @tiancaiamao

### DIFF
--- a/dev/reference/security/privilege-system.md
+++ b/dev/reference/security/privilege-system.md
@@ -294,10 +294,10 @@ Requires the `SUPER` privilege.
 
 The following system tables are special because all the privilege-related data is stored in them:
 
-- mysql.user (user account, global privilege)
-- mysql.db (database-level privilege)
-- mysql.tables_priv (table-level privilege)
-- mysql.columns_priv (column-level privilege; not currently supported)
+- `mysql.user` (user account, global privilege)
+- `mysql.db` (database-level privilege)
+- `mysql.tables_priv` (table-level privilege)
+- `mysql.columns_priv` (column-level privilege; not currently supported)
 
 These tables contain the effective range and privilege information of the data. For example, in the `mysql.user` table:
 
--- a/dev/reference/security/privilege-system.md
+++ b/dev/reference/security/privilege-system.md
@@ -29,9 +29,20 @@ Use the following statement to grant the `xxx` user all privileges on all databa
 GRANT ALL PRIVILEGES ON *.* TO 'xxx'@'%';
 ```
 
-If the granted user does not exist, TiDB will automatically create a user.
+If the granted user does not exist, TiDB will not automatically create a user by default. This beahavior follows `NO_AUTO_CREATE_USER` in SQL Mode.
+If `NO_AUTO_CREATE_USER` is deleted from SQL Mode, TiDB will automatically create a user when the granted user does not exist.
 
 ```sql
+mysql> select @@sql_mode;                                                                                                                                                                             +-------------------------------------------------------------------------------------------------------------------------------------------+
+| @@sql_mode                                                                                                                                |
++-------------------------------------------------------------------------------------------------------------------------------------------+
+| ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION |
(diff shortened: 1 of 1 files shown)


dev/reference/security/privilege-system.md
```diff
@@ -29,9 +29,19 @@ Use the following statement to grant the `xxx` user all privileges on all databa
 GRANT ALL PRIVILEGES ON *.* TO 'xxx'@'%';
 ```
 
-If the granted user does not exist, TiDB will automatically create a user.
+If the granted user does not exist, TiDB will not automatically create a user by default. This beahavior follows `NO_AUTO_CREATE_USER` in SQL Mode. If `NO_AUTO_CREATE_USER` is deleted from SQL Mode, TiDB will automatically create a user when the granted user does not exist.
```
morgo:
```suggestion
If the granted user does not exist, TiDB will not automatically create a user by default. This behavior can be changed by un-setting the `NO_AUTO_CREATE_USER` SQL Mode. This is **not recommended** since it presents a security risk: miss-spelling a username will result in a new user created with an empty password.
```

dev/reference/security/privilege-system.md
```diff
@@ -29,9 +29,19 @@ Use the following statement to grant the `xxx` user all privileges on all databa
...
+| @@sql_mode                                                                                                                                |
++-------------------------------------------------------------------------------------------------------------------------------------------+
+| ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION |
++-------------------------------------------------------------------------------------------------------------------------------------------+
+1 row in set (0.00 sec)
+
+mysql> set @@sql_mode='ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION';
+Query OK, 0 rows affected (0.00 sec)
+
 mysql> SELECT * FROM mysql.user WHERE user='xxxx';
```
morgo: We should make it clearer that the user doesn't exist. I suggest changing to:

```suggestion
mysql> SELECT * FROM mysql.user WHERE user='idontexist';
```
This will make the next lines also change to (note that I've also removed the `IDENTIFIED BY` to highlight the security issue. It is also not recommended to combine authentication + authorization into the same statements):

```
GRANT ALL PRIVILEGES ON test.* TO 'idontexist';
```
and:
```
SELECT user,host,password FROM mysql.user WHERE user='idontexist';
```

IzabelWang:
@morgo Thanks~

dev/reference/security/privilege-system.md
```diff
@@ -29,10 +29,20 @@ Use the following statement to grant the `xxx` user all privileges on all databa
...
++-------------------------------------------------------------------------------------------------------------------------------------------+
+1 row in set (0.00 sec)
+
+mysql> set @@sql_mode='ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION';
+Query OK, 0 rows affected (0.00 sec)
+
+mysql> SELECT * FROM mysql.user WHERE user='idontexist';
 Empty set (0.00 sec)
 
 mysql> GRANT ALL PRIVILEGES ON test.* TO 'xxxx'@'%' IDENTIFIED BY 'yyyyy';
```
morgo:
```suggestion
mysql> GRANT ALL PRIVILEGES ON test.* TO 'idontexist';
```

morgo: The lines below will also need editing: 
```
SELECT user,host FROM mysql.user WHERE user='xxxx';  (and the result)
In this example, `xxxx@%` is the user that is automatically created.
```

tiancaiamao:
LGTM

morgo:
This requires v3.0 updates as well. I don't believe v2.1 supports this sql mode, so maybe we should add a note about the security risk.

morgo:
I think this will be easier to merge and fix. I will submit a followup PR.

PR #1231, opened 2019-06-18
faq/tidb: fix mis-linked file in FAQ

dcalvin:
### What is changed, added or deleted? (required)

This PR fixed an incorrect link that pointed to a wrong place and added back the right file removed in refactoring. 

### What is the related PR or file link(s)? (required) 

https://github.com/pingcap/docs-cn/pull/1458

### Which version does your change affect? (required)

V3.0

### DIFF
--- a/dev/reference/configuration/tidb-server/server-command-option.md
+++ b/dev/reference/configuration/tidb-server/server-command-option.md
@@ -1,0 +1,233 @@
+---
+title: The TiDB Command Options
+summary: Learn about TiDB command options and configuration files.
+category: user guide
+---
+
+# The TiDB Command Options
+
+This document describes the startup options and TiDB server configuration files.
+
+## TiDB startup options
+
+When you start TiDB processes, you can specify some program options.
+
+TiDB supports a lot of startup options. Run the following command to get a brief introduction:
+
+```
+./tidb-server --help
+```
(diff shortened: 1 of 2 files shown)


lilin90:
@morgo FYI. Some links point to the content in the deleted file, so we add it back and will remove it when new files are ready.

lilin90:
Do we need to add an alias for this file? @dcalvin

dcalvin:
> Do we need to add an alias for this file? @dcalvin

We can add one in case any user has bookmarked this page.

dev/reference/configuration/tidb-server/server-command-option.md
```diff
@@ -0,0 +1,233 @@
+---
+title: The TiDB Command Options
+summary: Learn about TiDB command options and configuration files.
+category: user guide
```
dcalvin:
```suggestion
category: user guide
aliases: ['/docs/sql/server-command-option/']
```

morgo:
LGTM. If we need the file, can we add it to the TOC?

dcalvin:
> LGTM. If we need the file, can we add it to the TOC?

I think it makes sense, as it contains options that could not otherwise be found. We can add it to TiDB server configs

PR #1242, opened 2019-06-18
 reference/mysql-compatibility: Version comments

morgo:
### What is changed, added or deleted? (required)

The way TiDB parses and executes version-comments can potentially break apps. See https://github.com/pingcap/tidb/issues/10804 for example.

### What is the related PR or file link(s)? (required) 

Relates to https://github.com/pingcap/tidb/issues/10804 (but another fix will be used for this bug)

### Which version does your change affect? (required)

It applies to all versions.

### Did you use Tab spaces in this PR? (required)

No


morgo:
This applies to all versions, the PR needs to be updated.

morgo:
@kolbe PTAL again

PR #1252, opened 2019-06-24
generate_pdf.sh: stop generating PDFs for v3.0 and v2.1

YiniXu9506:

CI failed at generating PDF for v2.1 and v3.0, so remove v2.1 and v3.0 from generate_pdf.sh

### Check list 

- [x] No Tab spaces anywhere 

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -30,19 +30,15 @@ jobs:
           command: |
             sudo bash -c 'echo "222.222.95.49 uc.qbox.me" >> /etc/hosts';
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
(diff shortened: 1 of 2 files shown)


kolbe:
Shouldn't the script instead be fixed so that it's possible to generate PDFs of 2.1 and 3.0 documentation?

lilin90:
@kolbe Since we're working on differentiating content for each version, we'll add back generating PDF for the other two versions later.

lilin90:
> Shouldn't the script instead be fixed so that it's possible to generate PDFs of 2.1 and 3.0 documentation?

Fixed in #1283 by @YiniXu9506.

Issue #807, opened 2018-12-07
better document select for update


morgo:
The behavior is described here: https://github.com/pingcap/tidb/issues/6591
(I didn't realize it worked that way.)

Currently, we just say [it doesn't work the same way as MySQL](https://www.pingcap.com/docs/sql/mysql-compatibility/#transaction-model), which is correct.  But it would be nice to have a manual page with a better explanation.

PR #1260, opened 2019-06-24
reference/transactions: improve version diffs

morgo:

- Do not explain "get_lock" as different from MySQL: it doesn't work and should be on incompatibility.
- Explain in dev/ that optimistic is only the default.
- Better clarify that retry differs between 2.1 and 3.0 and could cause upgrade issues (document on both sides).
- Change dev/ mysql-compatibility for LOCK TABLES from incompatible to having an explanation.
- Document SELECT FOR UPDATE as a transaction model behavior difference

Also fixes https://github.com/pingcap/docs/issues/807

All

### Check list 

- [ ] Add a new file to `TOC.md`
- [x] No Tab spaces anywhere 


morgo:
@kolbe @jackysp @crazycs520 PTAL

(Note the versions in the documents. If I am wrong for any, please comment..)

PR #1261, opened 2019-06-24
how-to/maintain: fix broken link

morgo:

Solve the 2.1.8 log format change a little better, by including
a manual page for each for 2.1 series only.

Previously the page linked direct to GitHub (now a broken link), which should not be done.

2.1 only.

### Check list 

- [x] No Tab spaces anywhere 


morgo:
@kolbe PTAL

PR #1269, opened 2019-06-25
reference/sql/statements: update sqlgram media

morgo:

This changes sqlgram to point to version specific directories. This is really the only media that is likely to conflict between versions, so we decided to have a directory per sqlgram rather than a media directory per version.

2.1, 3.0, dev.

### Checklist 

- [ ] Add a new file to `TOC.md`
- [ ] No Tab spaces anywhere 
- [ ] Leave a blank line both before and after a code block
- [ ] Keep the first level heading consistent with `title` in metadata
- [ ] Use *four* spaces for each level of indentation except that in `TOC.md`


morgo:
@lilin90 I suggest we push this before https://github.com/pingcap/docs/pull/1267

PR #1274, opened 2019-06-26
dev, v3: add a doc of window functions

CaitinChen:

Add a doc of window functions

https://github.com/pingcap/docs-cn/pull/1437

v3.0, dev

When the content in dev is fixed, I will update that in v3.0 accordingly.

@kolbe @lamxtyler PTAL

### DIFF
--- a/dev/reference/sql/functions-and-operators/window-functions.md
+++ b/dev/reference/sql/functions-and-operators/window-functions.md
@@ -6,7 +6,7 @@ category: reference
 
 # Window Functions
 
-The usage of window functions in TiDB is similar to that in MySQL. For details, see [MySQL Window Functions](https://dev.mysql.com/doc/refman/8.0/en/window-functions.html). 
+The usage of window functions in TiDB is similar to that in MySQL 8.0. For details, see [MySQL Window Functions](https://dev.mysql.com/doc/refman/8.0/en/window-functions.html). 
 
 Because window functions may use some reserved or key words, some SQL statements that could be executed normally cannot be parsed after TiDB is upgrade. In this case, you can set `tidb_enable_window_function` (`1` by default) to `0`.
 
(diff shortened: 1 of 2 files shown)


dev/reference/sql/functions-and-operators/window-functions.md
```diff
@@ -0,0 +1,27 @@
+---
+title: Window Functions
+summary: This document introduces window functions supported in TiDB.
+category: reference
+---
+
+# Window Functions
+
+The usage of window functions in TiDB is similar to that in MySQL. For details, see [MySQL Window Functions](https://dev.mysql.com/doc/refman/8.0/en/window-functions.html).
```
morgo: Just a small suggestion. It is similar to MySQL, but not the MySQL version TiDB is compatible with:
```suggestion
The usage of window functions in TiDB is similar to that in MySQL 8.0. For details, see [MySQL Window Functions](https://dev.mysql.com/doc/refman/8.0/en/window-functions.html). 
```

dev/reference/sql/functions-and-operators/window-functions.md
```diff
@@ -0,0 +1,27 @@
...
+title: Window Functions
+summary: This document introduces window functions supported in TiDB.
+category: reference
+---
+
+# Window Functions
+
+The usage of window functions in TiDB is similar to that in MySQL. For details, see [MySQL Window Functions](https://dev.mysql.com/doc/refman/8.0/en/window-functions.html). 
+
+Because window functions may use some reserved or key words, some SQL statements that could be executed normally cannot be parsed after TiDB is upgrade. In this case, you can set `tidb_enable_window_function` (`1` by default) to `0`.
```
morgo: The reserved/keywords is not a possibility. The 'may' is if it impacts your application.

```suggestion
Because window functions reserve additional words in the parser, TiDB provides an option to disable window functions. If you receive errors parsing SQL statements after upgrading, try setting `tidb_enable_window_function=0`.
```

kolbe:
LGTM after Morgan's suggestions are accepted.

CaitinChen:
@morgo Thanks ^_^

PR #1279, opened 2019-06-27
dev, v3.0, media: add v3.0 vs v2.1 sysbench benchmark results

lilin90:
### What is changed, added or deleted? 

Add v3.0 vs v2.1 sysbench benchmark results

@jebter @jackysp PTAL

### What is the related PR or file link(s)? 

- https://github.com/pingcap/docs-cn/pull/1516
- https://github.com/pingcap/docs-cn/pull/1541

### Which version does your change affect? 

- v3.0
- dev

### Checklist 

- [x] Add a new file to `TOC.md`
- [x] No Tab spaces anywhere 
- [x] Leave a blank line both before and after a code block
- [x] Keep the first level heading consistent with `title` in metadata
- [x] Use *four* spaces for each level of indentation except that in `TOC.md`

PR #1285, opened 2019-06-27
dev, v2.1, v3.0: update the sysbench link

CaitinChen:

Update the sysbench link.

docs-cn:
- https://github.com/pingcap/docs-cn/pull/1504
- https://github.com/pingcap/docs-cn/pull/1499

docs: https://github.com/pingcap/docs/pull/1279

dev, v2.1, v3.0

@lilin90 PTAL


morgo:
I think it is better to use relative links here to the manual pages (vs. linking to GitHub). It should be better for SEO to drive everything to pingcap.com, plus redirects will work as files change.

CaitinChen:
@morgo Currently, we only maintain sysbench files on GitHub, so users can obtain more information if we use absolute links here.

CaitinChen:
@lilin90 @jackysp PTAL

CaitinChen:
@jackysp PTAL again.

PR #1289, opened 2019-06-28
*/releases, TOC: publish 3.0 GA release notes

dcalvin:
### What is changed, added or deleted? 

This PR updated the release notes for GA 3.0

### What is the related PR or file link(s)? 

https://github.com/pingcap/docs-cn/pull/1540

### Which version does your change affect? 
V3.0/Dev

### Checklist 

- [x] Add a new file to `TOC.md`
- [x] No Tab spaces anywhere 
- [x] Leave a blank line both before and after a code block
- [x] Keep the first level heading consistent with `title` in metadata
- [x] Use *four* spaces for each level of indentation except that in `TOC.md`

PR #1290, opened 2019-06-28
Add 3.0 release to 2.1 manual

morgo:

This PR forgot to add the 3.0 RN to the 2.1 manual:
https://github.com/pingcap/docs/pull/1289

We should consider scripting the parts of the manual which are identical between versions.

2.1 (other versions already handled)

### Checklist 

- [x] No Tab spaces anywhere 
- [ ] Leave a blank line both before and after a code block
- [ ] Keep the first level heading consistent with `title` in metadata
- [ ] Use *four* spaces for each level of indentation except that in `TOC.md`


morgo:
cc @dcalvin

dcalvin:
Thanks @morgo updated the same for docs-cn.

PR #1384, opened 2019-07-29
dev, v3.0/tidb-in-kubernetes: warn the user to not use multiple disks in production for now

cofyc:

mark GKE support alpha because a node will be stuck as NotReady when restarting using LVM

https://github.com/cofyc/docs/blob/mark-gke-alpha/dev/tidb-in-kubernetes/deploy/gcp-gke.md

dev,3.0

### Checklist 

- [x] No Tab spaces anywhere 

### DIFF
--- a/dev/tidb-in-kubernetes/deploy/gcp-gke.md
+++ b/dev/tidb-in-kubernetes/deploy/gcp-gke.md
@@ -10,7 +10,8 @@ This document describes how to deploy a TiDB cluster on GCP GKE with your laptop
 
   > **Notice：**
   >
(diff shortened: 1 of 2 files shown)


cofyc:
@gregwebs suggests explaining the actual issue and warn the user to not use multiple disks in production. I've updated the PR, do you think we need to remove the link from the overview docs?

aylei:
> @gregwebs suggests explaining the actual issue and warn the user to not use multiple disks in production. I've updated the PR, do you think we need to remove the link from the overview docs?

I agree

PR #1418, opened 2019-08-06
scripts: add verify links check and fix some broken links

cofyc:

This has been merged into [docs-cn repo](https://github.com/pingcap/docs-cn/pull/1693).

Add a script to verify internal links and fix some broken links under dev and v3.0 directories.

- External links are ignored in CI because these links may go down out of our control.
- Anchors are ignored
- Internal links conventions
   - Links starting with '/media/' are relative to the site root
   - Links starting with one or two dots are relative to the directory in which files reside
   - Other links are relative to the root of the version directory (e.g. v1.0, v2.0, v3.0)
 - When a file was moved, all other references are required to be updated for now, even if alias is given
   - This is recommended because of fewer redirects and better anchors support.

Why: Avoid broken links in CI phrase.

dev
v3.0

### Checklist 

- [x] No Tab spaces anywhere 

### DIFF
--- a/scripts/verify-links.sh
+++ b/scripts/verify-links.sh
@@ -1,0 +1,74 @@
+#!/bin/bash
+#
+# This script is used to verify links in markdown docs.
+#
+# - External links are ignored in CI because these links may go down out of our contorl.
+# - Anchors are ignored
+# - Internal links conventions
+#   - Links starting with '/media/' are relative to site root
+#   - Links starting with one or two dots are relative to the directory in which files reside
+#   - Other links are relative to the root of the version directory (e.g. v1.0, v2.0, v3.0)
+# - When a file was moved, all other references are required to be updated for now, even if alias are given
+#   - This is recommended because of less redirects and better anchors support.
+#
+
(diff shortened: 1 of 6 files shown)


scripts/verify-links.sh
```diff
@@ -0,0 +1,74 @@
...
+#   - Other links are relative to the root of the version directory (e.g. v1.0, v2.0, v3.0)
+# - When a file was moved, all other references are required to be updated for now, even if alias are given
+#   - This is recommended because of less redirects and better anchors support.
+#
+
+ROOT=$(unset CDPATH && cd $(dirname "${BASH_SOURCE[0]}")/.. && pwd)
+cd $ROOT
+
+if ! which markdown-link-check &>/dev/null; then
+    sudo npm install -g markdown-link-check@3.7.3
```
gregwebs: I always avoid global installs.

cofyc: yes, but the problem is this is not an NPM project, without `-g` the script will create `node_modules` directory and `package.json` file which confuses the users. On some platforms (e.g. OSX), the default global prefix is `/usr/local`, I think it's acceptable. IMO, we can relax the script for now and try to vendor `markdown-link-check` and other packages used in CI in the future.

PR #1493, opened 2019-08-30
Adds warning about scaling in tikv count leading to data loss

jlerche:
Adds note about scaling in tikv count via terraform leading to data loss

N/A

v3.0, dev


gregwebs:
We should explain why. Link to https://pingcap.com/docs/v3.0/tidb-in-kubernetes/scale-in-kubernetes/ and say that instance deletion here does not wait for rebalancing.

PR #1549, opened 2019-09-26
dev, v3.0, v2.1: add more items in loader config template

amyangfei:

loader config template is not up to date. I will update the Chinese doc after this is merged

Add more config items.

v3.0, v2.1, dev


amyangfei:
PTAL @csuzhangxc @TomShawn

dev/reference/tools/loader.md
```diff
@@ -97,6 +97,14 @@ user = "root"
 password = ""
 port = 4000
 
+# sql_mode used when loading data. If not provided or set to "@DownstreamDefault", downstream global sql_mode will be used
```
TomShawn: If what is not provided or set to "@DownstreamDefault"? downstream global sql_mode?

dev/reference/tools/loader.md
```diff
@@ -97,6 +97,14 @@ user = "root"
...
+# sql-mode = ""
+
+# max-allowed-packet sets max_allowed_packet for database connection,
```
TomShawn:
```suggestion
# max-allowed-packet sets `max_allowed_packet` for database connection.
```

dev/reference/tools/loader.md
amyangfei: addressed, PTAL

dev/reference/tools/loader.md
```diff
@@ -97,6 +97,14 @@ user = "root"
 password = ""
 port = 4000
 
+# sql_mode used when loading data. If the `sql-mode` value is not provided or set to "@DownstreamDefault", downstream global sql_mode will be used
```
TomShawn: Are `sql_mode` and `sql-mode` used interchangeably? If not, please frame `sql_mode` with back quotes.

amyangfei: `sql_mode` is used in TiDB, `sql-mode` is used in  loader config. Both of them are added back quotes wrapping

PR #1630, opened 2019-11-06
*: add best practices for highly concurrent write scenario

TomShawn:

This PR adds the best practices for highly-concurrent write scenarios.

https://github.com/pingcap/docs-cn/pull/1823
https://github.com/pingcap/docs-cn/pull/2043

dev, v3.0, v3.1


lilin90:
@TomShawn Please resolve conflicts.

dev/reference/best-practices/high-concurrency.md
```diff
@@ -0,0 +1,211 @@
...
+
+![QPS8](/media/best-practices/QPS8.png)
+
+You can see that the hotspot has been eliminated now.
+
+In this case, the table is simple. In other cases, you might also need to consider the hotspot problem of index. For more details on how to pre-split the index Region, refer to [Split Region](/dev/reference/sql/statements/split-region.md).
+
+## More complex hotspot problem
+
+If a table does not have a primary key, or the primary key is not the `Int` type and you do not want to generate a randomly distributed primary key ID, TiDB provides an implicit `_tidb_rowid` column as the row ID. If you do not use the `SHARD_ROW_ID_BITS` parameter, the values of the `_tidb_rowid` column might be monotonically increasing, which also causes hotspot. Refer to [`SHARD_ROW_ID_BITS` description](/dev/reference/configuration/tidb-server/tidb-specific-variables.md#shard_row_id_bits) for more detail.
```
lilin90:
```suggestion
If a table does not have a primary key, or the primary key is not the `Int` type and you do not want to generate a randomly distributed primary key ID, TiDB provides an implicit `_tidb_rowid` column as the row ID. Generally, when you do not use the `SHARD_ROW_ID_BITS` parameter, the values of the `_tidb_rowid` column are also monotonically increasing, which might causes hotspots too. Refer to [`SHARD_ROW_ID_BITS` description](/dev/reference/configuration/tidb-server/tidb-specific-variables.md#shard_row_id_bits) for more details.
```

dev/reference/best-practices/high-concurrency.md
```diff
@@ -0,0 +1,211 @@
...
+
+To avoid the hotspot problem in this situation, you can use `SHARD_ROW_ID_BITS` and `PRE_SPLIT_REGIONS` when creating a table. For more details about `PRE_SPLIT_REGIONS`, refer to [here](/dev/reference/sql/statements/split-region.md#pre_split_regions).
```
lilin90: Avoid `here`. Use the specific description instead.

```suggestion
To avoid the hotspot problem in this situation, use `SHARD_ROW_ID_BITS` and `PRE_SPLIT_REGIONS` when creating a table. For more details about `PRE_SPLIT_REGIONS`, refer to [Pre-split regions](/dev/reference/sql/statements/split-region.md#pre_split_regions).
```

dev/reference/best-practices/high-concurrency.md
```diff
@@ -0,0 +1,211 @@
...
+{{< copyable "sql" >}}
+
+```sql
+create table t (a int, b int) shard_row_id_bits = 4 pre_split_regions=·3;
+```
+
+- `SHARD_ROW_ID_BITS = 4` means that the values of `tidb_rowid` will be randomly distributed into 16 (16=2^4) ranges.
+- `pre_split_regions=3` means that the table will be pre-split into 8 (2^3) Regions after it is created.
+
+When data starts to be written into table t, the data is written into the pre-split 8 Regions, which avoids the hotspot problem that might be caused if there is only one Region after the table creation.
```
lilin90:
```suggestion
When data starts to be written into table `t`, the data is written into the pre-split 8 Regions, which avoids the hotspot problem that might be caused if only one Region exists after table creation.
```

dev/reference/best-practices/high-concurrency.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Parameter configuration
+
+In v2.1, the [latch mechanism](/dev/reference/configuration/tidb-server/configuration-file.md#txn-local-latches) is introduced in TiDB to identify transaction conflicts in advance in scenarios where write conflicts frequently appear. The aim is to reduce the retry of transaction commits in TiDB and TiKV caused by write conflicts. Generally, batch tasks use the data already stored in TiDB, so the write conflicts of transaction do not exist. In this situation, you can disable the latch in TiDB to reduce memory allocation for small objects:
```
lilin90:
```suggestion
In v2.1, the [latch mechanism](/dev/reference/configuration/tidb-server/configuration-file.md#txn-local-latches) is introduced in TiDB to identify transaction conflicts in advance in scenarios where write conflicts frequently appear. The aim is to reduce the retry of transaction commits in TiDB and TiKV caused by write conflicts. Generally, batch tasks use the data already stored in TiDB, so write conflicts of transactions do not occur. In this situation, you can disable the latch in TiDB to reduce memory allocation for small objects:
```

lilin90: `to reduce memory allocation for small objects`: I'm not sure about this description and the Chinese version is also confusing to me `以减少细小内存对象的分配`. @winkyao PTAL

dev/reference/best-practices/high-concurrency.md
```diff
@@ -0,0 +1,211 @@
...
+
+```sql
+SPLIT TABLE table_name [INDEX index_name] BY (value_list) [, (value_list)]
+```
+
+However, TiDB does not automatically perform this pre-split operation. The reasons are related to the data distribution in TiDB as one of its feature:
+
+![Table Region Range](/media/best-practices/table-Region-range.png)
+
+As can be seen from the image above, the key of the row data is so encoded that `rowID` is the only variable. In TiDB, `rowID` is an `Int64` integer. However, the user might not need to evenly split the `Int64` integer range to the desired number of ranges and then to distribute these ranges to different nodes, because Region split must also be based on the actual situation.
```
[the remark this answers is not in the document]
lilin90: @winkyao This sentence (`Table 行数据 key 的编码中`) is confusing to me. PTAL

dev/reference/best-practices/high-concurrency.md
```diff
@@ -0,0 +1,211 @@
...
+
+During the continuous write process, after finding that hotspot is caused on Node 1, PD evenly distributes the concentrated Leaders to other nodes. If the number of TiKV nodes is more than the number of Region replicas, TiKV will try to migrate these Regions to idle nodes. These two operations during the write process are also reflected in the PD's monitoring metrics:
+
+![QPS5](/media/best-practices/QPS5.png)
+
+After a period of continuous writes, PD automatically schedules the entire TiKV cluster to a state where pressure is evenly distributed. By that time, the capacity of the whole cluster can be fully used.
+
+In most cases, the above process of causing a hotspot is normal, which is the Region warm-up phase of database. However, you need to avoid this phase in highly-concurrent write-intensive scenarios.
+
+## Hotspot solutions
```
lilin90: It seems that we only provide one solution.
```suggestion
## Hotspot solution
```

PR #1670, opened 2019-11-26
tools: fix a usage of setting a region to tombstone

TomShawn:

Fix the usage of setting a Region to Tombstone.

https://github.com/pingcap/docs-cn/pull/2060

dev, v2.1, v3.0, v3.1


v2.1/reference/tools/tikv-control.md
```diff
@@ -148,11 +148,21 @@ Use the `compact-cluster` command to manually compact data of the whole TiKV clu
 
 The `tombstone` command is usually used in circumstances where the sync-log is not enabled, and some data written in the Raft state machine is lost caused by power down.
 
-In a TiKV instance, you can use this command to set the status of some Regions to Tombstone. Then when you restart the instance, those Regions are skipped. Those Regions need to have enough healthy replicas in other TiKV instances to be able to continue writing and reading through the Raft mechanism.
+In a TiKV instance, you can use this command to set the status of some Regions to Tombstone. Then when you restart the instance, those Regions are skipped. Those Regions need to have enough healthy replicas in other TiKV instances to be able to continue writing and reading through the Raft mechanism. For example:
```
lilin90: It's good to add a phrase to connect this para and the following code blocks, but I don't think it's an example. Instead, they are steps used to set a Region to tombstone.

v2.1/reference/tools/tikv-control.md
```diff
@@ -148,7 +148,9 @@ Use the `compact-cluster` command to manually compact data of the whole TiKV clu
 
 The `tombstone` command is usually used in circumstances where the sync-log is not enabled, and some data written in the Raft state machine is lost caused by power down.
 
-In a TiKV instance, you can use this command to set the status of some Regions to Tombstone. Then when you restart the instance, those Regions are skipped. Those Regions need to have enough healthy replicas in other TiKV instances to be able to continue writing and reading through the Raft mechanism. For example:
+In a TiKV instance, you can use this command to set the status of some Regions to Tombstone. Then when you restart the instance, those Regions are skipped. Those Regions need to have enough healthy replicas in other TiKV instances to be able to continue writing and reading through the Raft mechanism.
+
+Follow the two steps to set a Region to Tombstone:
```
lilin90: If we make it into steps, then we should make it clearer, more like steps (1. 2.). Refer to the Note below and consider taking out some content from it to make this section more like steps. You can first confirm it with the author.

dev/reference/tools/tikv-control.md
```diff
@@ -165,15 +165,31 @@ The `tombstone` command is usually used in circumstances where the sync-log is n
...
 In a TiKV instance, you can use this command to set the status of some Regions to Tombstone. Then when you restart the instance, those Regions are skipped. Those Regions need to have enough healthy replicas in other TiKV instances to be able to continue writing and reading through the Raft mechanism.
 
-```bash
-pd-ctl>> operator add remove-peer <region_id> <peer_id>
-$ tikv-ctl --db /path/to/tikv/db tombstone -p 127.0.0.1:2379 -r 2
-success!
-```
+Follow the two steps to set a Region to Tombstone:
+
+1. Take off the corresponding Peer of this Region on the machine in `pd-ctl`:
```
lilin90: Since `take off` has other meanings, how about using a clearer word?

```suggestion
1. Remove the corresponding Peer of this Region on the machine in `pd-ctl`:
```

dev/reference/tools/tikv-control.md
```diff
@@ -165,15 +165,31 @@ The `tombstone` command is usually used in circumstances where the sync-log is n
...
+
+    ```
+    success!
+    ```
 
 > **Note:**
 >
-> - This command only supports the local mode.
+> - The `tombstone` command only supports the local mode.
 > - The argument of the `-p` option specifies the PD endpoints without the `http` prefix. Specifying the PD endpoints is to query whether PD can securely switch to Tombstone. Therefore, before setting a PD instance to Tombstone, you need to take off the corresponding Peer of this Region on the machine in `pd-ctl`.
```
lilin90: Remove the content that has been changed to a step.

```suggestion
> - The argument of the `-p` option specifies the PD endpoints without the `http` prefix. Specifying the PD endpoints is to query whether PD can safely switch to Tombstone.
```

PR #1675, opened 2019-11-27
reference/performance: add the follower read doc

juliezhang1112:
Add the Follower Read doc

https://github.com/pingcap/docs-cn/pull/2019

dev, v3.1


dev/reference/performance/follower-reads.md
```diff
@@ -0,0 +1,50 @@
...
+
+In order to allow data reading in the follower without violating linearizability or affecting the transaction isolation level (Snapshot Isolation), the follower needs to use `ReadIndex` of the Raft protocol, to ensure that the read request can read the latest data that has been committed on the leader. At the TiDB level, follower reads simply sends read requests to a Region to the follower based on the load balancing policy.
+
+### Strongly consistent read
+
+When the follower processes read requests, it first uses `ReadIndex` of the Raft protocol to interact with the leader of the Region, to obtain the latest commit index of the current Raft group. After the latest commit index of the leader is applied locally to the follower, the processing of read requests starts.
```
lilin90:
```suggestion
When the follower node processes a read request, it first uses `ReadIndex` of the Raft protocol to interact with the leader of the Region, to obtain the latest commit index of the current Raft group. After the latest commit index of the leader is applied locally to the follower, the processing of a read request starts.
```

dev/reference/performance/follower-reads.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Follower selection strategy
```
lilin90:
```suggestion
### Follower replica selection strategy
```

dev/reference/performance/follower-reads.md
```diff
@@ -0,0 +1,50 @@
...
+
+Because follower reads guarantees linearizability without affecting the transaction isolation level, TiDB can apply Round Robin strategy for selecting the follower. Although TiKV can select any follower to handle any read request, considering the different replication speed among followers, if the load balancing granularity is too fine, it may cause significant fluctuation of latency. Currently, the granularity of the load balancing policy of follower reads is at the connection level. For a TiDB client connected to a specific Region, the selected follower is fixed, and is switched only when it fails or the scheduling policy is adjusted.
```
lilin90:
```suggestion
Because the follower reads feature guarantees linear consistency without affecting the Snapshot Isolation, TiDB adopts the round-robin strategy to select the follower replica. Although TiKV can select any follower replica to handle any read request, considering the different replication speed among followers, if the load balancing granularity is too fine, it might cause significant fluctuation of latency. Currently, the granularity of the follower reads load balancing policy is at the connection level. For a TiDB client connected to a specific Region, the selected follower is fixed, and is switched only when it fails or the scheduling policy is adjusted.
```

juliezhang1112:
PTAL @lilin90  Thanks~

dev/reference/performance/follower-reads.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Follower Reads
+summary: This document describes the use and implementation of Follower Reads.
+category: reference
+---
+
+# Follower Reads
```
zz-jason: how about `Follower Read`?

dev/reference/performance/follower-reads.md
```diff
@@ -0,0 +1,50 @@
...
+
+When a read hotspot appears in a Region, the Region leader can become a read bottleneck for the entire system. In this situation, enabling follower reads can significantly reduce the load of the leader, and improve the throughput of the whole system by balancing the load among multiple followers. This document introduces the use and implementation mechanism of follower reads.
```
zz-jason:
```suggestion
When a read hotspot appears in a Region, the Region leader can become a read bottleneck for the entire system. In this situation, enabling the Follower Read feature can significantly reduce the load of the leader, and improve the throughput of the whole system by balancing the load among multiple followers. This document introduces the use and implementation mechanism of the Follower Read feature.
```

dev/reference/performance/follower-reads.md
lilin90: @zz-jason Also a choice. Then we can keep the file name of Chinese version and English version consistent.

lilin90:
@zz-jason All comments addressed. PTAL again~ Thanks!

dev/reference/performance/follower-read.md
```diff
@@ -0,0 +1,50 @@
...
+Default: leader
+
+This variable is used to set the data read mode expected by the current session.
+
+- When the value of `tidb_replica_read` is set to `leader` or an empty string, TiDB maintains its original behavior and sends all read operations to the leader replica to perform.
+- When the value of `tidb_replica_read` is set to `follower`, TiDB selects a follower replica of the Region to perform all read operations.
+
+## Implementation mechanism
+
+Before the Follower Read feature was introduced, TiDB applied the strong leader policy and submitted all read and write requests to the leader node of a Region to handle. Although TiKV can distribute Regions evenly on multiple physical nodes, for each Region, only the leader can provide external services. The other two followers can do nothing to handle read requests but receive the data replicated from the leader at all times and prepare for voting to elect a leader in case of a failover.
```
zz-jason:
```suggestion
Before the Follower Read feature was introduced, TiDB applied the strong leader policy and submitted all read and write requests to the leader node of a Region to handle. Although TiKV can distribute Regions evenly on multiple physical nodes, for each Region, only the leader can provide external services. The other followers can do nothing to handle read requests but receive the data replicated from the leader at all times and prepare for voting to elect a leader in case of a failover.
```

I removed `two`, because in a TiDB cluster, users can config more than 3 replicas for a Region.

lilin90: @zz-jason Got it.

PR #2015, opened 2020-03-16
reference: add doc for expression index

ran-huang:

Add document for expression index.

- [x] master (the latest development version, including v4.0 changes for now)

**If you select two or more versions from above**, to trigger the bot to cherry-pick this PR to your desired release version branch(es), you **must** add corresponding labels such as **needs-cherry-pick-3.1**, **needs-cherry-pick-3.0**, and **needs-cherry-pick-2.1**.

- This PR is translated from: https://github.com/pingcap/docs-cn/pull/2312
- Other reference link(s):

### DIFF
--- a/reference/sql/statements/create-index.md
+++ b/reference/sql/statements/create-index.md
@@ -93,7 +93,7 @@ Take the following query as an example:
 SELECT * FROM t WHERE lower(name) = "pingcap";
 ```
 
-If the following expression index is established, you can use the index to speed up the query above:
+If the following expression index is built, you can use the index to speed up the above query:
 
 {{< copyable "sql" >}}
 
@@ -101,13 +101,13 @@ If the following expression index is established, you can use the index to speed
 CREATE INDEX idx ON t ((lower(name)));
 ```
 
-The cost of maintaining an expression index is higher than that of other indexes, because the value of the expression needs to be calculated whenever a row is inserted or updated. Since the value of the expression is already stored in the index, it does not require recalculating when the optimizer selects the expression index.
+The cost of maintaining an expression index is higher than that of maintaining other indexes, because the value of the expression needs to be calculated whenever a row is inserted or updated. The value of the expression is already stored in the index, so this value does not require recalculation when the optimizer selects the expression index.
 
 Therefore, when the query speed outweighs the insert and update speed, you can consider indexing the expressions.
 
-Expression indexes have the same syntax and limitations as in MySQL. They are implemented by building indexes on generated virtual columns that are invisible, so the supported expressions inherit all [limitations of virtual generated columns](/reference/sql/generated-columns.md#Limitations).
+Expression indexes have the same syntax and limitations as in MySQL. They are implemented by building indexes on generated virtual columns that are invisible, so the supported expressions inherit all [limitations of virtual generated columns](/reference/sql/generated-columns.md#limitations).
 
-Currently, the optimizer can only use the indexed expressions when the expressions are in the `FIELD` clause, `WHERE` clause, and `ORDER BY` clause. The `GROUP BY` clause will be supported in the future updates.
+Currently, the optimizer can use the indexed expressions when the expressions are only in the `FIELD` clause, `WHERE` clause, and `ORDER BY` clause. The `GROUP BY` clause will be supported in future updates.
 
 ## Associated session variables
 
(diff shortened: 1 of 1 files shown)


reference/sql/statements/create-index.md
```diff
@@ -81,6 +81,34 @@ mysql> CREATE UNIQUE INDEX c1 ON t1 (c1);
...
+
+Take the following query as an example:
+
+{{< copyable "sql" >}}
+
+```sql
+SELECT * FROM t WHERE lower(name) = "pingcap";
+```
+
+If the following expression index is established, you can use the index to speed up the query above:
```
TomShawn: To make the wording consistent with the previous paragraph.
```suggestion
If the following expression index is built, you can use the index to speed up the above query:
```

reference/sql/statements/create-index.md
```diff
@@ -81,6 +81,34 @@ mysql> CREATE UNIQUE INDEX c1 ON t1 (c1);
...
+
+{{< copyable "sql" >}}
+
+```sql
+CREATE INDEX idx ON t ((lower(name)));
+```
+
+The cost of maintaining an expression index is higher than that of other indexes, because the value of the expression needs to be calculated whenever a row is inserted or updated. Since the value of the expression is already stored in the index, it does not require recalculating when the optimizer selects the expression index.
```
TomShawn:
```suggestion
The cost of maintaining an expression index is higher than that of maintaining other indexes, because the value of the expression needs to be calculated whenever a row is inserted or updated. The value of the expression is already stored in the index, so this value does not require recalculation when the optimizer selects the expression index.
```

reference/sql/statements/create-index.md
```diff
@@ -81,6 +81,34 @@ mysql> CREATE UNIQUE INDEX c1 ON t1 (c1);
...
+
+Therefore, when the query speed outweighs the insert and update speed, you can consider indexing the expressions.
+
+Expression indexes have the same syntax and limitations as in MySQL. They are implemented by building indexes on generated virtual columns that are invisible, so the supported expressions inherit all [limitations of virtual generated columns](/reference/sql/generated-columns.md#Limitations).
```
TomShawn:
```suggestion
Expression indexes have the same syntax and limitations as in MySQL. They are implemented by building indexes on generated virtual columns that are invisible, so the supported expressions inherit all [limitations of virtual generated columns](/reference/sql/generated-columns.md#limitations).
```

reference/sql/statements/create-index.md
```diff
@@ -81,6 +81,34 @@ mysql> CREATE UNIQUE INDEX c1 ON t1 (c1);
...
+
+Currently, the optimizer can only use the indexed expressions when the expressions are in the `FIELD` clause, `WHERE` clause, and `ORDER BY` clause. The `GROUP BY` clause will be supported in the future updates.
```
TomShawn:
```suggestion
Currently, the optimizer can use the indexed expressions when the expressions are only in the `FIELD` clause, `WHERE` clause, and `ORDER BY` clause. The `GROUP BY` clause will be supported in future updates.
```

TomShawn:
LGTM
@wjhuang2016 PTAL

ran-huang:
Comments addressed. PTAL again~ @TomShawn

reference/sql/statements/create-index.md
```diff
@@ -81,6 +81,34 @@ mysql> CREATE UNIQUE INDEX c1 ON t1 (c1);
...
+
+{{< copyable "sql" >}}
+
+```sql
+CREATE INDEX idx ON t ((lower(name)));
+```
+
+The cost of maintaining an expression index is higher than that of maintaining other indexes, because the value of the expression needs to be calculated whenever a row is inserted or updated. The value of the expression is already stored in the index, so this value does not require recalculation when the optimizer selects the expression index.
+
+Therefore, when the query speed outweighs the insert and update speed, you can consider indexing the expressions.
```
wjhuang2016: I think 'performance' is better than 'speed'.

ran-huang: Done. Thanks!

ran-huang:
PTAL again~ @wjhuang2016 thanks!

ran-huang:
@TomShawn Could you please approve this PR?

PR #2154, opened 2020-04-03
tools/br: explain why we need to copy SSTs file when restoring from local

kennytm:

BR requires copying SSTs to every machine before restore, but did not explain why. Some customers are confused by such requirements, while others simply skipped it. 

This PR added the reason behind.

- [x] master (the latest development version, including v4.0 changes for now)

- This PR is translated from:
- Other reference link(s):


kennytm:
cc @GregoryIan

reference/tools/br/br.md
```diff
@@ -262,7 +262,14 @@ To restore the cluster data, use the `br restore` command. You can add the `full
...
 >
-> If the backed up cluster does not have a network storage, before the restoration, copy the backup SST files to the path specified by `--storage` on each TiKV node.
+> If you use the local storage, you **must** copy all back up SST files to every TiKV node in the path specified by `--storage`.
+>
+> Even if each TiKV node eventually only need to read a part of the all SST files, they all need full access to the complete archive because:
+>
+> * Data are replicated into multiple peers. When ingesting SSTs, these files have to be present on *all* peers. This is unlike back up where reading from a single node is enough.
+> * Where each peer is scattered to during restore is random. We don't know in advance which node will read which file.
+>
+> These can be avoided using shared storage, e.g. mounting an NFS on the local path, or using S3. With network storage, every node can automatically read every SST files, so these caveats no longer apply.
```
ran-huang, line 272:
```suggestion
> These can be avoided using shared storage, e.g. mounting an NFS on the local path, or using S3. With network storage, every node can automatically read every SST file, so these caveats no longer apply.
```

kennytm:
PTAL @3pointer

PR #2252, opened 2020-04-13
follower read: add new option for tidb_replica_read

hicqu:

Add an new option for tidb_replica_read.

https://github.com/pingcap/tidb/pull/15721

- This PR is translated from:
- Other reference link(s):


reference/performance/follower-read.md
```diff
@@ -34,6 +34,7 @@ This variable is used to set the data read mode expected by the current session.
 
 - When the value of `tidb_replica_read` is set to `leader` or an empty string, TiDB maintains its original behavior and sends all read operations to the leader replica to perform.
 - When the value of `tidb_replica_read` is set to `follower`, TiDB selects a follower replica of the Region to perform all read operations.
+- When the value of `tidb_replica_read` is set to `leader-and-follower`, TiDB can select any replicas to perform read operations.
```
zhangjinpeng1987, line 37: Not include learner replica

hicqu: Learners are included. In fact `follower` also includes them.

zhangjinpeng1987: leader-and-follower also can read learner?

hicqu: Both `leader-and-follower` and `follower` can read learner.

TomShawn:
@hicqu Please specify all the affected document versions, thanks!

reference/performance/follower-read.md
zhangjinpeng1987: How about if a learner doesn't have enough data? For example, the replica is in balancing status.

hicqu: If it's applying snapshots, new proposals will be dropped and TiDB clients will switch to a next peer.

shenli: can we optimize it latter?  For example, peer could report its status to pd, and TiDB will not choose the peer who is applying a snapshot (or not ready).

hicqu: Good idea. Seems it's possible.

PR #2327, opened 2020-04-19
reference/performance: add a use case for SPM evolve explanation

ChristianLeow:

Add an example to make SPM evolve easier to understand.

- This PR is translated from: https://github.com/pingcap/docs-cn/pull/2391
- Other reference link(s):


reference/performance/execution-plan-bind.md
```diff
@@ -99,7 +99,41 @@ set global tidb_evolve_plan_baselines = on;
 
 The default value of `tidb_evolve_plan_baselines` is `off`.
 
-After the automatic binding evolution feature is enabled, if the optimal execution plan selected by the optimizer is not among the bound execution plans, the optimizer marks the plan as an execution plan that waits for verification. Every `bind-info-lease` (the default value is `3s`), an execution plan to be verified is selected and compared with a bound execution plan with the least cost, in terms of the actual execution time. If the plan to be verified is better, it is marked as a usable binding.
+> **Note:**
+>
+> The global variable doesn't take effect in the current session. It only takes effect in a newly created session. To enable automatic binding evolution in the current session, change the key word `global` to `session`.
```
TomShawn, line 104:
```suggestion
> The global variable does not take effect in the current session. It only takes effect in a newly created session. To enable automatic binding evolution in the current session, change the key word `global` to `session`.
```

reference/performance/execution-plan-bind.md
```diff
@@ -99,7 +99,41 @@ set global tidb_evolve_plan_baselines = on;
...
+
+After the automatic binding evolution feature is enabled, if the optimal execution plan selected by the optimizer is not among the bound execution plans, the optimizer marks the plan as an execution plan that waits for verification. Every `bind-info-lease` (the default value is `3s`), an execution plan to be verified is selected and compared with a bound execution plan with the least cost, in terms of the actual execution time. If the plan to be verified is better, it is marked as a usable binding. The following use case describes the preceding process.
```
TomShawn, line 106:
```suggestion
After the automatic binding evolution feature is enabled, if the optimal execution plan selected by the optimizer is not among the bound execution plans, the optimizer marks the plan as an execution plan that waits for verification. Every `bind-info-lease` (the default value is `3s`) interval, an execution plan to be verified is selected and compared with the bound execution plan with the least cost in terms of the actual execution time. If the plan to be verified performs better in the execution time, this plan is marked as a usable binding. The following use case describes the above process.
```

reference/performance/execution-plan-bind.md
```diff
@@ -99,7 +99,41 @@ set global tidb_evolve_plan_baselines = on;
...
+
+Perform the following query on table `t`:
+
+{{< copyable "sql" >}}
+
+```sql
+select * from t where a < 100 and b < 100;
+```
+
+Rows in the table that satisfy condition `a < 100` are rare. However, for some reasons, the optimizer mistakenly chooses the full table scan instead of the optimal execution plan that uses index `a`. Users can use the following statement to create a binding first:
```
TomShawn, line 124:
```suggestion
In the table defined above, few rows meet the `a < 100` condition. However, for some reason, the optimizer chooses the full table scan instead of the optimal execution plan that uses index `a`. You can first use the following statement to create a binding:
```

reference/performance/execution-plan-bind.md
```diff
@@ -99,7 +99,41 @@ set global tidb_evolve_plan_baselines = on;
...
+
+{{< copyable "sql" >}}
+
+```sql
+create global binding for select * from t where a < 100 and b < 100 using select * from t use index(a) where a < 100 and b < 100;
+```
+
+When the previous query is executed again, the optimizer chooses index `a` under the interference of the binding created previously to reduce the query time.
```
TomShawn, line 132:
```suggestion
When the above query is executed again, the optimizer chooses index `a` under the interference of the binding created above to reduce the query time.
```

reference/performance/execution-plan-bind.md
```diff
@@ -99,7 +99,41 @@ set global tidb_evolve_plan_baselines = on;
...
+
+Assuming that as insertions and deletions are performed on the table, the number of rows that satisfy condition `a < 100` grows while the number of rows that satisfy condition `b < 100` not, still using index `a` under the binding's interference might no longer be optimal.
```
TomShawn, line 134:
```suggestion
Assuming that as insertions and deletions are performed on table `t`, an increasing number of rows meet the `a < 100` condition and a decreasing number of rows meet the `b < 100` condition. At this time, using index `a` under the binding's interference might no longer be the optimal plan.
```

reference/performance/execution-plan-bind.md
```diff
@@ -99,7 +99,41 @@ set global tidb_evolve_plan_baselines = on;
...
+
+The binding evolution can address this kind of problems. When the optimizer senses changes in table data, it generates an execution plan against the query that uses index `b`. However, because of the binding's presence, this query plan is not adopted and executed. Instead, it is stored in the backend evolution lists. During the evolution process, if it is verified that the time spent on executing the query plan is significantly shorter than that using index `a` (that is the current binding's execution plan), index `b` is added into available binding lists. After this process, when the query is executed again, the optimizer first generates the execution plan that uses index `b` and makes sure that it is in the binding lists. Then the optimizer adopts and executes it to reduce the query time after data updates.
```
TomShawn, line 136:
```suggestion
The binding evolution can address this kind of issues. When the optimizer recognizes data change in a table, it generates an execution plan for the query that uses index `b`. However, because the binding of the current plan exists, this query plan is not adopted and executed. Instead, this plan is stored in the backend evolution list. During the evolution process, if this plan is verified to have an obviously shorter execution time than that of the current execution plan that uses index `a`, index `b` is added into the available binding list. After this process, when the query is executed again, the optimizer first generates the execution plan that uses index `b` and makes sure that this plan is in the binding list. Then the optimizer adopts and executes this plan to reduce the query time after data updates.
```

TomShawn:
LGTM
@eurekaka PTAL

reference/performance/execution-plan-bind.md
```diff
@@ -99,7 +99,41 @@ set global tidb_evolve_plan_baselines = on;
...
+> **Note:**
+>
+> The global variable does not take effect in the current session. It only takes effect in a newly created session. To enable automatic binding evolution in the current session, change the keyword `global` to `session`.
+
+After the automatic binding evolution feature is enabled, if the optimal execution plan selected by the optimizer is not among the binding execution plans, the optimizer marks the plan as an execution plan that waits for verification. At every `bind-info-lease` (the default value is `3s`) interval, an execution plan to be verified is selected and compared with the binding execution plan that has the least cost in terms of the actual execution time. If the plan to be verified has shorter execution time, this plan is marked as a usable binding. The following example describes the process above.
+
+Let table `t` be defined as:
```
TomShawn, line 108:
```suggestion
Assume that table `t` is defined as follows:
```

PR #2369, opened 2020-04-22
reference: update execution plan bind doc

TomShawn:

- This PR is translated from: https://github.com/pingcap/docs-cn/pull/2608
- Other reference link(s):

### DIFF
--- a/reference/performance/execution-plan-bind.md
+++ b/reference/performance/execution-plan-bind.md
@@ -20,7 +20,7 @@ CREATE [GLOBAL | SESSION] BINDING FOR SelectStmt USING SelectStmt
 
 This statement binds SQL execution plans at the GLOBAL or SESSION level. The default scope is SESSION. The bound SQL statement is parameterized and stored in the system table. When a SQL query is processed, as long as the parameterized SQL statement and a bound one in the system table are consistent and the system variable `tidb_use_plan_baselines` is set to `on` (the default value is `on`), the corresponding optimizer hint is available. If multiple execution plans are available, the optimizer chooses to bind the plan with the least cost.
 
-When a SQL statement has bound execution plans in both GLOBAL and SESSION scopes, because the optimizer ignores the bound execution plan in the GLOBAL scope when it encounters the SESSION binding, the bound execution plan of this statement in the SESSION scope overrides the execution plan in the GLOBAL scope.
+When a SQL statement has bound execution plans in both GLOBAL and SESSION scopes, because the optimizer ignores the bound execution plan in the GLOBAL scope when it encounters the SESSION binding, the bound execution plan of this statement in the SESSION scope shields the execution plan in the GLOBAL scope.
 
 For example:
 
@@ -82,7 +82,7 @@ DROP [GLOBAL | SESSION] BINDING FOR SelectStmt
 
 This statement removes a specified execution plan binding at the GLOBAL or SESSION level. The default scope is SESSION.
 
-Generally, the binding in the SESSION scope is mainly used for test or in special situations. For a binding to take effect in all TiDB processes, you need to use the GLOBAL binding. The SESSION binding overrides the GLOBAL binding until the end of the SESSION.
+Generally, the binding in the SESSION scope is mainly used for test or in special situations. For a binding to take effect in all TiDB processes, you need to use the GLOBAL binding. The SESSION binding shields the GLOBAL binding until the end of the SESSION.
 
 In the following example, the SESSION binding is dropped.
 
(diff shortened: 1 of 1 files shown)


TomShawn:
@lawyerphx PTAL

reference/performance/execution-plan-bind.md
```diff
@@ -56,6 +82,20 @@ DROP [GLOBAL | SESSION] BINDING FOR SelectStmt
 
 This statement removes a specified execution plan binding at the GLOBAL or SESSION level. The default scope is SESSION.
 
+Generally, the binding in the SESSION scope is mainly used for test or in special situations. For a binding to take effect in all TiDB processes, you need to use the GLOBAL binding. The SESSION binding overrides the GLOBAL binding until the end of the SESSION.
+
+In the following example, the SESSION binding is dropped.
```
ran-huang, line 87: @lawyerphx Please confirm L87 means the same thing as in the Chinese version.

TomShawn:
> `The SESSION binding overrides the GLOBAL binding until the end of the SESSION.`
> I think we need to emphasize shielding instead of overwriting, because, from the user's perspective, the bindings in session scope disappear after being dropped.
> And, it may help users to understand it by changing that word or emphasizing such shields will not disappear after drop.

Thanks. `override` is replaced with `shield`.

reference/performance/execution-plan-bind.md
```diff
@@ -56,6 +82,20 @@ DROP [GLOBAL | SESSION] BINDING FOR SelectStmt
 
 This statement removes a specified execution plan binding at the GLOBAL or SESSION level. The default scope is SESSION.
 
+Generally, the binding in the SESSION scope is mainly used for test or in special situations. For a binding to take effect in all TiDB processes, you need to use the GLOBAL binding. The SESSION binding overrides the GLOBAL binding until the end of the SESSION.
```
lawyerphx: This may sound clearer.
```suggestion
Generally, the binding in the SESSION scope is mainly used for test or in special situations. For a binding to take effect in all TiDB processes, you need to use the GLOBAL binding.  A created SESSION binding will overrides the corresponding GLOBAL binding until the end of the SESSION, even if the SESSION binding is dropped before the session closes. (in this case, no binding will take effect and the plan will be selected by the optimizer).
```

reference/performance/execution-plan-bind.md
TomShawn: I was thinking about not to translate `承接上面关于 SESSION 绑定屏蔽 GLOBAL 绑定的例子，继续执行` literally because the two examples seem too far apart from each other and I can neither use an anchor. Now I translate the sentence literally, which might seem a second-best option. PTAL again.

reference/performance/execution-plan-bind.md
lawyerphx: Shield can be used to replace overrides. rest LGTM

TomShawn: Thanks! Suggestion applied.

reference/performance/execution-plan-bind.md
ran-huang: Well done!

reference/performance/execution-plan-bind.md
```diff
@@ -56,6 +82,22 @@ DROP [GLOBAL | SESSION] BINDING FOR SelectStmt
 
 This statement removes a specified execution plan binding at the GLOBAL or SESSION level. The default scope is SESSION.
 
+Generally, the binding in the SESSION scope is mainly used for test or in special situations. For a binding to take effect in all TiDB processes, you need to use the GLOBAL binding.  A created SESSION binding shields the corresponding GLOBAL binding until the end of the SESSION, even if the SESSION binding is dropped before the session closes. In this case, no binding takes effect and the plan is selected by the optimizer.
```
ran-huang, line 85:
```suggestion
Generally, the binding in the SESSION scope is mainly used for test or in special situations. For a binding to take effect in all TiDB processes, you need to use the GLOBAL binding. A created SESSION binding shields the corresponding GLOBAL binding until the end of the SESSION, even if the SESSION binding is dropped before the session closes. In this case, no binding takes effect and the plan is selected by the optimizer.
```

reference/performance/execution-plan-bind.md
```diff
@@ -56,6 +82,22 @@ DROP [GLOBAL | SESSION] BINDING FOR SelectStmt
...
+
+In the following example, the SESSION binding is dropped.
+
+The following example is based on the example above in which the SESSION binding shields the GLOBAL binding:
```
ran-huang, line 89: We should retain only one of L87 and L89.

TomShawn:
@ran-huang Comments addressed, PTAL again, thanks!

PR #3089, opened 2020-06-30
add topn-limit-push-down

SE-Bin:
add topn-limit-push-down

**Note:** CI fails because this doc links to a new file in this unmerged PR https://github.com/pingcap/docs/pull/3043.

- [x] v2.1 (TiDB 2.1 versions)

- This PR is translated from: [docs-cn/pull/3307 ](https://github.com/pingcap/docs-cn/pull/3307 )
- Other reference link(s):


topn-limit-push-down.md
```diff
@@ -0,0 +1,124 @@
...
++----------------------------+----------+-----------+---------------+--------------------------------+
+| TopN_7                     | 10.00    | root      |               | test.t.a, offset:0, count:10   |
+| └─TableReader_15           | 10.00    | root      |               | data:TopN_14                   |
+|   └─TopN_14                | 10.00    | cop[tikv] |               | test.t.a, offset:0, count:10   |
+|     └─TableFullScan_13     | 10000.00 | cop[tikv] | table:t       | keep order:false, stats:pseudo |
++----------------------------+----------+-----------+---------------+--------------------------------+
+4 rows in set (0.00 sec)
+```
+
+In this query, the TopN operator node was pushed down to TiKV for data filtering, and each Coprocessor returns only 10 records to TiDB. After TiDB integrates the data, the final filtering is performed.
```
ran-huang, line 39:
```suggestion
In this query, the TopN operator node is pushed down to TiKV for data filtering, and each Coprocessor returns only 10 records to TiDB. After TiDB aggregates the data, the final filtering is performed.
```

topn-limit-push-down.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Example 2: TopN can be pushed down into the Join (sorting only depends on the columns in the outer table)
```
ran-huang, line 41:
```suggestion
### Example 2: TopN can be pushed down into Join (sorting only depends on the columns in the outer table)
```

topn-limit-push-down.md
```diff
@@ -0,0 +1,124 @@
...
+|   │ └─TableReader_26             | 10.00    | root      |               | data:TopN_25                                    |
+|   │   └─TopN_25                  | 10.00    | cop[tikv] |               | test.t.a, offset:0, count:10                    |
+|   │     └─TableFullScan_24       | 10000.00 | cop[tikv] | table:t       | keep order:false, stats:pseudo                  |
+|   └─TableReader_30(Probe)        | 10000.00 | root      |               | data:TableFullScan_29                           |
+|     └─TableFullScan_29           | 10000.00 | cop[tikv] | table:s       | keep order:false, stats:pseudo                  |
++----------------------------------+----------+-----------+---------------+-------------------------------------------------+
+8 rows in set (0.01 sec)
+```
+
+In this query, the sorting of the TopN operator only depends on the columns in the outer table t, so TopN can be pushed down before Join, to reduce the computational of Join. Besides, TiDB also pushed TopN down to the storage layer.
```
ran-huang, line 67:
```suggestion
In this query, the collation of the TopN operator only depends on the columns in the outer table `t`, so a calculation can be performed before pushing down TopN to Join, to reduce the calculation cost of the Join operation. Besides, TiDB also pushes TopN down to the storage layer.
```
We usually translate `排序规则` as `collation`~

topn-limit-push-down.md
```diff
@@ -0,0 +1,124 @@
...
+| └─HashJoin_16                 | 12500.00 | root      |               | inner join, equal:[eq(test.t.a, test.s.a)] |
+|   ├─TableReader_21(Build)     | 10000.00 | root      |               | data:TableFullScan_20                      |
+|   │ └─TableFullScan_20        | 10000.00 | cop[tikv] | table:s       | keep order:false, stats:pseudo             |
+|   └─TableReader_19(Probe)     | 10000.00 | root      |               | data:TableFullScan_18                      |
+|     └─TableFullScan_18        | 10000.00 | cop[tikv] | table:t       | keep order:false, stats:pseudo             |
++-------------------------------+----------+-----------+---------------+--------------------------------------------+
+6 rows in set (0.00 sec)
+```
+
+TopN can't be pushed down into the Inner Join. Taking the query above as an example, if you get 100 records after Join, then do TopN can leave 10 records. If the remaining 10 records are filtered before TopN, there may be 5 left after the Join is done. It will make results differences.
```
ran-huang, line 93:
```suggestion
TopN cannot be pushed down before `Inner Join`. Taking the query above as an example, if you get 100 records after Join, then you can have 10 records left after TopN. However, if TopN is performed first to get 10 records, only 5 records are left after Join. The pushdown might result in different results. 
```
Please check if this is technically correct.

topn-limit-push-down.md
```diff
@@ -0,0 +1,124 @@
...
+
+That is the reason why opN can't be pushed down into the Outer Join for the Inner table. TopN can not be pushed down when sorting by multiple columns on multiple tables. such as `t.a+s.a`.Only if the TopN exclusively depends on sorting outer table columns, it can be pushed down.
```
ran-huang, line 95:
```suggestion
Similarly, TopN can neither be pushed down into the Outer Join for the inner table, nor can it be pushed down when its collation is related to columns on multiple tables, such as `t.a+s.a`. Only when the collation of TopN exclusively depends on columns on the outer table, can it be pushed down. 
```

topn-limit-push-down.md
```diff
@@ -0,0 +1,124 @@
...
+|     └─TableFullScan_18        | 10000.00 | cop[tikv] | table:t       | keep order:false, stats:pseudo             |
++-------------------------------+----------+-----------+---------------+--------------------------------------------+
+6 rows in set (0.00 sec)
+```
+
+TopN can't be pushed down into the Inner Join. Taking the query above as an example, if you get 100 records after Join, then do TopN can leave 10 records. If the remaining 10 records are filtered before TopN, there may be 5 left after the Join is done. It will make results differences.
+
+That is the reason why opN can't be pushed down into the Outer Join for the Inner table. TopN can not be pushed down when sorting by multiple columns on multiple tables. such as `t.a+s.a`.Only if the TopN exclusively depends on sorting outer table columns, it can be pushed down. 
+
+### Example 4: Conversion of TopN to Limit
```
ran-huang, line 97:
```suggestion
### Example 4: Convert TopN to Limit
```

topn-limit-push-down.md
```diff
@@ -0,0 +1,124 @@
...
+|   │   └─Limit_30                 | 10.00    | cop[tikv] |               | offset:0, count:10                              |
+|   │     └─TableFullScan_29       | 10.00    | cop[tikv] | table:t       | keep order:true, stats:pseudo                   |
+|   └─TableReader_35(Probe)        | 10000.00 | root      |               | data:TableFullScan_34                           |
+|     └─TableFullScan_34           | 10000.00 | cop[tikv] | table:s       | keep order:false, stats:pseudo                  |
++----------------------------------+----------+-----------+---------------+-------------------------------------------------+
+8 rows in set (0.00 sec)
+
+```
+
+In the above query, TopN is first pushed to the outer table t. Because TopN needs to sort by `t.id` which is the primary key,  it can be directly readout in order  (`keep order: true`) without extra sorting in TopN, and Simplified to Limit.
```
ran-huang, line 124:
```suggestion
In the query above, TopN is first pushed to the outer table `t`. TopN needs to sort by `t.id`, which is the primary key and can be directly read in order  (`keep order: true`) without extra sorting in TopN. Therefore, TopN is simplied as Limit.
```

ran-huang:
> It's better to replace `collation` with `ordering rule` to avoid misunderstandings with the collation of sql.

Is there any ref on this one? Like, how would you choose between `ordering rule` and `sorting rule`?

topn-limit-push-down.md
```diff
@@ -0,0 +1,127 @@
+---
+title: TopN and Limit Operator Pushdown
+summary: Learn the implementation of TopN and Limit operator pushdown.
+category: performance
```
TomShawn, line 4:
```suggestion
```

LittleFall: > > It's better to replace `collation` with `ordering rule` to avoid misunderstandings with the collation of sql.
> 
> Is there any ref on this one? Like, how would you choose between `ordering rule` and `sorting rule`?

No, `collation` generally refers to the collation of the character set. You can choose from other options that you think are more appropriate.

LittleFall:
> > It's better to replace `collation` with `ordering rule` to avoid misunderstandings with the collation of sql.
> 
> Is there any ref on this one? Like, how would you choose between `ordering rule` and `sorting rule`?

No, collation generally refers to the collation of the character set. You can choose from other options that you think are more appropriate.

ran-huang:
@LittleFall ok, thanks. Let's go with `sorting rules`, as `sort` introduces less ambiguity than `order`.
I'll update this PR and then request your review again 😀

PR #3332, opened 2020-07-20
perf-tuning: update statistics.md

toutdesuite:

update statistics.md
**`在[索引的选择](/index-choose.md)章节中，提到了 TiDB 会使用统计信息来决定选择哪个索引。` has not been translated because `index-choose.md` does not exist in the docs repo.**

- This PR is translated from: https://github.com/pingcap/docs-cn/pull/3223
- Other reference link(s):


statistics.md
```diff
@@ -6,10 +6,26 @@ aliases: ['/docs/dev/statistics/','/docs/dev/reference/performance/statistics/']
...
+
+Here "equal-depth" means that the number of values ​​falling into each bucket is as equal as possible. For example, for a given set {1.6, 1.9, 1.9, 2.0, 2.4, 2.6, 2.7, 2.7, 2.8, 2.9, 3.4, 3.5}, you want to generate 4 buckets. The equal-depth histogram is as follows. It contains four buckets [1.6, 1.9], [2.0, 2.6], [2.7, 2.8], [2.9, 3.5]. The bucket depth is 3.
+
+![Equal-depth Histogram Example](/media/statistics-1.png)
+
+For details about the parameter that determines the upper limit to the number of histogram buckets, refer to [Manual Collection](#manual-collection). When the number of buckets is larger, the accuracy of the histogram is higher; however, higher accuracy is at the cost of the usage of memory resources. You can adjust this number appropriately according to the actual scenario.
+
+## Count-Min Sketch
+
+Count-Min Sketch is a hash structure. When an equivalent query contains `a = 1` or `IN` query (for example, `a in (1, 2, 3)`), TiDB uses this data structure for estimation.
```
ran-huang, line 23:
```suggestion
Count-Min Sketch is a hash structure. When an equivalence query contains `a = 1` or `IN` query (for example, `a in (1, 2, 3)`), TiDB uses this data structure for estimation.
```

IMO, `equivalent query` refers to sql statements that takes the same effect, but `equivalence query` refers to a query that contains a `=` (and the like) as its condition. What do you think?

statistics.md
```diff
@@ -6,10 +6,26 @@ aliases: ['/docs/dev/statistics/','/docs/dev/reference/performance/statistics/']
...
+
+- Modify the `WITH NUM TOPN` parameter. TiDB stores the high-frequency (top x) data separately, with the other data stored in Count-Min Sketch. Therefore, to prevent a larger value and a smaller value from being hashed together, you can increase the value of `WITH NUM TOPN`. In TiDB, its default value is 20. The maximum value is 1024. For more information about this parameter, see [Full Collection](#full-collection).
+- Modify two parameters `WITH NUM CMSKETCH DEPTH` and `WITH NUM CMSKETCH WIDTH`. Both affect the number of hash buckets and the collision probability. You can increase the values of the two parameters appropriately according to the actual scenario to reduce the probability of hash collision, but at the cost of higher memory usage of statistical information. In TiDB, the default value of `DEPTH` is 5, and the default value of `WIDTH` is 2048. For more information about the two parameters, see [Full Collection](#full-collection).
```
ran-huang, line 28:
```suggestion
- Modify two parameters `WITH NUM CMSKETCH DEPTH` and `WITH NUM CMSKETCH WIDTH`. Both affect the number of hash buckets and the collision probability. You can increase the values of the two parameters appropriately according to the actual scenario to reduce the probability of hash collision, but at the cost of higher memory usage of statistics. In TiDB, the default value of `WITH NUM CMSKETCH DEPTH` is 5, and the default value of `WITH NUM CMSKETCH WIDTH` is 2048. For more information about the two parameters, see [Full Collection](#full-collection).
```

statistics.md
toutdesuite: Confirmed in https://logicalread.com/oracle-11g-hash-indexes-mc02/#.XxVFRUczZPY and https://docs.microsoft.com/en-us/dotnet/csharp/linq/, `equivalence query` is the right expression

PR #3386, opened 2020-07-22
perf-tuning: add document for partition pruning

Joyinqin:

- Add document for logical plan optimization rule: partition pruning.
- Translate the updated documentation according to Chinese file: [分区裁剪](https://docs.pingcap.com/zh/tidb/stable/partition-pruning#%E5%88%86%E5%8C%BA%E8%A3%81%E5%89%AA)

- This PR is translated from: https://github.com/pingcap/docs-cn/pull/3184
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases
- [ ] Might cause conflicts


partition-pruning.md
```diff
@@ -0,0 +1,259 @@
+# Partition pruning
+
+Partition pruning is a performance optimization only when the target table is a partitioned table. It analyzes the filter conditions in the query statements, selects only the partitions that may meet the conditions, and does not scan the partitions that do not match, thereby significantly reducing the amount of calculated data.
```
nullnotnil, line 3: - The optimization applies only to partitioned tables vs. the current language could imply it's always there and its an optimization when tables are partitioned/a negative in other cases.
- It avoids accessing the tables. There are other accesses besides scans - although TiDB calls what I would call a seek a scan, it is easier not to go into this detail here.
- Whether or not this is significant depends a lot on the query. If there are 10 partitions and pruning only eliminates 1 partition, it is unlikely to be significant.

```suggestion
Partition pruning is a performance optimization that applies to partitioned tables. It analyzes the filter conditions in query statements, and eliminates (_prunes_) partitions from consideration when they do not contain any data that will be required. By eliminating the non-required partitions, TiDB is able to reduce the amount of data that needs to be accessed and potentially significantly improving query execution times.
```

partition-pruning.md
```diff
@@ -0,0 +1,259 @@
+# Partition pruning
```
nullnotnil, line 1: Please include the header metadata. For h1 titles, I believe we also capitalize the whole title:

```suggestion
# Partition Pruning
```

partition-pruning.md
```diff
@@ -0,0 +1,259 @@
...
+
+Partition pruning is a performance optimization only when the target table is a partitioned table. It analyzes the filter conditions in the query statements, selects only the partitions that may meet the conditions, and does not scan the partitions that do not match, thereby significantly reducing the amount of calculated data.
+
```
nullnotnil, line 4: Before going into application scenarios, is it possible to give a very simple example of partition pruning applying?
i.e.
```sql
CREATE TABLE t1 (
 id INT NOT NULL PRIMARY KEY,
 pad VARCHAR(100)
)
PARTITION BY RANGE COLUMNS(id) (
 PARTITION p0 VALUES LESS THAN (100),
 PARTITION p1 VALUES LESS THAN (200),
 PARTITION p2 VALUES LESS THAN (MAXVALUE)
);

INSERT INTO t1 VALUES (1, 'test1'),(101, 'test2'), (201, 'test3');
EXPLAIN SELECT * FROM t1 WHERE id BETWEEN 80 AND 120;

...

mysql> EXPLAIN SELECT * FROM t1 WHERE id BETWEEN 80 AND 120;
+----------------------------+---------+-----------+------------------------+------------------------------------------------+
| id                         | estRows | task      | access object          | operator info                                  |
+----------------------------+---------+-----------+------------------------+------------------------------------------------+
| PartitionUnion_8           | 80.00   | root      |                        |                                                |
| ├─TableReader_10           | 40.00   | root      |                        | data:TableRangeScan_9                          |
| │ └─TableRangeScan_9       | 40.00   | cop[tikv] | table:t1, partition:p0 | range:[80,120], keep order:false, stats:pseudo |
| └─TableReader_12           | 40.00   | root      |                        | data:TableRangeScan_11                         |
|   └─TableRangeScan_11      | 40.00   | cop[tikv] | table:t1, partition:p1 | range:[80,120], keep order:false, stats:pseudo |
+----------------------------+---------+-----------+------------------------+------------------------------------------------+
5 rows in set (0.00 sec)
```

Joyinqin: @wshwsh12 PTAL

wshwsh12: Good idea.

partition-pruning.md
```diff
@@ -0,0 +1,296 @@
+---
+title: Partition Pruning
+summary：Introduce application scenarios of TiDB partition pruning.
```
lilin90, line 3: Please do not use Chinese punctuation in an English doc, which might cause website build failure.

lilin90:
@wshwsh12 Please have a tech review. Thanks!

wshwsh12:
lgtm

lilin90:
> lgtm

@wshwsh12 Thanks for your tech review~

PR #3838, opened 2020-09-04
stmt-reference, sysvars: Document dml-batch-size

nullnotnil:

Part of https://github.com/pingcap/docs/issues/3155

Only relevant to master.

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases
- [ ] Might cause conflicts

### DIFF
--- a/sql-statements/sql-statement-load-data.md
+++ b/sql-statements/sql-statement-load-data.md
@@ -104,13 +104,16 @@ In the above example, `x'2c'` is the hexadecimal representation of the `,` chara
 
 ## MySQL compatibility
 
-* TiDB will by default commit every 20 000 rows. This behavior is similar to MySQL NDB Cluster, but not the default configuration with the InnoDB storage engine.
+This statement is understood to be fully compatible with MySQL. Any compatibility differences should be [reported via an issue](https://github.com/pingcap/tidb/issues/new/choose) on GitHub.
 
 > **Note:**
 >
-> Committing through splitting a transaction is at the expense of breaking the atomicity and isolation of the transaction. When performing this operation, you must ensure that there are **no other** ongoing operations on the table. When an error occurs, **manual intervention is required to check the consistency and integrity of the data**. Therefore, it is not recommended to use `LOAD DATA` on any tables which are actively being read from or written to.
+> In earlier releases of TiDB, `LOAD DATA` committed every 20 000 rows. By default, TiDB now commits all rows in one transaction. This can result in the error `ERROR 8004 (HY000) at line 1: Transaction is too large, size: 100000058` after upgrading from TiDB 4.0 or earlier.
+>
+> The recommended way to resolve this error is to increase the `txn-total-size-limit` in your tidb.toml file. If you are unable to increase this limit, you can also restore the previous behavior by setting [`tidb_dml_batch_size`](/system-variables.md#tidb_dml_batch_size) to `20000`.
 
 ## See also
 
 * [INSERT](/sql-statements/sql-statement-insert.md)
 * [Import Example Database](/import-example-data.md)
+* [TiDB Lightning](/tidb-lightning/tidb-lightning-overview.md)
--- a/sql-statements/sql-statement-load-data.md
+++ b/sql-statements/sql-statement-load-data.md
@@ -108,7 +108,7 @@ This statement is understood to be fully compatible with MySQL. Any compatibilit
 
 > **Note:**
 >
-> In earlier releases of TiDB, `LOAD DATA` committed every 20 000 rows. By default, TiDB now commits all rows in one transaction. This can result in the error `ERROR 8004 (HY000) at line 1: Transaction is too large, size: 100000058` after upgrading from TiDB 4.0 or earlier.
+> In earlier releases of TiDB, `LOAD DATA` committed every 20000 rows. By default, TiDB now commits all rows in one transaction. This can result in the error `ERROR 8004 (HY000) at line 1: Transaction is too large, size: 100000058` after upgrading from TiDB 4.0 or earlier versions.
 >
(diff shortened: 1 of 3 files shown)


nullnotnil:
It looks like this was removed in https://github.com/pingcap/docs/pull/1806 , but it needs to be added back because of the [upgrade issue mentioned here](https://github.com/pingcap/docs/issues/3155#issuecomment-687253114) :(

jackysp:
lgtm

cfzjywxk:
LGTM

sql-statements/sql-statement-load-data.md
```diff
@@ -104,13 +104,16 @@ In the above example, `x'2c'` is the hexadecimal representation of the `,` chara
 
 ## MySQL compatibility
 
-* TiDB will by default commit every 20 000 rows. This behavior is similar to MySQL NDB Cluster, but not the default configuration with the InnoDB storage engine.
+This statement is understood to be fully compatible with MySQL. Any compatibility differences should be [reported via an issue](https://github.com/pingcap/tidb/issues/new/choose) on GitHub.
 
 > **Note:**
 >
-> Committing through splitting a transaction is at the expense of breaking the atomicity and isolation of the transaction. When performing this operation, you must ensure that there are **no other** ongoing operations on the table. When an error occurs, **manual intervention is required to check the consistency and integrity of the data**. Therefore, it is not recommended to use `LOAD DATA` on any tables which are actively being read from or written to.
+> In earlier releases of TiDB, `LOAD DATA` committed every 20 000 rows. By default, TiDB now commits all rows in one transaction. This can result in the error `ERROR 8004 (HY000) at line 1: Transaction is too large, size: 100000058` after upgrading from TiDB 4.0 or earlier.
```
TomShawn, line 111:
```suggestion
> In earlier releases of TiDB, `LOAD DATA` committed every 20,000 rows. By default, TiDB now commits all rows in one transaction. This can result in the error `ERROR 8004 (HY000) at line 1: Transaction is too large, size: 100000058` after upgrading from TiDB 4.0 or earlier versions.
```

system-variables.md
```diff
@@ -271,6 +271,14 @@ Constraint checking is always performed in place for pessimistic transactions (d
 - Use a bigger value in OLAP scenarios, and a smaller value in OLTP scenarios.
 - For OLAP scenarios, the maximum value cannot exceed the number of CPU cores of all the TiKV nodes.
 
+### tidb_dml_batch_size
+
+- Scope: SESSION
+- Default value: 0
+- Example value: 20000
+- When greater than `0`, TiDB will batch commit statements such as `INSERT` or `LOAD DATA` into smaller transactions. This reduces memory usage and helps ensure that the `txn-total-size-limit` is not reached by bulk modifications.
```
TomShawn, line 279:
```suggestion
- When this value is greater than `0`, TiDB will batch commit statements such as `INSERT` or `LOAD DATA` into smaller transactions. This reduces memory usage and helps ensure that the `txn-total-size-limit` is not reached by bulk modifications.
```

sql-statements/sql-statement-load-data.md
```diff
@@ -104,13 +104,16 @@ In the above example, `x'2c'` is the hexadecimal representation of the `,` chara
...
+>
+> The recommended way to resolve this error is to increase the `txn-total-size-limit` in your tidb.toml file. If you are unable to increase this limit, you can also restore the previous behavior by setting [`tidb_dml_batch_size`](/system-variables.md#tidb_dml_batch_size) to `20000`.
```
TomShawn, line 113:
```suggestion
> The recommended way to resolve this error is to increase the `txn-total-size-limit` value in your `tidb.toml` file. If you are unable to increase this limit, you can also restore the previous behavior by setting [`tidb_dml_batch_size`](/system-variables.md#tidb_dml_batch_size) to `20000`.
```

sql-statements/sql-statement-load-data.md
TomShawn, line 111: Do you mean ``` In earlier releases of TiDB, `LOAD DATA` committed 20,000 rows of data in a transaction ```?

nullnotnil: Europeans use `,` as a decimal separator. I've used `20000` instead to avoid ambiguity.

nullnotnil: It is every 20K rows. So if you are importing 101K rows, it would have been broken into 6 commits. The new behavior is always one commit. Only this is "correct" according to ACID semantics.

TomShawn (approved):
LGTM

ti-srebot (approved):

system-variables.md
```diff
@@ -271,6 +271,14 @@ Constraint checking is always performed in place for pessimistic transactions (d
 - Use a bigger value in OLAP scenarios, and a smaller value in OLTP scenarios.
 - For OLAP scenarios, the maximum value cannot exceed the number of CPU cores of all the TiKV nodes.
 
+### tidb_dml_batch_size
+
+- Scope: SESSION
+- Default value: 0
+- Example value: 20000
```
TomShawn, line 278: @nullnotnil Could you explain what `example value` is? I find it confusing when reviewing the translation of this PR. From the context, `20000` does not appear in this document but in `sql-statement-load-data.md `. Did you mean the `example` in `sql-statement-load-data.md `?

nullnotnil: I picked `20000` as an example value because this is the prior batch size for `LOAD DATA` before it was changed to not be batched in master.

I think it is potentially useful for other sysvars, because the min/max values may be a very wide range.

TomShawn: Can we say `In earlier releases of TiDB, the default batch size for LOAD DATA was 20000`? Here the `example` seems a little confusing.

nullnotnil: Let's just remove it. I don't think people will know to look up this value when they get an error, they will look up the reference for `LOAD DATA`, which mentions the previous default.

TomShawn: Removed in https://github.com/pingcap/docs/pull/3977.

PR #3922, opened 2020-09-17
explain: Add joins

nullnotnil:

Fixes https://github.com/pingcap/tidb/issues/18576 

This is a followup PR to #3858 - but they can be reviewed independently, and merge out of order.

This PR does not add this page to the TOC. I will only do that once all PRs are merged (this is one ~3-4 followup PRs).

The content here is based on/duplicates [query-execution-plan](https://docs.pingcap.com/tidb/stable/query-execution-plan). I will merge all of the replacement pages, and then update query-execution-plan (soon to be called explain-overview) as a last step, and in the same PR which adds this page to the TOC.

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases
- [ ] Might cause conflicts


nullnotnil:
I have improved the examples to be reproducible. I would like to expand this to explain more about hash joins, but I hit a bug on https://github.com/pingcap/tidb/issues/20085 -- I will take another look when there is an answer on this.

explain-joins.md
```diff
@@ -0,0 +1,219 @@
...
+6 rows in set (0.00 sec)
+```
+
+Index join is efficient in memory usage, but may be slower to execute than other join methods when a large number of probe operations are required. Consider also the following query:
+
+```sql
+SELECT * FROM t1 INNER JOIN t2 ON t1.id=t2.t1_id WHERE t1.pad1 = 'value' and t2.pad1='value';
+```
+
+On an inner join, TiDB will apply join reordering and might access either `t1` or `t2` first. Assuming that TiDB selects `t1` as the first table to apply the `Build` step, it will be able to filter on the predicate `t1.col = 'value'` before probing the table `t2`. The filter for the predicate `t2.col='value'` will be applied on each probe of table `t2`, which may be less efficient than other join methods.
```
SunRunAway, line 69: Could we provide a query example showing how we tune Hash Join into Index Join by adding a suitable index for better performance?

explain-joins.md
```diff
@@ -0,0 +1,219 @@
+---
+title: Explain Statements Using Joins
+summary: Learn about the execution plan information returned by the `EXPLAIN` statement in TiDB.
+---
+
+# Explain Statements Using Joins
```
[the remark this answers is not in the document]
nullnotnil: I try to if it is relevant in context. The TOC uses the following flow:

- Understanding the Query Execution Plan
- SQL Optimization Process (aka understanding more on how it arrives at plans)
- Control Execution Plans (aka now that you know how the optimizations work, here's how to override them when they are in error)

explain-joins.md
```diff
@@ -0,0 +1,219 @@
...
+INSERT INTO t2 SELECT NULL, a.id, RANDOM_BYTES(1024), RANDOM_BYTES(1024), RANDOM_BYTES(1024) FROM t1 a JOIN t1 b JOIN t1 c LIMIT 10000;
+INSERT INTO t2 SELECT NULL, a.id, RANDOM_BYTES(1024), RANDOM_BYTES(1024), RANDOM_BYTES(1024) FROM t1 a JOIN t1 b JOIN t1 c LIMIT 10000;
+INSERT INTO t2 SELECT NULL, a.id, RANDOM_BYTES(1024), RANDOM_BYTES(1024), RANDOM_BYTES(1024) FROM t1 a JOIN t1 b JOIN t1 c LIMIT 10000;
+SELECT SLEEP(1);
+ANALYZE TABLE t1, t2;
+```
+
+## Index Join (Index Nested Loop Join)
+
+If the estimated rows that need to be joined is small (typically less than 10000 rows), using the Index Join method is preferable. This method of join works similar to the primary method of join used in MySQL. In the following example, the operator `IndexReader_31(Build)` first reads the table `t1`. For each row that matches, TiDB will probe the table `t2`:
```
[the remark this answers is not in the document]
nullnotnil: Good suggestions, I'll add them. The `IndexReader_31` must have dropped when I rebuilt the examples on more sample data, I'll fix it.

explain-joins.md
```diff
@@ -0,0 +1,219 @@
...
+INSERT INTO t2 SELECT NULL, a.id, RANDOM_BYTES(1024), RANDOM_BYTES(1024), RANDOM_BYTES(1024) FROM t1 a JOIN t1 b JOIN t1 c LIMIT 10000;
+INSERT INTO t2 SELECT NULL, a.id, RANDOM_BYTES(1024), RANDOM_BYTES(1024), RANDOM_BYTES(1024) FROM t1 a JOIN t1 b JOIN t1 c LIMIT 10000;
+INSERT INTO t2 SELECT NULL, a.id, RANDOM_BYTES(1024), RANDOM_BYTES(1024), RANDOM_BYTES(1024) FROM t1 a JOIN t1 b JOIN t1 c LIMIT 10000;
+INSERT INTO t2 SELECT NULL, a.id, RANDOM_BYTES(1024), RANDOM_BYTES(1024), RANDOM_BYTES(1024) FROM t1 a JOIN t1 b JOIN t1 c LIMIT 10000;
+INSERT INTO t2 SELECT NULL, a.id, RANDOM_BYTES(1024), RANDOM_BYTES(1024), RANDOM_BYTES(1024) FROM t1 a JOIN t1 b JOIN t1 c LIMIT 10000;
+SELECT SLEEP(1);
+ANALYZE TABLE t1, t2;
+```
+
+## Index Join (Index Nested Loop Join)
```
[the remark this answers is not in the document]
nullnotnil: Yes we should. I didn't understand this, so I left it from the original document - but I will merge them in.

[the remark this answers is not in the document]
nullnotnil: I have added this. @SunRunAway can you review https://github.com/pingcap/docs/commit/1db44f166802ab66711a50782f3bdc2a77e743a1 and make sure it is correct?

SunRunAway:
Should the related variables be involved? like `tidb_hash_join_concurrency`, `tidb_index_join_batch_size`, `tidb_index_lookup_join_concurrency`

nullnotnil:
> Should the related variables be involved? like `tidb_hash_join_concurrency`, `tidb_index_join_batch_size`, `tidb_index_lookup_join_concurrency`

@SunRunAway I have added this in https://github.com/pingcap/docs/pull/3922/commits/ad7c61d83744b291425dccfc797d542aabb45fc5 PTAL. Thx :-)

explain-joins.md
```diff
@@ -0,0 +1,267 @@
...
+3. Read the data at the `Probe` side.
+4. Use the data of the `Probe` side to probe the Hash Table.
+5. Return qualified data to the user.
+
+The operator info column in the `explain` table also records other information about `Hash Join`, including whether the query is Inner Join or Outer Join, and what are the conditions of Join. In the above example, the query is an Inner Join, where the Join condition `equal:[eq(test.t1.id, test.t2.id)]` partly corresponds with the query statement `WHERE t1.id = t2.id`. The operator info of the other Join operators in the following examples is similar to this one.
+
+### Runtime Statistics
+
+If [`tidb_mem_quota_query`](/system-variables.md#tidb_mem_quota_query) (default: 1GB) is exceeded, TiDB will attempt to use temporary storage provided that `oom-use-tmp-storage=TRUE` (default). This means that the hash table used as part of the hash join is created on disk. Runtime statistics, such as memory usage are visible in the `execution info` of `EXPLAIN ANALYZE`. The following example shows the output of `EXPLAIN ANALYZE` with a 1GB (default) `tidb_mem_quota_query` quota, and a 500MB quota. At 500MB the hash table spills to disk:
```
zz-jason, line 197: > This means that the hash table used as part of the hash join is created on disk.

@SunRunAway please confirm: is the build side data cached in memory spilled to disk or the hash table itself is spilled to disk?

nullnotnil: I think we should update https://docs.pingcap.com/tidb/stable/system-variables#tidb_mem_quota_query as well so its a bit clearer what this does.

SunRunAway: For now, it is that build side data cached in memory spilled to disk, but we should also consider the spilling for hash table in the future.

zz-jason: It's better to clearly describe the current behavior on the document. we can update the document once we spilled hash table to disk.

BTW, for technical reasons, it's not recommended to spill hash table to disk. There should be other methods to achieve the same goal.

nullnotnil: I have attempted to clarify this in https://github.com/pingcap/docs/pull/3922/commits/bcb16dd1551c17a9e16abd5fefa027cbd4505778 - PTAL, thx :-)

explain-joins.md
```diff
@@ -0,0 +1,267 @@
...
+An index join using the hint [`INL_JOIN`](/optimizer-hints.md#inl_joint1_name--tl_name-) will create a hash table of the intermediate results before joining on the outer table. TiDB also supports creating a hash table on the outer table with the hint [`INL_HASH_JOIN`](/optimizer-hints.md#inl_hash_join). If the column sets on the inner table match the columns of the outer table, the [`INL_MERGE_JOIN`](/optimizer-hints.md#inl_merge_join) index join can apply. Each of these variations of Index Join will be automatically selected by the SQL Optimizer.
+
+### Configuration
+
+Index Join performance is influenced by the following system variables:
+
+* [`tidb_index_join_batch_size`](/system-variables.md#tidb_index_join_batch_size) (default: 25000) - the batch size of index lookup operations.
+* [`tidb_index_lookup_join_concurrency`](/system-variables.md#tidb_index_lookup_join_concurrency) (default: 4) - the number of concurrent index lookup tasks.
+
+## Hash Join
```
SunRunAway, line 160: Should `Hash Join` be above on `Index Join` because the example in `Index Join` relies on `Hash Join`?

explain-joins.md
```diff
@@ -0,0 +1,267 @@
...
+If the index on `t1.int_col` is not available, the hash join execution plan remains the same. However, the plan changes for the Index Join and performance is reduced:
+
+{{< copyable "sql" >}}
+
+```sql
+ALTER TABLE t2 DROP INDEX t1_id; # drop the index
+EXPLAIN ANALYZE SELECT /*+ INL_JOIN(t1, t2) */  * FROM t1 INNER JOIN t2 ON t1.id = t2.t1_id WHERE t1.int_col = 1;
+EXPLAIN ANALYZE SELECT /*+ HASH_JOIN(t1, t2) */  * FROM t1 INNER JOIN t2 ON t1.id = t2.t1_id WHERE t1.int_col = 1;
+ALTER TABLE t2 ADD INDEX (t1_id); # Re-add the index
+```
```
SunRunAway, line 117: Could we provide such an example? So this can guide users to add an index to speed up their Joins.
```sql
--- a table without index `t1_id`
EXPLAIN ANALYZE SELECT * FROM t1 INNER JOIN t2 ON t1.id = t2.t1_id WHERE t1.int_col = 1; --- run Join without hint
--- and get a slow Hash Join result
ALTER TABLE t1 ADD INDEX  t1_id (id);
EXPLAIN ANALYZE SELECT * FROM t1 INNER JOIN t2 ON t1.id = t2.t1_id WHERE t1.int_col = 1; --- run Join without hint
--- and get a fast Index Join result
```

explain-joins.md
nullnotnil: I think because MySQL only had index join until recently, it is better to have index join first. It even says in the second sentence "This method of join works similar to the primary method of join used in MySQL.".

There are two sentences which may be hard to parse:

> Index Join is effective if the build side is small and the probe side is pre-indexed and large. Consider the following query where the index join performs much better than the hash join:

I could reword this and say "where the index join performs well".

> If the index on `t1.int_col` is not available, the hash join execution plan remains the same. However, the plan changes for the Index Join and performance is reduced:

This is in reference to an example, where it can be seen the first example uses INL_JOIN and the second uses HASH_JOIN. I think it is understandable in context that hash join will be explained after.

explain-joins.md
nullnotnil: See https://github.com/pingcap/docs/pull/3922/commits/d3df1326fc80c128bb78ce302119467d04e7ff0a

SunRunAway (commented):
LGTM

explain-joins.md
```diff
@@ -0,0 +1,299 @@
...
+5 rows in set (0.00 sec)
+```
+
+The execution process of `Hash Join` is as follows:
+
+1. Cache the data of the `Build` side in memory.
+2. Construct a Hash Table on the `Build` side based on the cached data.
+3. Read the data at the `Probe` side.
+4. Use the data of the `Probe` side to probe the Hash Table.
+5. Return qualified data to the user.
```
TomShawn, line 223: Are the logical subjects of the 5 sentences (execution process) all `TiDB`?

explain-joins.md
```diff
@@ -0,0 +1,299 @@
...
+
+The operator info column in the `explain` table also records other information about `Hash Join`, including whether the query is Inner Join or Outer Join, and what are the conditions of Join. In the above example, the query is an Inner Join, where the Join condition `equal:[eq(test.t1.id, test.t2.id)]` partly corresponds with the query statement `WHERE t1.id = t2.id`. The operator info of the other Join operators in the following examples is similar to this one.
```
TomShawn, line 225:
```suggestion
The `operator info` column in the `EXPLAIN` result table also records other information about `Hash Join`, including whether the query is Inner Join or Outer Join, and what are the conditions of Join. In the above example, the query is an Inner Join, where the Join condition `equal:[eq(test.t1.id, test.t2.id)]` partly corresponds with the query condition `WHERE t1.id = t2.id`. The operator info of the other Join operators in the following examples is similar to this one.
```

explain-joins.md
```diff
@@ -0,0 +1,299 @@
...
+
+### Runtime Statistics
+
+If [`tidb_mem_quota_query`](/system-variables.md#tidb_mem_quota_query) (default: 1GB) is exceeded, TiDB will attempt to use temporary storage provided that `oom-use-tmp-storage=TRUE` (default). This means that the `Build` operator used as part of the hash join might be created on disk. Runtime statistics, such as memory usage are visible in the `execution info` of `EXPLAIN ANALYZE`. The following example shows the output of `EXPLAIN ANALYZE` with a 1GB (default) `tidb_mem_quota_query` quota, and a 500MB quota. At 500MB disk is used for temporary storage:
```
TomShawn, line 229:
```suggestion
If [`tidb_mem_quota_query`](/system-variables.md#tidb_mem_quota_query) (default value: 1GB) is exceeded, TiDB will attempt to use temporary storage on condition that the `oom-use-tmp-storage` value is `true` (default). This means that the `Build` operator used as part of the hash join might be created on disk. Runtime statistics such as memory usage are visible in the `execution info` of the `EXPLAIN ANALYZE` result table. The following example shows the output of `EXPLAIN ANALYZE` with a 1GB (default) `tidb_mem_quota_query` quota, and a 500MB quota. At 500MB, disk is used for temporary storage:
```

explain-joins.md
```diff
@@ -0,0 +1,299 @@
...
++-----------------------------+-----------+---------+-----------+---------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------------------------------------------------+-----------------------+----------------------+
+5 rows in set (0.98 sec)
+```
+
+### Configuration
+
+Hash Join performance is influenced by the following system variables:
+
+* [`tidb_mem_quota_query`](/system-variables.md#tidb_mem_quota_query) (default: 1GB) - if the memory quota for a query is exceeded, TiDB will attempt to spill the `Build` operator of a hash join to disk to save memory.
+* [`tidb_hash_join_concurrency`](/system-variables.md#tidb_hash_join_concurrency) (default: 5) - the number of concurrent hash join tasks.
```
TomShawn, line 268:
```suggestion
* [`tidb_mem_quota_query`](/system-variables.md#tidb_mem_quota_query) (default value: 1GB) - if the memory quota for a query is exceeded, TiDB will attempt to spill the `Build` operator of a hash join to disk to save memory.
* [`tidb_hash_join_concurrency`](/system-variables.md#tidb_hash_join_concurrency) (default value: `5`) - the number of concurrent hash join tasks.
```

explain-joins.md
```diff
@@ -0,0 +1,299 @@
...
+| MergeJoin_7                 | 142020.00 | root      |               | inner join, left key:test.t1.id, right key:test.t2.id |
+| ├─TableReader_12(Build)     | 180000.00 | root      |               | data:TableFullScan_11                                 |
+| │ └─TableFullScan_11        | 180000.00 | cop[tikv] | table:t2      | keep order:true                                       |
+| └─TableReader_10(Probe)     | 142020.00 | root      |               | data:TableFullScan_9                                  |
+|   └─TableFullScan_9         | 142020.00 | cop[tikv] | table:t1      | keep order:true                                       |
++-----------------------------+-----------+-----------+---------------+-------------------------------------------------------+
+5 rows in set (0.00 sec)
+```
+
+The execution of the `Merge Join` operator is as follows:
```
TomShawn, line 295:
```suggestion
The execution process of the `Merge Join` operator is as follows:
```

explain-joins.md
```diff
@@ -0,0 +1,299 @@
...
+### Configuration
+
+Hash Join performance is influenced by the following system variables:
+
+* [`tidb_mem_quota_query`](/system-variables.md#tidb_mem_quota_query) (default: 1GB) - if the memory quota for a query is exceeded, TiDB will attempt to spill the `Build` operator of a hash join to disk to save memory.
+* [`tidb_hash_join_concurrency`](/system-variables.md#tidb_hash_join_concurrency) (default: 5) - the number of concurrent hash join tasks.
+
+## Merge Join
+
+Merge join is a special sort of join that applies when both sides of the join are read in sorted order. It can be described as similar to an _efficient zipper merge_, in that as data is read on both the `Build` and the `Probe` side of the join it can be compared as a streaming operation. Merge joins require far less memory than hash join, but do not execute in parallel.
```
TomShawn, line 272:
```suggestion
Merge join is a special sort of join that applies when both sides of the join are read in sorted order. It can be described as similar to an _efficient zipper merge_: as data is read on both the `Build` and the `Probe` sides of the join, the join operation works like a streaming operation. Merge joins require far less memory than hash join but do not execute in parallel.
```

explain-joins.md
```diff
@@ -0,0 +1,299 @@
...
+| │ └─TableFullScan_11        | 180000.00 | cop[tikv] | table:t2      | keep order:true                                       |
+| └─TableReader_10(Probe)     | 142020.00 | root      |               | data:TableFullScan_9                                  |
+|   └─TableFullScan_9         | 142020.00 | cop[tikv] | table:t1      | keep order:true                                       |
++-----------------------------+-----------+-----------+---------------+-------------------------------------------------------+
+5 rows in set (0.00 sec)
+```
+
+The execution of the `Merge Join` operator is as follows:
+
+1. Read all the data of a Join Group from the `Build` side into the memory
```
TomShawn, line 297:
```suggestion
1. Read all the data of a Join Group from the `Build` side into the memory.
```

explain-joins.md
```diff
@@ -0,0 +1,299 @@
...
+INSERT INTO t2 SELECT NULL, a.id, RANDOM_BYTES(1024), RANDOM_BYTES(1024), RANDOM_BYTES(1024) FROM t1 a JOIN t1 b JOIN t1 c LIMIT 10000;
+INSERT INTO t2 SELECT NULL, a.id, RANDOM_BYTES(1024), RANDOM_BYTES(1024), RANDOM_BYTES(1024) FROM t1 a JOIN t1 b JOIN t1 c LIMIT 10000;
+UPDATE t1 SET int_col = 1 WHERE pad1 = (SELECT pad1 FROM t1 ORDER BY RAND() LIMIT 1);
+SELECT SLEEP(1);
+ANALYZE TABLE t1, t2;
+```
+
+## Index Join
+
+If the estimated rows that need to be joined is small (typically less than 10000 rows), using the Index Join method is preferable. This method of join works similar to the primary method of join used in MySQL. In the following example, the operator `├─TableReader_28(Build)` first reads the table `t1`. For each row that matches, TiDB will probe the table `t2`:
```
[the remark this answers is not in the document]
nullnotnil: MySQL until 8.0.xx unly used this method. It now supports hash join, but the use cases where it will select this are minimal.

I am not sure if we should say "it is recommended", since this implies some action on the user's part, such as a query hint. I would say that typically we should not recommend using hints unless the optimizer selects a very bad plan, or the query is very frequent.

explain-joins.md
```diff
@@ -0,0 +1,299 @@
...
+| HashJoin_27                 | 142020.00 | root      |               | inner join, equal:[eq(test.t1.id, test.t2.id)] |
+| ├─TableReader_29(Build)     | 142020.00 | root      |               | data:TableFullScan_28                          |
+| │ └─TableFullScan_28        | 142020.00 | cop[tikv] | table:t1      | keep order:false                               |
+| └─TableReader_31(Probe)     | 180000.00 | root      |               | data:TableFullScan_30                          |
+|   └─TableFullScan_30        | 180000.00 | cop[tikv] | table:t2      | keep order:false                               |
++-----------------------------+-----------+-----------+---------------+------------------------------------------------+
+5 rows in set (0.00 sec)
+```
+
+The execution process of `Hash Join` is as follows:
```
nullnotnil, line 217:
```suggestion
The execution process of `HashJoin_27` is as follows:
```

explain-joins.md
nullnotnil: Yes, this is all TiDB.

explain-joins.md
```diff
@@ -0,0 +1,299 @@
...
+| HashJoin_27                 | 142020.00 | root      |               | inner join, equal:[eq(test.t1.id, test.t2.id)] |
+| ├─TableReader_29(Build)     | 142020.00 | root      |               | data:TableFullScan_28                          |
+| │ └─TableFullScan_28        | 142020.00 | cop[tikv] | table:t1      | keep order:false                               |
+| └─TableReader_31(Probe)     | 180000.00 | root      |               | data:TableFullScan_30                          |
+|   └─TableFullScan_30        | 180000.00 | cop[tikv] | table:t2      | keep order:false                               |
++-----------------------------+-----------+-----------+---------------+------------------------------------------------+
+5 rows in set (0.00 sec)
+```
+
+The execution process of `HashJoin_27` is as follows:
```
TomShawn, line 217:
```suggestion
For the execution process of `HashJoin_27`, TiDB performs the following operations in order:
```

explain-joins.md
```diff
@@ -0,0 +1,299 @@
...
+| MergeJoin_7                 | 142020.00 | root      |               | inner join, left key:test.t1.id, right key:test.t2.id |
+| ├─TableReader_12(Build)     | 180000.00 | root      |               | data:TableFullScan_11                                 |
+| │ └─TableFullScan_11        | 180000.00 | cop[tikv] | table:t2      | keep order:true                                       |
+| └─TableReader_10(Probe)     | 142020.00 | root      |               | data:TableFullScan_9                                  |
+|   └─TableFullScan_9         | 142020.00 | cop[tikv] | table:t1      | keep order:true                                       |
++-----------------------------+-----------+-----------+---------------+-------------------------------------------------------+
+5 rows in set (0.00 sec)
+```
+
+The execution process of the `Merge Join` operator is as follows:
```
TomShawn, line 295:
```suggestion
For the execution process of the `Merge Join` operator, TiDB performs the following operations:
```

explain-joins.md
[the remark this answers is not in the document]
TomShawn: Thanks for explaining!

explain-joins.md
TomShawn: @nullnotnil PTAL is this description acceptable?

nullnotnil: Yes.

TomShawn (approved):
LGTM

ti-srebot (approved):

TomShawn:
Follow-up: TOC

PR #3948, opened 2020-09-23
explain: add subqueries

nullnotnil:

This is a followup PR to #3858 - but they can be reviewed independently, and merge out of order.

This PR does not add this page to the TOC. I will only do that once all PRs are merged (this is one ~3-4 followup PRs).

The content here is based on/duplicates [query-execution-plan](https://docs.pingcap.com/tidb/stable/query-execution-plan). I will merge all of the replacement pages, and then update query-execution-plan (soon to be called explain-overview) as a last step, and in the same PR which adds this page to the TOC.

**Note to reviewers:** I would love to link `index join` to the explain-joins.md document which explains what that means. But I have not for now, since I want this PR or https://github.com/pingcap/docs/pull/3922 to be able to merge out of order. Thanks!

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases
- [ ] Might cause conflicts


lzmhhh123 (commented):
LGTM

explain-subqueries.md
```diff
@@ -0,0 +1,146 @@
...
++-----------------------------+-----------+-----------+------------------------------+--------------------------------------------------------------------------------------------------------+
+6 rows in set (0.00 sec)
+```
+
+Here we can see that TiDB is using a `semi join`. Semi-join differs from an `INNER JOIN`, in that it will only permit the first value on the right key (`t2.t1_id`). i.e. the duplicates are eliminated as part of the join operator. The join algorithm is also a Merge Join, which is like an efficient zipper-merge as it reads data from both the left and the right side in sorted order.
+
+The query is said to be a _correlated subquery_ because the column `t1.int_col` is pushed directly into the subquery, but actually from the `EXPLAIN` output we can see that it is not correlated at all. This is because TiDB has rewritten `t1_id != t1.int_col` to `t1.id != t1.int_col` and moved it out of the subquery. This is a lot more efficient to process, because TiDB can perform this check in `└─Selection_21` as it is reading from the table `t1`.
```
TomShawn, line 122: + `is said to be` seems a bit weird.
+ We can reduce the use of `result + because`, especially in the same passage.
```suggestion
The column `t1.int_col` is pushed directly into the subquery, so this semi-join operation might seem a kind of correlated subquery. However, from the `EXPLAIN` output above, you can see that this column is not correlated at all, because TiDB has rewritten `t1_id != t1.int_col` to `t1.id != t1.int_col` and moved it out of the subquery. Then, TiDB can perform this check in `└─Selection_21` as it is reading data from the table `t1`, which makes the execution a lot more efficient.
```

explain-subqueries.md
```diff
@@ -0,0 +1,146 @@
...
+| │ └─TableFullScan_27        | 1998.00 | cop[tikv] | table:t3      | keep order:false                                                                    |
+| └─TableReader_15(Probe)     | 1.00    | root      |               | data:Selection_14                                                                   |
+|   └─Selection_14            | 1.00    | cop[tikv] |               | lt(test.t1.int_col, 100)                                                            |
+|     └─TableRangeScan_13     | 1.00    | cop[tikv] | table:t1      | range: decided by [test.t3.t1_id], keep order:true                                  |
++-----------------------------+---------+-----------+---------------+-------------------------------------------------------------------------------------+
+6 rows in set (0.00 sec)
+```
+
+This query starts by reading the table `t3`, and then probes the table `t1` based on the `PRIMARY KEY`. The join type is an _anti semi join_; anti because the test is for the non-existence of the value (`NOT IN`) and semi-join because only the first row needs to match before the join is rejected.
```
TomShawn, line 146: Is there any difference between `PRIMARY KEY` and primary key?
```suggestion
This query starts by reading the table `t3` and then probes the table `t1` based on the `PRIMARY KEY`. The join type is an _anti semi join_; anti because this example is for the non-existence of the value (`NOT IN`) and semi-join because only the first row needs to match before the join is rejected.
```

explain-subqueries.md
```diff
@@ -0,0 +1,146 @@
...
+| │     └─IndexFullScan_37       | 90000.00 | cop[tikv] | table:t2, index:t1_id(t1_id) | keep order:false                                                                |
+| └─TableReader_14(Probe)        | 1.00     | root      |                              | data:TableRangeScan_13                                                          |
+|   └─TableRangeScan_13          | 1.00     | cop[tikv] | table:t1                     | range: decided by [test.t2.t1_id], keep order:true                              |
++--------------------------------+----------+-----------+------------------------------+---------------------------------------------------------------------------------+
+7 rows in set (0.00 sec)
+```
+
+TiDB uses an Index Join (merge variant) and first reads the index on `t2.t1_id`. The values of `t1_id` are de-duplicated inside TiKV first as part of `└─HashAgg_31`, and then deduplicated again in TiDB as part of `├─HashAgg_38(Build)`. The deduplication is performed by the aggregation function `firstrow(test.t2.t1_id)`. The result is then joined against the table `t1` against its `PRIMARY KEY`.
```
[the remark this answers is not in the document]
nullnotnil: No, same thing.

explain-subqueries.md
nullnotnil: I will edit it to be clearer. It starts out correlated, but TiDB applies decorrelation as a logical optimization.

nullnotnil: I have updated the wording and included a link to the optimization (subquery decorrelation). PTAL again, thx :-)

explain-subqueries.md
```diff
@@ -0,0 +1,146 @@
...
+|   └─Selection_21            | 56808.00  | cop[tikv] |                              | ne(test.t1.id, test.t1.int_col)                                                                        |
+|     └─TableFullScan_20      | 71010.00  | cop[tikv] | table:t1                     | keep order:true                                                                                        |
++-----------------------------+-----------+-----------+------------------------------+--------------------------------------------------------------------------------------------------------+
+6 rows in set (0.00 sec)
+```
+
+From the result above, you can see that TiDB uses a `Semi Join` algorithm. Semi-join differs from `INNER JOIN`: semi-join only permits the first value on the right key (`t2.t1_id`), which means that the duplicates are eliminated as a part of the join operator task. The join algorithm is also Merge Join, which is like an efficient zipper-merge as the operator reads data from both the left and the right side in sorted order.
```
TomShawn, line 120:
```suggestion
From the result above, you can see that TiDB uses a `Semi Join` algorithm. Semi-join differs from inner join: semi-join only permits the first value on the right key (`t2.t1_id`), which means that the duplicates are eliminated as a part of the join operator task. The join algorithm is also Merge Join, which is like an efficient zipper-merge as the operator reads data from both the left and the right side in sorted order.
```

explain-subqueries.md
```diff
@@ -0,0 +1,146 @@
...
+
+The original statement is considered a _correlated subquery_, since the subquery refers to a column (`t1.int_col`) that exists outside of the subquery. However, the output of `EXPLAIN` shows the execution plan after the [subquery decorrelation optimization](/correlated-subquery-optimization.md) has been applied. The condition `t1_id != t1.int_col` is rewritten to `t1.id != t1.int_col`. This is a lot more efficient, because TiDB can perform this check in `└─Selection_21` as it is reading data from the table `t1`.
```
TomShawn, line 122: + Does `the original statement` mean `SELECT t1_id FROM t2 WHERE t1_id != t1.int_col` or `SELECT * FROM t1 WHERE id IN (SELECT t1_id FROM t2 WHERE t1_id != t1.int_col);`?
+ `since` could be ambiguous in tech documentation.
+ It looks like `t1.int_col` is right inside `SELECT t1_id FROM t2 WHERE t1_id != t1.int_col);` ...
```suggestion
The original statement is considered a _correlated subquery_, because the subquery refers to a column (`t1.int_col`) that exists outside of the subquery. However, the output of `EXPLAIN` shows the execution plan after the [subquery decorrelation optimization](/correlated-subquery-optimization.md) has been applied. The condition `t1_id != t1.int_col` is rewritten to `t1.id != t1.int_col`. This rewriting makes the execution a lot more efficient, because TiDB can perform this check in `└─Selection_21` as it is reading data from the table `t1`.
```

explain-subqueries.md
```diff
@@ -0,0 +1,146 @@
...
+6 rows in set (0.00 sec)
+```
+
+From the result above, you can see that TiDB uses a `Semi Join` algorithm. Semi-join differs from inner join: semi-join only permits the first value on the right key (`t2.t1_id`), which means that the duplicates are eliminated as a part of the join operator task. The join algorithm is also Merge Join, which is like an efficient zipper-merge as the operator reads data from both the left and the right side in sorted order.
+
+The original statement is considered a _correlated subquery_, because the subquery refers to a column (`t1.int_col`) that exists outside of the subquery. However, the output of `EXPLAIN` shows the execution plan after the [subquery decorrelation optimization](/correlated-subquery-optimization.md) has been applied. The condition `t1_id != t1.int_col` is rewritten to `t1.id != t1.int_col`. This decorrelation and rewriting make the execution a lot more efficient, because TiDB can perform this check in `└─Selection_21` as it is reading data from the table `t1`.
```
TomShawn, line 122:
```suggestion
The original statement is considered a _correlated subquery_, because the subquery refers to a column (`t1.int_col`) that exists outside of the subquery. However, the output of `EXPLAIN` shows the execution plan after the [subquery decorrelation optimization](/correlated-subquery-optimization.md) has been applied. The condition `t1_id != t1.int_col` is rewritten to `t1.id != t1.int_col`. TiDB can perform this in `└─Selection_21` as it is reading data from the table `t1`, so this decorrelation and rewriting make the execution a lot more efficient.
```

TomShawn (approved):
LGTM

TomShawn:
Follow-up: TOC

PR #4741, opened 2021-01-27
faq, sysvars: add tidb_multi_statement_mode

morgo:

This is the documentation for https://github.com/pingcap/tidb/pull/22351

It has version specific changes, because the default is `WARN` in 4.0.11 and `OFF` in master/5.0.

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases
- [x] Need modification after applied to another branch 


morgo:
PTAL @AilinKid @xhebox

xhebox (commented):
Based on the implementation, I think you can mention that ON and WARN actually will only take effects if clients set multistmt off.

morgo:
> Based on the implementation, I think you can mention that ON and WARN actually will only take effects if clients set multistmt off.

It describes this case in this line:

> `OFF` multi-statement is prevented, unless the client sets the multi-statement attribute.

All client libraries disable multi-statement by default. It also says in the note that it is preferrable to use the client setting rather than the server setting:

> If your application requires multiple statement support, it is recommended to use the setting provided by your client library instead of the `tidb_multi_statement_mode` option.

xhebox:
> It describes this case in this line:

But I did not see that in ON and WARN. I always feel like ON will allow clients multi-query even if their setting are off. Warn will report warnings whatever the settings is.

AilinKid:
> Based on the implementation, I think you can mention that ON and WARN actually will only take effects if clients set multistmt off.

how about making a list to describe this?

|  client   | tidb  | result |
|  ----  | ----  | ---- |
| A | OFF | follow A |
| xxx  | xxx | xxx |
| ... | |

morgo:
> > Based on the implementation, I think you can mention that ON and WARN actually will only take effects if clients set multistmt off.
> 
> how about making a list to describe this?
> 
> client	tidb	result
> OFF	whatever	follow tidb
> ON	whatever	ON
> ...

Sure, that will make it clearer.

morgo:
The table looks like this:
| Client setting         | `tidb_multi_statement_mode` value | Multiple statements permitted? |
|------------------------|-----------------------------------|--------------------------------|
| Multi Statements = ON  | OFF                               | Yes                            |
| Multi Statements = ON  | ON                                | Yes                            |
| Multi Statements = ON  | WARN                              | Yes                            |
| Multi Statements = OFF | OFF                               | No                             |
| Multi Statements = OFF | ON                                | Yes                            |
| Multi Statements = OFF | WARN                              | Yes (+warning returned)        |

The client setting will differ per client library, so I generically described it as "Multi Statements = ON/OFF".

xhebox (commented):
LGTM

AilinKid (commented):
LGTM

system-variables.md
```diff
@@ -710,6 +710,31 @@ For a system upgraded to v5.0.0-rc from an earlier version, if you have not modi
 - Default value: 60
 - This variable is used to set the step of the Prometheus statement generated when querying `METRIC_SCHEMA`. The unit is second.
 
+## tidb_multi_statement_mode <span class="version-mark">New in v4.0.11</span>
+
+- Scope: SESSION | GLOBAL
+- Default value: OFF
+- Permitted values: OFF, ON, WARN
+- To reduce the impact from SQL injection attacks, TiDB now prevents multiple queries being executed in the same `COM_QUERY` call by default.
+- This variable is intended to be used as part of an upgrade path from earlier versions of TiDB. The following behaviors apply:
```
TomShawn, line 719: Could you specify which versions "TiDB now" and "earlier versions of TiDB" might refer to?
```suggestion
- To reduce the impact of SQL injection attacks, TiDB now prevents multiple queries from being executed in the same `COM_QUERY` call by default. This variable is intended to be used as part of an upgrade path from earlier versions of TiDB. The following behaviors apply:
```

system-variables.md
```diff
@@ -710,6 +710,31 @@ For a system upgraded to v5.0.0-rc from an earlier version, if you have not modi
...
+
+| Client setting         | `tidb_multi_statement_mode` value | Multiple statements permitted? |
+|------------------------|-----------------------------------|--------------------------------|
+| Multi Statements = ON  | OFF                               | Yes                            |
```
TomShawn, line 723: Are `Multi Statements` and `Multiple statements` the same thing? If so, shall we unify the spelling?

system-variables.md
```diff
@@ -710,6 +710,31 @@ For a system upgraded to v5.0.0-rc from an earlier version, if you have not modi
...
+| Multi Statements = ON  | ON                                | Yes                            |
+| Multi Statements = ON  | WARN                              | Yes                            |
+| Multi Statements = OFF | OFF                               | No                             |
+| Multi Statements = OFF | ON                                | Yes                            |
+| Multi Statements = OFF | WARN                              | Yes (+warning returned)        |
+
+> **Note:**
+>
+> Only the default value of `OFF` can be considered safe. Setting `tidb_multi_statement_mode=ON` might be required if your application was specifically designed for an earlier version of TiDB. If your application requires multiple statement support, it is recommended to use the setting provided by your client library instead of the `tidb_multi_statement_mode` option. For example:
```
TomShawn, line 732: Why "If your application **was** specifically designed for an earlier version" and "If your application **requires**" are in different tenses?
```suggestion
> Only the default value of `OFF` can be considered safe. If your application was specifically designed for an earlier version of TiDB, you might need to set `tidb_multi_statement_mode=ON`. If your application requires the "multiple statements" support, it is recommended to use the setting provided by your client library instead of the `tidb_multi_statement_mode` option. Some of your client libraries might be as follows:
```

system-variables.md
morgo: Kind of.

We can unify the spelling, but each client library has a different name for this. "Multi statements"  is used here to refer to the client setting, which is more a concept than a specific feature.

system-variables.md
morgo: This reads kind of awkward because it starts two sentences with "If your application". Because "multiple statements" in this context is a concept, I don't think we should put it in quotes.

Can we leave as is?

morgo:
@TomShawn PTAL

TomShawn:
LGTM
This PR will be merged after the v4.0.11 release.

TomShawn (approved):

PR #5028, opened 2021-03-12
update use tiup client access database

King-Dylan:
update use tiup client access database

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/quick-start-with-tidb.md
+++ b/quick-start-with-tidb.md
@@ -60,7 +60,7 @@ As a distributed system, a basic TiDB test cluster usually consists of 2 TiDB in
         tiup playground <version> --db 2 --pd 3 --kv 3 --monitor
         ```
 
-        The command downloads a <version> cluster to the local machine and starts it, such as v4.0.11. `--monitor` means that the monitoring component is also deployed.
+        The command downloads a <version> cluster to the local machine and starts it. `--monitor` means that the monitoring component is also deployed.
 
         To view the latest version, run `tiup list tidb`.
 
--- a/quick-start-with-tidb.md
+++ b/quick-start-with-tidb.md
@@ -57,10 +57,10 @@ As a distributed system, a basic TiDB test cluster usually consists of 2 TiDB in
         {{< copyable "shell-regular" >}}
 
         ```shell
-        tiup playground {version} --db 2 --pd 3 --kv 3 --monitor
+        tiup playground v4.0.11 --db 2 --pd 3 --kv 3 --monitor
         ```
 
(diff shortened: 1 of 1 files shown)


TomShawn:
@King-Dylan Please involve a technical review.

quick-start-with-tidb.md
```diff
@@ -76,15 +76,15 @@ As a distributed system, a basic TiDB test cluster usually consists of 2 TiDB in
...
 4. Start a new session to access TiDB:
 
-    1. Install the MySQL client. If it is already installed, skip this step.
+    1. Use the TiUP client to connect to TiDBs.
 
         {{< copyable "shell-regular" >}}
 
         ```shell
-        yum -y install mysql
+        tiup client
```
morgo, line 84: This client doesn't support reading SQL files on STDIN, so it will be hard to use it in examples where we suggest loading data. We should skip switching it for now, and improve the client first,

quick-start-with-tidb.md
```diff
@@ -57,10 +57,10 @@ As a distributed system, a basic TiDB test cluster usually consists of 2 TiDB in
         {{< copyable "shell-regular" >}}
 
         ```shell
-        tiup playground v4.0.0 --db 2 --pd 3 --kv 3 --monitor
+        tiup playground {version} --db 2 --pd 3 --kv 3 --monitor
         ```
 
-        The command downloads a v4.0.0 cluster to the local machine and starts it. `--monitor` means that the monitoring component is also deployed.
+        The command downloads a {version} cluster to the local machine and starts it，such as v4.0.11. `--monitor` means that the monitoring component is also deployed.
```
morgo, line 63: I think it is better if it shows v4.0.11 in the example and then says you can change it. This makes the command copy+pasteable.

It is ideally updated as new versions are released, but being stuck on 4.0.11 for a little while is at least better than 4.0.0.

quick-start-with-tidb.md
```diff
@@ -57,10 +57,10 @@ As a distributed system, a basic TiDB test cluster usually consists of 2 TiDB in
         {{< copyable "shell-regular" >}}
 
         ```shell
-        tiup playground v4.0.0 --db 2 --pd 3 --kv 3 --monitor
+        tiup playground v4.0.11 --db 2 --pd 3 --kv 3 --monitor
```
TomShawn, line 60: @morgo @King-Dylan Does this need version-specific change in master and release-5.0?

morgo: That would be ideal. master and v5.0 do not technically have releases yet, so I think we can keep it as is for this merge.

PR #5203, opened 2021-04-05
tiflash: add limitation for TiFlash usage

TomShawn:

Add limitation for TiFlash usage.

- This PR is translated from: https://github.com/pingcap/docs-cn/pull/5891
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases


tiflash/use-tiflash.md
```diff
@@ -277,11 +273,46 @@ explain select count(*) from customer c join nation n on c.c_nationkey=n.n_natio
...
+
+## Notes
+
+Currently, TiFlash does not support some features. These features might be incompatible with the native TiDB:
+
+* In the TiFlash computation layer:
+    * Checking overflowed numerical values is not supported. For example, adding two maximum values of the `BIGINT` type `9223372036854775807 + 9223372036854775807`. The expected behavior of this calculation in TiDB is to return the `ERROR 1690 (22003): BIGINT value is out of range` error. However, if this calculation is performed in TiFlash, an overflow value of `-2` is returned without any error.
+    * The window function is not supported.
+    * Reading data from TiKV is not supported.
+    * Currently, the `sum` function in TiFlash does not support the string-type argument. But TiDB cannot find out whether any string-type argument has been passed into the `sum` function at compile time. Therefore, when you execute statements similar to `select sum(string_col) from t`, TiFlash returns the `[FLASH:Coprocessor:Unimplemented] CastStringAsReal is not supported.` error. To avoid such an error in this case, you need to modify this SQL statement as `select sum(cast(string_col as double)) from t`.
```
qiancai, line 289:
```suggestion
    * Currently, the `sum` function in TiFlash does not support the string-type argument. But TiDB cannot identify whether any string-type argument has been passed into the `sum` function during the compiling. Therefore, when you execute statements similar to `select sum(string_col) from t`, TiFlash returns the `[FLASH:Coprocessor:Unimplemented] CastStringAsReal is not supported.` error. To avoid such an error in this case, you need to modify this SQL statement to `select sum(cast(string_col as double)) from t`.
```

tiflash/use-tiflash.md
```diff
@@ -277,11 +273,46 @@ explain select count(*) from customer c join nation n on c.c_nationkey=n.n_natio
...
+    * Currently, TiFlash's decimal division calculation is incompatible with that of TiDB. For example, when dividing decimal, TiFlash performs the calculation always using the type inferred at compile time. However, TiDB performs this calculation using a type that is more precise than that inferred at compile time. Therefore, some SQL statements involving decimal division have different execution results when executed in TiDB + TiKV and in TiDB + TiFlash. For example:
```
qiancai, line 290:
```suggestion
    * Currently, TiFlash's decimal division calculation is incompatible with that of TiDB. For example, when dividing decimal, TiFlash performs the calculation always using the type inferred from the compiling. However, TiDB performs this calculation using a type that is more precise than that inferred from the compiling. Therefore, some SQL statements involving the decimal division return different execution results when executed in TiDB + TiKV and in TiDB + TiFlash. For example:
```

tiflash/use-tiflash.md
```diff
@@ -277,11 +273,46 @@ explain select count(*) from customer c join nation n on c.c_nationkey=n.n_natio
...
+        mysql> set session tidb_isolation_read_engines='tiflash';
+        Query OK, 0 rows affected (0.00 sec)
+        mysql> select a/b, a/b + 0.0000000000001 from t where a/b;
+        Empty set (0.01 sec)
+        ```
+
+        In the example above, `a/b`'s inferred type at compile time is `Decimal(7,4)` both in TiDB and in TiFlash. Constrained by `Decimal(7,4)`, `a/b`'s returned type should be `0.0000`. In TiDB, `a/b`'s runtime precision is higher than `Decimal(7,4)`, so the original table data is not filtered by the `where a/b` condition. However, in TiFlash, the calculation of `a/b` uses `Decimal(7,4)` as the result type, so the original table data is filtered by the `where a/b` condition.
+
+* The MPP mode in TiFlash does not support the following features:
+    * The partitioned table is not supported. For queries with the partitioned table, the MPP mode is not chosen by default.
```
qiancai, line 317:
```suggestion
    * The partitioned table is not supported. For queries on the partitioned tables, the MPP mode is not chosen by default.
```

zanmato1984 (approved):
LGTM

tiflash/use-tiflash.md
```diff
@@ -277,11 +273,46 @@ explain select count(*) from customer c join nation n on c.c_nationkey=n.n_natio
...
+        | 0.0000 |       0.0000000410001 |
+        +--------+-----------------------+
+        1 row in set (0.00 sec)
+        mysql> set session tidb_isolation_read_engines='tiflash';
+        Query OK, 0 rows affected (0.00 sec)
+        mysql> select a/b, a/b + 0.0000000000001 from t where a/b;
+        Empty set (0.01 sec)
+        ```
+
+        In the example above, `a/b`'s inferred type at compile time is `Decimal(7,4)` both in TiDB and in TiFlash. Constrained by `Decimal(7,4)`, `a/b`'s returned type should be `0.0000`. In TiDB, `a/b`'s runtime precision is higher than `Decimal(7,4)`, so the original table data is not filtered by the `where a/b` condition. However, in TiFlash, the calculation of `a/b` uses `Decimal(7,4)` as the result type, so the original table data is filtered by the `where a/b` condition.
```
TomShawn, line 314:
```suggestion
        In the example above, `a/b`'s inferred type from the compiling is `Decimal(7,4)` both in TiDB and in TiFlash. Constrained by `Decimal(7,4)`, `a/b`'s returned type should be `0.0000`. In TiDB, `a/b`'s runtime precision is higher than `Decimal(7,4)`, so the original table data is not filtered by the `where a/b` condition. However, in TiFlash, the calculation of `a/b` uses `Decimal(7,4)` as the result type, so the original table data is filtered by the `where a/b` condition.
```

tiflash/use-tiflash.md
```diff
@@ -277,11 +273,46 @@ explain select count(*) from customer c join nation n on c.c_nationkey=n.n_natio
...
-> By default, TiFlash does not support using the MPP mode in partitioned tables.
-
 TiFlash provides the following two global/session variables to control whether to use Broadcast Hash Join:
 
 - [`tidb_broadcast_join_threshold_size`](/system-variables.md#tidb_broadcast_join_threshold_count-new-in-v50): The unit of the value is bytes. If the table size (in the unit of bytes) is less than the value of the variable, the Broadcast Hash Join algorithm is used. Otherwise, the Shuffled Hash Join algorithm is used.
 - [`tidb_broadcast_join_threshold_count`](/system-variables.md#tidb_broadcast_join_threshold_count-new-in-v50): The unit of the value is rows. If the objects of the join operation belong to a subquery, the optimizer cannot estimate the size of the subquery result set, so the size is determined by the number of rows in the result set. If the estimated number of rows in the subquery is less than the value of this variable, the Broadcast Hash Join algorithm is used. Otherwise, the Shuffled Hash Join algorithm is used.
+
+## Notes
+
+Currently, TiFlash does not support some features. These features might be incompatible with the native TiDB:
```
TomShawn, line 283:
```suggestion
Currently, TiFlash does not support some features. These features might be incompatible with the native TiDB: 
```

tiflash/use-tiflash.md
```diff
@@ -277,11 +273,46 @@ explain select count(*) from customer c join nation n on c.c_nationkey=n.n_natio
...
```
TomShawn, line 283:
```suggestion
Currently, TiFlash does not support some features. These features might be incompatible with the native TiDB:
```

PR #5217, opened 2021-04-06
update configuration page for enable-enum-length-limit

TomShawn:
Added the description for a new configuration entry for the 'tidb-configuration-file'.

- This PR is translated from: https://github.com/pingcap/docs-cn/pull/5937
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/tidb-configuration-file.md
+++ b/tidb-configuration-file.md
@@ -165,7 +165,7 @@ The TiDB configuration file supports more options than command-line parameters. 
 
 + Determines whether to limit the maximum length of a single `ENUM` element and a single `SET` element.
 + Default value: `true`
-+ When this configuration value is `true`, the maximum length of a single `ENUM` element and a single `SET` element is 255 characters, which is compatible with [MySQL 8.0](https://dev.mysql.com/doc/refman/8.0/en/string-type-syntax.html); when this configuration value is `false`, there is no limit on the length of a single element, which is compatible with TiDB (earlier than v5.0).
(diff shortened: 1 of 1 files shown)


tidb-configuration-file.md
```diff
@@ -155,12 +155,18 @@ The TiDB configuration file supports more options than command-line parameters.
...
 
 - Enables or disables listening on TCP4 only.
 - Default value: `false`
 - Enabling this option is useful when TiDB is used with LVS for load balancing because the [real client IP from the TCP header](https://github.com/alibaba/LVS/tree/master/kernel/net/toa) can be correctly parsed by the "tcp4" protocol.
 
+### `enable-enum-length-limit` <span class="version-mark">New in v5.0</span>
+
++ Determines whether to limit the maximum length of a single `ENUM` element and a single `SET` element.
++ Default value: `true`
++ When this configuration value is `true`, the maximum length of a single `ENUM` element and a single `SET` element is 255 characters, which is compatible with [MySQL 8.0](https://dev.mysql.com/doc/refman/8.0/en/string-type-syntax.html); when this configuration value is `false`, there is no limit on the length of a single element, which is compatible with TiDB (earlier than v5.0).
```
qiancai, line 168:
```suggestion
+ When this configuration value is `true`, the maximum length of a single `ENUM` element and a single `SET` element is 255 characters, which is compatible with [MySQL 8.0](https://dev.mysql.com/doc/refman/8.0/en/string-type-syntax.html). When this configuration value is `false`, there is no limit on the length of a single element, which is compatible with TiDB (earlier than v5.0).
```

PR #5283, opened 2021-04-09
check-before-deployment: Add info about chronyd/chronyc on CentOS 8

dveeden:
On CentOS 8 Chrony is used by default for NTP services.
- https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/using-chrony-to-configure-ntp#sect-checking_if_chrony_is_synchronized

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/check-before-deployment.md
+++ b/check-before-deployment.md
@@ -249,7 +249,7 @@ To check whether the NTP service is installed and whether it synchronizes with t
     chronyc tracking
     ```
 
-    - If it returns `Leap status     : Normal`, then the synchronization process is normal.
+    - If the command returns `Leap status     : Normal`, the synchronization process is normal.
 
         ```
         Reference ID    : 5EC69F0A (ntp1.time.nl)
--- a/check-before-deployment.md
+++ b/check-before-deployment.md
@@ -267,13 +267,13 @@ To check whether the NTP service is installed and whether it synchronizes with t
         Leap status     : Normal
         ```
 
-    - The following situation indicates the synchronization is not synchronizing normally:
+    - If the command returns the following result, an error occurs in the synchronization:
 
         ```
         Leap status    : Not synchronised
         ```
 
-    - The following situation indicates the `chronyd` service is not running normally:
+    - If the command returns the following result, the `chronyd` service is not running normally:
 
         ```
         506 Cannot talk to daemon
--- a/check-before-deployment.md
+++ b/check-before-deployment.md
@@ -241,7 +241,7 @@ To check whether the NTP service is installed and whether it synchronizes with t
 
     > **Note:**
     >
-    > This only applies to system that use Chrony instead of NTPd.
+    > This only applies to systems that use Chrony instead of NTPd.
 
     {{< copyable "shell-regular" >}}
 
--- a/check-before-deployment.md
+++ b/check-before-deployment.md
@@ -203,7 +203,7 @@ To check whether the NTP service is installed and whether it synchronizes with t
     Active: active (running) since Mon 2021-04-05 09:55:29 EDT; 3 days ago
     ```
 
-    This would mean that your system is configured to use `chronyd` instead of `ntpd` to do synchronization with NTP. If this is the case proceed to step 3.
+    If your system is configured to use `chronyd`, proceed to step 3.
 
 2. Run the `ntpstat` command to check whether the NTP service synchronizes with the NTP server.
 
--- a/check-before-deployment.md
+++ b/check-before-deployment.md
@@ -189,7 +189,7 @@ To check whether the NTP service is installed and whether it synchronizes with t
     Active: active (running) since 一 2017-12-18 13:13:19 CST; 3s ago
     ```
 
-    - If it returns `Unit ntpd.service could not be found.` then try the following:
+    - If it returns `Unit ntpd.service could not be found.`, then try the following command to see whether your system is configured to use `chronyd` instead of `ntpd` to perform clock synchronization with NTP:
 
     {{< copyable "shell-regular" >}}
 
--- a/check-before-deployment.md
+++ b/check-before-deployment.md
@@ -191,19 +191,19 @@ To check whether the NTP service is installed and whether it synchronizes with t
 
     - If it returns `Unit ntpd.service could not be found.`, then try the following command to see whether your system is configured to use `chronyd` instead of `ntpd` to perform clock synchronization with NTP:
 
-    {{< copyable "shell-regular" >}}
-
-    ```bash
-    sudo systemctl status cronyd.service
-    ```
-
-    ```
-    chronyd.service - NTP client/server
-    Loaded: loaded (/usr/lib/systemd/system/chronyd.service; enabled; vendor preset: enabled)
-    Active: active (running) since Mon 2021-04-05 09:55:29 EDT; 3 days ago
-    ```
-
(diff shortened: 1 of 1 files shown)


lucklove:
We don't support CentOS 8 yet. It's not fully tested.

dveeden:
> We don't support CentOS 8 yet. It's not fully tested.
https://github.com/pingcap/docs/blob/master/hardware-and-software-requirements.md#linux-os-version-requirements

This list has "CentOS: 7.3 or later". If CentOS 8 is not supported this should probably be changed to exclude CentOS 8.

It looks like Ubuntu is also using chrony now: https://ubuntu.com/blog/ubuntu-bionic-using-chrony-to-configure-ntp

TomShawn:
> This list has "CentOS: 7.3 or later". If CentOS 8 is not supported this should probably be changed to exclude CentOS 8.

Could you please do that in this PR? Thanks!

dveeden:
> > This list has "CentOS: 7.3 or later". If CentOS 8 is not supported this should probably be changed to exclude CentOS 8.
> 
> Could you please do that in this PR? Thanks!

I've created #5587 for this

TomShawn:
> > > This list has "CentOS: 7.3 or later". If CentOS 8 is not supported this should probably be changed to exclude CentOS 8.
> > 
> > 
> > Could you please do that in this PR? Thanks!
> 
> I've created #5587 for this

Shall we also update "On CentOS 8 you need to use `chronyc tracking` instead of `ntpstat` if you are using `chronyd` instead of `ntpd`." in this PR?

dveeden:
@TomShawn PTAL

TomShawn:
@lucklove PTAL, thanks!

lucklove:
/lgtm

But I shoud say that we didn't test TiDB at CentOS 8 at all, and there is no plan to support that at this time.

check-before-deployment.md
```diff
@@ -189,6 +189,22 @@ To check whether the NTP service is installed and whether it synchronizes with t
     Active: active (running) since 一 2017-12-18 13:13:19 CST; 3s ago
     ```
 
+    - If it returns `Unit ntpd.service could not be found.` then try the following:
```
TomShawn, line 192: This tells users what this command is about before they try it.
```suggestion
    If it returns `Unit ntpd.service could not be found.`, then try the following command, which configures your system to use `chronyd` instead of `ntpd` to perform clock synchronization with NTP:
```

check-before-deployment.md
```diff
@@ -189,6 +189,22 @@ To check whether the NTP service is installed and whether it synchronizes with t
...
+    sudo systemctl status cronyd.service
+    ```
+
+    ```
+    chronyd.service - NTP client/server
+    Loaded: loaded (/usr/lib/systemd/system/chronyd.service; enabled; vendor preset: enabled)
+    Active: active (running) since Mon 2021-04-05 09:55:29 EDT; 3 days ago
+    ```
+
+    This would mean that your system is configured to use `chronyd` instead of `ntpd` to do synchronization with NTP. If this is the case proceed to step 3.
```
TomShawn, line 206:
```suggestion
    If your system is configured to use `chronyd`, proceed to step 3.
```

check-before-deployment.md
```diff
@@ -221,6 +237,48 @@ To check whether the NTP service is installed and whether it synchronizes with t
         Unable to talk to NTP daemon. Is it running?
         ```
 
+3. Run the `chronyc tracking` command to check wheter the Chrony service synchronizes with the NTP server.
+
+    > **Note:**
+    >
+    > This only applies to system that use Chrony instead of NTPd.
```
TomShawn, line 244:
```suggestion
    > This only applies to system that uses Chrony instead of NTPd.
```

check-before-deployment.md
```diff
@@ -221,6 +237,48 @@ To check whether the NTP service is installed and whether it synchronizes with t
...
+
+    {{< copyable "shell-regular" >}}
+
+    ```bash
+    chronyc tracking
+    ```
+
+    - If it returns `Leap status     : Normal`, then the synchronization process is normal.
```
TomShawn, line 252:
```suggestion
    - If the command returns `Leap status     : Normal`, the synchronization process is normal.
```

check-before-deployment.md
```diff
@@ -221,6 +237,48 @@ To check whether the NTP service is installed and whether it synchronizes with t
...
+        Frequency       : 2.286 ppm slow
+        Residual freq   : -0.000 ppm
+        Skew            : 0.012 ppm
+        Root delay      : 0.012706812 seconds
+        Root dispersion : 0.000430042 seconds
+        Update interval : 1029.8 seconds
+        Leap status     : Normal
+        ```
+
+    - The following situation indicates the synchronization is not synchronizing normally:
```
TomShawn, line 270:
```suggestion
    - If the command returns the following result, an error occurs in the synchronization:
```

check-before-deployment.md
```diff
@@ -221,6 +237,48 @@ To check whether the NTP service is installed and whether it synchronizes with t
...
+
+        ```
+        Leap status    : Not synchronised
+        ```
+
+    - The following situation indicates the `chronyd` service is not running normally:
```
TomShawn, line 276:
```suggestion
    - If the command returns the following result, the `chronyd` service is not running normally:
```

check-before-deployment.md
dveeden: I think the original is sentence is correct.

TomShawn: Grammatically, the logical subject of the verb "use" is "system", which is a singular noun. Therefore, in this case, "uses" is grammatically correct. If we want to use "use", "system" should be in the plural form "systems".

dveeden: Hmz. in that case both look wrong to me.

Would any of these be ok?

```suggestion
    > This only applies to systems that use Chrony instead of NTPd.
```

```suggestion
    > This only applies to a system that uses Chrony instead of NTPd.
```

TomShawn: Yes. Both LGTM

check-before-deployment.md
```diff
@@ -189,6 +189,22 @@ To check whether the NTP service is installed and whether it synchronizes with t
...
+    sudo systemctl status cronyd.service
+    ```
+
+    ```
+    chronyd.service - NTP client/server
+    Loaded: loaded (/usr/lib/systemd/system/chronyd.service; enabled; vendor preset: enabled)
+    Active: active (running) since Mon 2021-04-05 09:55:29 EDT; 3 days ago
+    ```
+
+    If your system is configured to use `chronyd`, proceed to step 3.
```
TomShawn, line 206:
```suggestion
        {{< copyable "shell-regular" >}}
    
        ```bash
        sudo systemctl status cronyd.service
        ```
    
        ```
        chronyd.service - NTP client/server
        Loaded: loaded (/usr/lib/systemd/system/chronyd.service; enabled; vendor preset: enabled)
        Active: active (running) since Mon 2021-04-05 09:55:29 EDT; 3 days ago
        ```
    
        If your system is configured to use `chronyd`, proceed to step 3.
```

TomShawn (approved):
LGTM

PR #5346, opened 2021-04-15
*: update some sentences.

bb7133:

A tiny update to modify some sentences.

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases


xiongjiwei (approved):

error-codes.md
```diff
@@ -98,7 +98,7 @@ TiDB is compatible with the error codes in MySQL, and in most cases returns the
 
 * Error Number: 8027
 
-    The table schema version is outdated. TiDB uses the F1 online schema change algorithm to execute DDL statements. When the table schema version of the TiDB server is earlier than that of the entire system, this error is returned if you execute a SQL statement.
+    The table schema version is outdated. TiDB uses the online schema change algorithm to execute DDL statements. When the table schema version of the TiDB server is earlier than that of the entire system, this error is returned if you execute a SQL statement.
```
morgo, line 101: I think because F1 is removed, "the" should also be removed because otherwise it implies there is only one algorithm to perform online schema change, which is not true.

Thus the text can be further simplified to something like:
"TiDB applies schema changes online. When the table schema version of the TiDB server is earlier than the.."

TomShawn: Updated in https://github.com/pingcap/docs/pull/5352.

PR #5349, opened 2021-04-15
explain: add hint about disabling line wrapping with less

dveeden:

Add a hint to the `EXPLAIN` documentation about how to use `less -S` to view large explain output without line wrapping.

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/explain-overview.md
+++ b/explain-overview.md
@@ -8,7 +8,7 @@ aliases: ['/docs/dev/query-execution-plan/','/docs/dev/reference/performance/und
 
 > **Note:**
 >
-> If you use the `mysql` client to connect to TiDB you may want to use `pager less -S` to avoid line wrapping in your terminal and instead press the right arrow on your keyboard to horizontally scroll through the explain output.
+> When you use the MySQL client to connect to TiDB, to read the output result in a clearer way without line wrapping, you can use the `pager less -S` command. Then, after the `EXPLAIN` result is output, you can press the right arrow <kbd>→</kbd> button on your keyboard to horizontally scroll through the output.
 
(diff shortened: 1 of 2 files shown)


explain-overview.md
```diff
@@ -6,6 +6,10 @@ aliases: ['/docs/dev/query-execution-plan/','/docs/dev/reference/performance/und
 
 # `EXPLAIN` Overview
 
+> **Note:**
+>
+> If you use the `mysql` client to connect to TiDB you may want to use `pager less -S` to avoid line wrapping in your terminal and instead press the right arrow on your keyboard to horizontally scroll through the explain output.
```
TomShawn, line 11: + Does "the right arrow on your keyboard" mean <kbd>></kbd>?
+ `pager less -S` and "the right arrow on your keyboard" are used together?

dveeden: This is how this works:

```
mysql> pager less -S
mysql> EXPLAIN ANALYZE SELECT.....
```
The output is shown in `less` and instead of wrapping lines it chops off the part that doesn't fit on the screen. Then by pressing the right arrow key on the keyboard you can 'scroll' to the right. Here `\G` in MySQL doesn't help.

Example w/o this:
image

Example with less -S:
image

Here you can see that `less` indicates there is more output on the right with `>`. The output can be viewed by pressing the right arrow key on the keyboard.

This improves the output of the tree-like explain output that is in the `id` column of the explain output.

TomShawn, line 11:
```suggestion
> When you use the MySQL client to connect to TiDB, to read the output result in a clearer way without line wrapping, you can use the `pager less -S` command. Then, after the `EXPLAIN` result is output, you can press the right arrow <kbd>→</kbd> button on your keyboard to horizontally scroll through the output.
```

TomShawn:
/label requires-followup
Follow-up: update Line 33 in sql-statements/sql-statement-explain.md in master.

Joyinqin:
> /label requires-followup
> Follow-up: update Line 33 in sql-statements/sql-statement-explain.md in master.

Done in #5407

PR #5555, opened 2021-04-29
Add loadbalancer documentation

dveeden:

Add documentation about how to use loadbalancers with TiDB

### DIFF
--- a/loadbalancing.md
+++ b/loadbalancing.md
@@ -1,0 +1,371 @@
+---
+title: Loadbalancing
+summary: How to use loadbalancing with TiDB
+---
+
+# Loadbalancing
+
+Loadbalancing is used to distribute connections from applications to TiDB Server instances. This helps to distribute the load over multiple machines and (depending on the loadbalancing option selected) can automatically re-route connections should a TiDB instance become unavailable.
+
+## Loadbalancing types
+
+There are many different ways implementing loadbalancer. In this section the most common types of loadbalancing are described.
+
+| Loadbalancer type | Usage scenarios | Benefits | Drawbacks |
+| ------------------|-----------------|----------|-----------|
+| Connector based, such as MySQL Connector/J and MySQL Connector/C++   | Load balancing a specific application | No need for extra components | Connector specific implementation and behavior |
+| DNS (Round-robin or SRV) | Simple loadbalancing without requiring application modifications  | Re-using existing components and methods | Not automatically removing failed nodes from service |
+| Kubernetes Loadbalancer (e.g. metallb or Amazon ELB) | Load balancing in the cloud | Well integrated with Kubernetes | Requires Kubernetes |
+| Software based loadbalancer | A daemon that implements loadbalancing | Flexible | Often service specific |
+| Hardware based loadbalancer (e.g. F5) | A hardware loadbalancer | Can use hardware accelleration | Expensive |
+
+The first type of loadbalancing that is available is **connector based loadbalancing**. Many of the MySQL connectors like MySQL Connector/J and MySQL Connector/C++. The benefits from this are that there is no extra network hop and the application has more information about which server it is connected to. The drawback of this is that there is no central administration and when changing the configuration you have to do this on all your application hosts. Depending on the programming language you are using the loadbalancing might not be available or not offer more advanced options. This can work with most third-party applications as this is often configured in the connection string (JDBC url for Java for example).
+
+It is also possible to use **DNS round-robin**. However this is not advisable as this won't prevent connections from going to a machine that is unavailable. This means that your application might have to re-try connections more often if not all servers are available. Some of the newer MySQL connectors have support for **[DNS SRV](https://dev.mysql.com/doc/refman/8.0/en/connecting-using-dns-srv.html)**. This is somewhat similar to round-robin DNS, but allows you to set priorities. The benefit from this is that it is an industry standard and that this makes the client-side configuration easier.
+
+If you are using Kubernetes you probably want to use the [LoadBalancer](https://docs.pingcap.com/tidb-in-kubernetes/stable/access-tidb#loadbalancer) service type that works with Amazon ELB and similar services from other cloud vendors. For on-premise Kubernetes something like [metallb](https://metallb.universe.tf/) can be used.
+
+When you are not using Kubernetes to deploy on a cloud service you can still use the loadbalancing service from your cloud provider.
+
+Another common option is to use a **software based** loadbalancer. Common ones are ProxySQL, haproxy and MySQL Router.
+
+The last type of loadbalancing is **hardware based** loadbalancers. These are physical machines where you plug in a network cable. These are often costly, but also offer a high throughput.
+
+## General loadbalancing requirements for TiDB
+
+The requirements that TiDB puts on loadbalancers are in some ways different that what a typical MySQL setup requires.
+
+Where loadbalancers might offer advanced features like read/write splitting for MySQL, this is not needed for TiDB. The loadbalancer does not need to be able to inspect the MySQL protocol. A TCP level loadbalacer works fine for TiDB.
+
+The loadbalancer only needs access to the TiDB servers (TCP port 4000 by default) and doesn't need access to PD or TiKV servers. When using the [status API (TCP port 10080 by default)](https://docs.pingcap.com/tidb/stable/tidb-monitoring-api#use-the-state-interface) is also required.
+
+It is recommended to use `http://<ip_of_tidb>:10080/status` as the health check since it supports the use of the _graceful shutdown_ feature. This feature allows you to first drain client connections before shutting down a TiDB Server and reduces the impact on applications. The [`graceful-wait-before-shutdown`](/tidb-configuration-file.md#graceful-wait-before-shutdown-new-in-v50) variable needs to be set to a non-zero value to benefit from this feature.
+
+Health checks that check for `read_only` or other MySQL or InnoDB specific variables should be avoided. This is because these variables are often implemented as "noop" on TiDB.
+
+## Where to put your loadbalancer
+
+For software-based loadbalancers you have the option to put your loadbalancer on the same host as your application or on separate hosts. Both have their own benefits.
+
+Installing the loadbalancer on the same host as a _sidecar_ to your application is good for performance as it doesn't need an extra network hop. However having to administer many instances of your loadbalancing software makes this slightly more complicated to administer.
+
+Installing the loadbalancer on separate hosts will typically increase latency as an extra network hop is required. There is also risk that (depending on your network) connectivity to and from the load balancer could become saturated. However, this method is less complicated to administer.
+
+High availability is also an important consideration when deciding between these options. With the first setup an unavailable loadbalancer only impacts a single application host. With the second solution you probably have to take some steps to make your loadbalancer software high-available, which complicates the configuration.
+
+## Testing loadbalancing
+
+Beore deploying loadbalancing in production it is advised to first test the configuration to make sure everything works as expected. It is also advised to test bigger configuration changes, upgrades and other maintenance tasks before doing these on a production setup.
+
+Local testing can be done with `tiup playground --db 2` to have two TiDB instances.
+
+When connecting via a loadbalancer you might want to use this query to check on which instance you landed:
+
+```
+SELECT @@hostname, @@port
+```
+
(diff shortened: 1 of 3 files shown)


dveeden:
We could link this to https://docs.pingcap.com/tidb/stable/java-app-best-practices

TomShawn:
/label needs-cherry-pick-5.0
/translation doing
/assign @Liuxiaozhen12 
@dveeden Could you please involve technical review(s)? Thanks!

TomShawn:
> We could link this to https://docs.pingcap.com/tidb/stable/java-app-best-practices

You can link this to these docs in this PR.

loadbalancing.md
```diff
@@ -0,0 +1,339 @@
...
+
+The last type of loadbalancer that we will discuss here is hardware based loadbalancers. These are physical machines where you plug in a network cable. These are often costly, but also offer a high throughput.
+
+## General loadbalancer requirements for TiDB
+
+Where loadbalancers may offer advanced features like read/write splitting for MySQL, this is not needed for TiDB. The loadbalancer does not need to be able to inspect the MySQL protocol. A TCP level loadbalacer works fine for TiDB.
+
+The loadbalancer only needs access to the TiDB servers (TCP port 4000 by default) and doesn't need access to PD or TiKV servers.
+
+In general a TCP health check is sufficient. Health checks that check for `read_only` or other MySQL or InnoDB variables should be avoided.
```
morgo, line 32: I think a HTTP health check is better along with configuration of `GracefulWaitBeforeShutdown`: https://github.com/pingcap/tidb/pull/20649

This helps ensure that topology changes have less impact on users sessions.

dveeden: I couldn't find this option in the docs, so I opened https://github.com/pingcap/docs/issues/5823

morgo: You can also get into the same trouble if it is a duration, since an integer is valid and will be interpreted in nanoseconds :-)

loadbalancing.md
```diff
@@ -0,0 +1,345 @@
...
+
+Another option is the HTTP status API that is by default available on `http://<ip_of_tidb>:10080/status`. The `graceful-wait-before-shutdown` variable can be used to configure tidb to wait for a number of seconds between starting to shutdown and really closing network connections. In this time the HTTP healthcheck will already fail allowing loadbalancers to start and redirect traffic elsewhere.
```
morgo, line 34: I think we can remove this sentence if we use the suggestion above. We should just recommend using this first.

dveeden: Updated. PTAL

loadbalancing.md
```diff
@@ -0,0 +1,345 @@
...
+
+Health checks that check for `read_only` or other MySQL or InnoDB variables should be avoided.
+
+## Where to put your loadbalancer
+
+For software based loadbalancers you have the option to put your loadbalancer on the same hosts as your application or on separate hosts. Both have their own benefits.
+
+Putting the loadbalancer on the same hosts as your application is good for performance as it doesn't need an extra network hop. However having to adminster many instances of your loadbalancing software makes this slightly more complicated to administer.
+
+Putting the loadbalancer on separate machines may be less performant as it needs an extra network hop and depending on your traffic may become a choke-point. However for administration this is might be easier.
```
morgo, line 42: Performant is kind of jargon. It is well understood but I don't think we should use it in docs.

The style-guide is also to s/may/might/ because may has multiple meanings. But I think in this context we could even say something like "will typically increase network latency" because I think it is unequivocal? The edge case would be something weird like a routing error (I also changed it to latency, since there is no guarantee throughput will drop, but it probably will accordingly for many workloads because the protocol is synchronous).

s/machines/hosts/ as physical is not implied, and host is used earlier.

```suggestion
Installing the loadbalancer on separate hosts will typically increase latency as an extra network hop is required. There is also risk that (depending on your network) connectivity to and from the load balancer could become saturated. However, this method is less complicated to administer.
```

loadbalancing.md
```diff
@@ -0,0 +1,345 @@
...
+
+Don't forget to include High-Availablility in your decision. With the first setup an unavailable loadbalancer only impacts a single application host. With the second solution you probably have to take some steps to make your loadbalancer software high-available, which complicates the configuration.
```
morgo, line 44: Maybe consider rewriting the above to use ordinal bullet points so the text below makes more sense?

I suggest removing "Don't forget" because it can make the user worry about perils. Both are accepted solutions, so I think we can just say "this should be a factor in your decision".

```suggestion
High availability is also an important consideration when deciding between these options. With the first setup an unavailable loadbalancer only impacts a single application host. With the second solution you probably have to take some steps to make your loadbalancer software high-available, which complicates the configuration.
```

morgo (approved):

dveeden:
PTAL @Liuxiaozhen12 @TomShawn

Ehco1996 (commented):
LGTM

loadbalancing.md
```diff
@@ -0,0 +1,345 @@
...
+
+Another common option is to use a software based loadbalancer. Common ones are ProxySQL, haproxy and MySQL Router.
+
+The last type of loadbalancing that we will discuss here is hardware based loadbalancers. These are physical machines where you plug in a network cable. These are often costly, but also offer a high throughput.
+
+## General loadbalancing requirements for TiDB
+
+Where loadbalancers may offer advanced features like read/write splitting for MySQL, this is not needed for TiDB. The loadbalancer does not need to be able to inspect the MySQL protocol. A TCP level loadbalacer works fine for TiDB.
+
+The loadbalancer only needs access to the TiDB servers (TCP port 4000 by default) and doesn't need access to PD or TiKV servers. When using the HTTP status access to TCP port 10080 is also required.
```
bb7133, line 30:
```suggestion
The loadbalancer only needs access to the TiDB servers (TCP port 4000 by default) and doesn't need access to PD or TiKV servers. When using the [status API (TCP port 10080 by default)](https://docs.pingcap.com/tidb/stable/tidb-monitoring-api#use-the-state-interface) is also required.
```

bb7133: IMO it's better to introduce the hyperlink here since 'HTTP API' may be new to many readers.

loadbalancing.md
```diff
@@ -0,0 +1,345 @@
...
+
+It is recommended to use `http://<ip_of_tidb>:10080/status` as the health check since it supports the use of the _graceful shutdown_ feature. This feature allows you to first drain client connections before shutting down a TiDB Server and reduces the impact on applications. The [`graceful-wait-before-shutdown`](/tidb-configuration-file.md#graceful-wait-before-shutdown-new-in-v50) variable needs to be set to a non-zero value to benefit from this feature.
+
+Health checks that check for `read_only` or other MySQL or InnoDB variables should be avoided.
```
bb7133, line 34: - Is it common to do the health check by accessing system variables?(I know little about this)

- If using system variables like `read_only` is not recommended, it's better to add an explanation here, for example, "because they are [no-op](https://docs.pingcap.com/tidb/stable/system-variables#tidb_enable_noop_functions-new-in-v40) in TiDB'?"

loadbalancing.md
```diff
@@ -0,0 +1,345 @@
...
+        }
+    }
+}
+```
+
+See [Best Practices for Developing Java Applications with TiDB](/best-practices/java-app-best-practices.md) for more details about how to use TiDB with Java Applications.
+
+### MySQL Connector/Python
+
+In this example I use MySQL Connector/Python 8.0.24 (`mysql-connector-python==8.0.24` according to `pip freeze` output).
```
bb7133, line 257: IMHO It's better to avoid the first-person pronoun "I" in this document. Could you use the passive voice here like:

"In this example MySQL Connector/Python 8.0.24 (`mysql-connector-python==8.0.24` according to `pip freeze` output) are used."

loadbalancing.md
dveeden: ProxySQL uses `read_only` to move servers between read and write pools: https://proxysql.com/documentation/backend-monitoring/

[this](https://severalnines.com/resources/database-management-tutorials/mysql-load-balancing-haproxy-tutorial) tutorial from severalnines also has a script that uses `read_only` to check if a master is "healty" or not.

For Galera based clusters [`wsrep_ready`](https://www.percona.com/doc/percona-xtradb-cluster/5.7/wsrep-status-index.html#wsrep_ready) is often used.

Besides this using [`Innodb_buffer_pool_load_status`](https://dev.mysql.com/doc/refman/8.0/en/server-status-variables.html#statvar_Innodb_buffer_pool_load_status) is also a common practice to only allow servers to enter a pool once they finished loading the pages that were in the buffer pool before the shutdown. This is to avoid cold caches.

loadbalancing.md
```diff
@@ -0,0 +1,345 @@
...
+    "listening_port": 4001,
+    "status_port": 10081,
+    "lease": "45s",
+    "binlog_status": "Off",
+    "start_timestamp": 1620024407,
+    "labels": {},
+    "server_id": 0
+  }
+}
+```
```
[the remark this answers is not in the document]
dveeden: markdownlint doesn't seem to allow that.

dveeden:
PTAL @bb7133

bb7133 (approved):

dveeden:
@TomShawn PTAL

loadbalancing.md
```diff
@@ -0,0 +1,345 @@
...
+
+## Loadbalancing types
+
+The first type of loadbalancing that is available is *connector based loadbalancing*. Many of the MySQL connectors like MySQL Connector/J and MySQL Connector/C++. The benefits from this are that there is no extra network hop and the application has more information about which server it is connected to. The drawback of this is that there is no central administration and when changing the configuration you have to do this on all your application hosts. Depending on the programming language you are using the loadbalancing may not be available or not offer more advanced options. This can work with most third-party applications as this is often configured in the connection string (JDBC url for Java for example).
+
+It is also possible to use DNS round-robin. However this is not advisable as this won't prevent connections from going to a machine that is unavailable. This means that your application may have to re-try connections more often if not all servers are available.
```
TomShawn, line 14: If this is not advisable, then we should not document it here.

TomShawn:
@dveeden This is really a great article with a comprehensive introduction to loadbalancing. These are my general comments. PTAL for reference.

Some part of it, to me, looks like a blog post article that uses plain passages than a user document which usually prefers listing things out in unordered/ordered lists or tables (in a really structured layout). For example, the "Loadbalancing types" section introduces several types. It might be better to organize the information of this passage using a table, which makes your information clearer and more readable to readers. 

|  Loadbalancing types |    User scenarios      |   Benefits   |    Drawbacks  |
|    :---------                   |    :--------------         |    :------      |      :----------   |
|    Connector-based loadbalancing, such as  MySQL Connector/J and MySQL Connector/C++ |  This can work with most third-party applications because this is often configured in the connection string (JDBC url for Java for example). |  There is no extra network hop and the application has more information about which server it is connected to.  |  There is no central administration. When changing the configuration on one host, you have to do this on all your application hosts. |
| DNS round-robin | 

Another question is the target audience. To me, this can be confusing. I think it would be better to clarify what this article introduces and **its aim** at the very beginning. Then readers will get a general idea of the article from the beginning and decide whether or not to go on reading.

+ If it is about the best practices of loadbalancing in TiDB, then maybe we should not introduce too much about the loadbalancing types (including the benefit and drawback, even for the unrecommended types). And its TOC position can also be changed. "Best Practices" might be a better place for it.
+ If it is about a task-oriented instruction/guide for TiDB maintenance, then maybe many of the concept introductions look unnecessary and can be removed.

Then it is about the language styling issue. In user documentation, we usually follow these styles:

+ Avoid using the first person (I/me/we/us/our/let's). Mostly, use "You" or passive voice.
+ Avoid abbreviations (let's, don't, e.g., etc...)
+ Use "may" because it is ambiguous. Use "might" instead.
+ Make sentences as short and concise as possible.
+ **Briefly introduce/summarize each section in one or two sentences at the beginning of the section.**

loadbalancing.md
dveeden: As DNS-RR is a common practice I think we should explain here why other solutions are better.

@morgo what's your take on this?

morgo: I agree with @dveeden , it's very common and users need to know why we don't recommend it.

TomShawn: OK, let's keep it.

dveeden:
@TomShawn I went over your comments.

For the loadbalancer types I don't know if we should:

- Use a table (as you suggested). This gets quite long
- Use paragraphs (as the original had) 
- Use a table an paragraphs (current state) This avoids a table with very high/long cells.
- Use subsections per type

What do you think?

dveeden:
In some places (pingcap/tidb, pingcap/docs, pingcap/tidb-operator) the solution with envoy an istio is used. However I don't think that should be part of this PR as that's mostly a TiDB Operator thing.

dveeden:
This has been published as [a blog](https://pingcap.com/blog/best-practices-for-tidb-load-balancing) instead of as part of the documentation.

PR #5600, opened 2021-05-11
read-historical-data: Add practical advise about how to use this feature

dveeden:

Add practical advise to the read-historical-data page and link to the related dumpling docs

### DIFF
--- a/read-historical-data.md
+++ b/read-historical-data.md
@@ -165,4 +165,6 @@ For simple cases you could use `SELECT` after setting the `tidb_snapshot` variab
 
 - Use [Dumpling](/dumpling-overview.md#export-historical-data-snapshot-of-tidb) to export a historical snapshot. Dumpling performs well in exporting larger sets of data.
 
-If you plan to use this feature to do restores you probably want to increase the GC life time from the default 10m to something like half an hour. This will result in additional versions of rows being retained, which is likely to require slightly more disk space. The performance of certain options such as scans may also degrade, when additional versions of the same rows need to be skipped when reading.
+> **Note:**
+>
+> Increasing the GC life time from the default 10 minutes to half an hour or more will result in additional versions of rows being retained, which might require more disk space. This might also affect the performance of certain operations such as scans when TiDB needs to skip these additional versions of the same rows during data reads.
--- a/read-historical-data.md
+++ b/read-historical-data.md
@@ -153,7 +153,7 @@ Pay special attention to the following:
     >
     > You should use `@@` instead of `@` before `tidb_snapshot` because `@@` is used to denote the system variable while `@` is used to denote the user variable.
 
-## Restoring old versions
+## Restoring historical versions
 
 Before you restore data from an older version, make sure that Garbage Collection (GC) does not clear the history data while you are working on it. This can be done by setting the `tidb_gc_life_time` variable as the following example shows. Do not forget to set the variable back to the previous value after the restore.
 
--- a/read-historical-data.md
+++ b/read-historical-data.md
@@ -161,7 +161,9 @@ Before you restore data from an older version, make sure that Garbage Collection
 SET GLOBAL tidb_gc_life_time="60m";
 ```
 
-For simple cases you could use `SELECT` after setting the `tidb_snapshot` variable and copy-paste the output or use `SELECT ... INTO LOCAL OUTFLE` and use `LOAD DATA` to import the data later on.
+To restore data from an older version, you can use one of the following methods:
+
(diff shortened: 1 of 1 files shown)


bb7133:
But I'm not a reviewer.

bb7133:
PTAL @morgo

read-historical-data.md
```diff
@@ -152,3 +152,17 @@ Pay special attention to the following:
     > **Note:**
     >
     > You should use `@@` instead of `@` before `tidb_snapshot` because `@@` is used to denote the system variable while `@` is used to denote the user variable.
+
+## Restoring old versions
+
+The first thing you probably want to do when you need to restore from an older version is to make sure GC doesn't remove it while you are working on it. This can be done by setting the GC as is done in the example below. Don't forget to set it back to the previous value after the restore.
+
+```sql
+SET GLOBAL tidb_gc_life_time="60m";
```
morgo, line 161: Note for other reviewers: this syntax is new to TiBB 5.0 (but that's OK, this PR does not target older versions).

read-historical-data.md
```diff
@@ -152,3 +152,17 @@ Pay special attention to the following:
...
+```
+
+For simple cases you could use `SELECT` after setting the `tidb_snapshot` variable and copy-paste the output or use `SELECT ... INTO LOCAL OUTFLE` and use `LOAD DATA` to import the data later on.
+
+Another option is to use [dumpling to export a historical snapshot](/dumpling-overview.md#export-historical-data-snapshot-of-tidb). This is a good tool for exporting larger sets of data.
+
+If you plan to use this feature to do restores you probably want to increase the GC life time from the default 10m to something like half an hour.
```
morgo, line 168: Please state the risks when doing this. For example by adding a sentence:

"This will result in additional versions of rows being retained, which may increase disk space requirements slightly. The performance of certain options such as scans may also degrade, when additional versions of the same rows need to be skipped when reading".

morgo (commented):
Just a small comment, otherwise LGTM.

read-historical-data.md
```diff
@@ -152,3 +152,17 @@ Pay special attention to the following:
...
+
+```sql
+SET GLOBAL tidb_gc_life_time="60m";
+```
+
+For simple cases you could use `SELECT` after setting the `tidb_snapshot` variable and copy-paste the output or use `SELECT ... INTO LOCAL OUTFLE` and use `LOAD DATA` to import the data later on.
+
+Another option is to use [dumpling to export a historical snapshot](/dumpling-overview.md#export-historical-data-snapshot-of-tidb). This is a good tool for exporting larger sets of data.
+
+If you plan to use this feature to do restores you probably want to increase the GC life time from the default 10m to something like half an hour. This will result in additional versions of rows being retained, which is likely to require slightly more disk space. The performance of certain options such as scans may also degrade, when additional versions of the same rows need to be skipped when reading.
```
TomShawn, line 168:
```suggestion
> **Note:**
>
> Increasing the GC life time from the default 10 minutes to half an hour or more will result in additional versions of rows being retained, which might require more disk space. This might also affect the performance of certain operations such as scans when TiDB needs to skip these additional versions of the same rows during data reads.
```

read-historical-data.md
```diff
@@ -152,3 +152,21 @@ Pay special attention to the following:
...
+
+To restore data from an older version, you can use one of the following methods:
+
+- For simple cases, use `SELECT` after setting the `tidb_snapshot` variable and copy-paste the output, or use `SELECT ... INTO LOCAL OUTFLE` and use `LOAD DATA` to import the data later on.
+
+- Use [Dumpling](/dumpling-overview.md#export-historical-data-snapshot-of-tidb) to export a historical snapshot. Dumpling performs well in exporting larger sets of data.
+
+> **Note:**
+>
+> Increasing the GC life time from the default 10 minutes to half an hour or more will result in additional versions of rows being retained, which might require more disk space. This might also affect the performance of certain operations such as scans when TiDB needs to skip these additional versions of the same rows during data reads.
```
TomShawn, line 172: Would it be better to move this note to Line 163? Line 163 is right under how to set the GC life time.

read-historical-data.md
```diff
@@ -152,3 +152,21 @@ Pay special attention to the following:
     > **Note:**
     >
     > You should use `@@` instead of `@` before `tidb_snapshot` because `@@` is used to denote the system variable while `@` is used to denote the user variable.
+
+## Restoring historical versions
```
TomShawn, line 156: For task-oriented topics in our documentation, we usually use headings that start with the prototype of verbs OR "how to". In this case, it is recommended to use "Restore historical data" or "How to restore historical data".

PR #5647, opened 2021-05-19
Add documentation for SEM

morgo:

This adds the docs for https://github.com/pingcap/tidb/blob/master/docs/design/2021-03-09-security-enhanced-mode.md

Which was merged in a series of PRs to master. The exact restrictions may evolve over time, so the wording is kept to use simple language to say "restrictions include". It is intended for use in cloud environments where the "root" privilege is considered too powerful.

This only affects master.

https://github.com/pingcap/tidb/pull/24416
https://github.com/pingcap/tidb/pull/24412
https://github.com/pingcap/tidb/pull/24279
https://github.com/pingcap/tidb/pull/23978

- This PR is translated from:
- Other reference link(s): https://github.com/pingcap/tidb/blob/master/docs/design/2021-03-09-security-enhanced-mode.md

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/system-variables.md
+++ b/system-variables.md
@@ -427,6 +427,17 @@ Constraint checking is always performed in place for pessimistic transactions (d
 - Default value: ON
 - This variable controls whether to record the execution information of each operator in the slow query log.
 
+### tidb_enable_enhanced_security
+
+- Scope: NONE
+- Default value: OFF
+- This variable indicates if the TiDB server you are connected to was started with Security Enhanced Mode (SEM) enabled.
+- SEM is inspired by the design of systems such as [Security-Enhanced Linux](https://en.wikipedia.org/wiki/Security-Enhanced_Linux). It reduces the capabilities of users with the MySQL `SUPER` privilege, and instead requires `RESTRICTED` fine grained privileges to be granted as a replacement. These include:
+    - `RESTRICTED_TABLES_ADMIN`: The ability to write to system tables in the `mysql` schema, and see sensitive columns on `information_schema` tables.
+    - `RESTRICTED_STATUS_ADMIN`: The ability to see sensitive variables in the command `SHOW STATUS`.
+    - `RESTRICTED_VARIABLES_ADMIN`: The ability to see and set sensitive variables in `SHOW [GLOBAL] VARIABLES` and `SET`.
+    - `RESTRICTED_USER_ADMIN`: The ability to prevent other users from making changes or dropping a user account.
+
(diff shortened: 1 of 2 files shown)


bb7133:
PTAL @SunRunAway

system-variables.md
```diff
@@ -427,6 +427,17 @@ Constraint checking is always performed in place for pessimistic transactions (d
 - Default value: ON
 - This variable controls whether to record the execution information of each operator in the slow query log.
 
+### tidb_enable_enhanced_security
```
SunRunAway, line 430: Is this variable readonly?

SunRunAway: @morgo PTAL

morgo: Yes it is. Or more specifically it is `Scope: None`, which is the same as read-only.

SunRunAway: > TiDB system variables behave similar to MySQL with some differences, in that settings might apply on a SESSION, INSTANCE, or GLOBAL scope, or on a scope that combines SESSION, INSTANCE, or GLOBAL.

> * Changes to GLOBAL scoped variables only apply to new connection sessions with TiDB. Currently active connection sessions are not affected. These changes are persisted and valid after restarts.
> * Changes to INSTANCE scoped variables apply to all active or new connection sessions with the current TiDB instance immediately after the changes are made. Other TiDB instances are not affected. These changes are not persisted and become invalid after TiDB restarts.

The document does not mention any about what `Scope: None` means. Cloud you add some explanations?

morgo: I agree with you. I will change it tomorrow to say Read Only instead.

(There are some minor differences, but they don't matter to users.)

morgo: This is improved in https://github.com/pingcap/docs/pull/5647/commits/c8758d7b8c130ed8c52aa331e27110c383fed166 , PTAL again thx :-)

SunRunAway (approved):

TomShawn:
Follow-up: Minor wording issue

system-variables.md
```diff
@@ -427,6 +428,17 @@ Constraint checking is always performed in place for pessimistic transactions (d
...
+
+- Scope: NONE
+- Default value: OFF
+- This variable indicates if the TiDB server you are connected to has Security Enhanced Mode (SEM) enabled, and can not be changed without restarting the TiDB server.
```
TomShawn, line 435: @morgo Can we say "To change its value, you need to restart the TiDB server for the change to be effective"?

PR #5720, opened 2021-05-28
Auto-generate sysvar documentation from TiDB server source

morgo:

This auto-generates the system variables documentation by using the TiDB server's source code. The advantage of doing this is that system-variables will be guaranteed accurate, and non-documented sysvars will be automatically discovered.

It has a few small differences in formatting, but I've tried to keep it as close to the current document as makes sense.

** This contributes the script for future use by others. Generated system-variables.md files will be contributed separately. **

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases


TomShawn:
@morgo I guess this PR also applies to release-5.0, because the most majority of the content in release-5.0 is the same as that in master.

morgo:
> @morgo This is just awesome ! I guess this PR also applies to release-5.0, because the most majority of the content in release-5.0 is the same with that in master.

That's actually complicated. The TiDB source code in release-5.0 does not support the required features to auto-generate. Only master does. I have also been [fixing bugs ](https://github.com/pingcap/tidb/pull/24964)based on the possible range of numeric types, so I think it's better to draw a line in the sand and say it only applies to docs in 5.1+

dveeden (approved):
To get this to build I had to let go modify the `go.mod` file:
```patch
diff --git a/scripts/go.mod b/scripts/go.mod
index 64727a0d6..a0d52ca8c 100644
--- a/scripts/go.mod
+++ b/scripts/go.mod
@@ -10,6 +10,7 @@ require (
        github.com/danjacques/gofslock v0.0.0-20200623023034-5d0bd0fa6ef0 // indirect
        github.com/dgryski/go-farm v0.0.0-20200201041132-a6ae2369ad13 // indirect
        github.com/go-ole/go-ole v1.2.6 // indirect
+       github.com/go-sql-driver/mysql v1.6.0 // indirect
        github.com/golang/snappy v0.0.4 // indirect
        github.com/google/btree v1.0.1 // indirect
        github.com/google/uuid v1.3.0 // indirect
```

Might be good to (eventually) add a CI job to do something like this:
```bash
cd scripts
diff ../system-variables.md $<(go run generate-system-variables.go)
```

scripts/go.mod
```diff
@@ -0,0 +1,46 @@
+module github.com/morgo/docs/scripts
```
dveeden, line 1:
```suggestion
module github.com/pingcap/docs/scripts
```

morgo:
My hope is that we do more work on this (add a chinese translation version) before merging it. I pushed the changes I have previously made locally so I could show the product manager (Eason) who is tasked with configuration.

qiancai:
Removed the needs-cherry-pick-release-6.6 label because the v6.6 docs have been archived at https://docs-archive.pingcap.com/tidb/v6.6 and will no longer receive new updates.

dveeden:
@morgo @qiancai I've updated this code in a branch: https://github.com/dveeden/docs/tree/fix-sysvar-correctness-202608

What direction should we take?
1. Close this PR
2. Update this PR and let it generate the sysvar docs automatically
3. Modify this PR to make it check specific things in the sysvar docs (e.g. correct options, correct units, etc)

PR #5729, opened 2021-05-31
tikv: add configuration docs for I/O rate limiter

tabokie:
Signed-off-by: tabokie <xy.tao@outlook.com>

- [x] I've signed [**Contributor License Agreement**](https://cla-assistant.io/pingcap/docs) that's required for repo owners to accept my contribution.

Add corresponding docs change to tikv/tikv#9197.

No.

### DIFF
--- a/tikv-configuration-file.md
+++ b/tikv-configuration-file.md
@@ -300,6 +300,20 @@ Configuration items related to the sharing of block cache among multiple RocksDB
 + Default value: 45% of the size of total system memory
 + Unit: KB|MB|GB
 
+## storage.io-rate-limit
+
+Configuration items related to the I/O rate limiter.
+
+### `max-bytes-per-sec`
+
++ Maximum I/O bytes that this server can write to or read from disk (determined by mode) in one second. Internally it prefers throttling background operations over foreground ones. This value should be set to the disk's optimal IO bandwidth, e.g. maximum IO bandwidth specified by cloud disk vendors. When set to zero, disk IO operations are not limited.
++ Default value: `"0MB"`
+
+### `mode`
+
++ Determine which types of IO operations are counted and restrained below threshold. Three different modes are: write-only, read-only, all-io. Only write-only mode is supported for now.
++ Default value: `"write-only"`
+
 ## raftstore
 
 Configuration items related to Raftstore
--- a/tikv-configuration-file.md
+++ b/tikv-configuration-file.md
@@ -311,7 +311,8 @@ Configuration items related to the I/O rate limiter.
 
 ### `mode`
 
-+ Determine which types of IO operations are counted and restrained below threshold. Three different modes are: write-only, read-only, all-io. Only write-only mode is supported for now.
++ Determine which types of IO operations are counted and restrained below threshold. Only write-only mode is supported for now.
++ Optional value: `"write-only"`, `"read-only"`, `"all-io"`
 + Default value: `"write-only"`
 
 ## raftstore
--- a/tikv-configuration-file.md
+++ b/tikv-configuration-file.md
@@ -306,13 +306,13 @@ Configuration items related to the I/O rate limiter.
 
 ### `max-bytes-per-sec`
 
-+ Maximum I/O bytes that this server can write to or read from disk (determined by mode) in one second. Internally it prefers throttling background operations over foreground ones. This value should be set to the disk's optimal IO bandwidth, e.g. maximum IO bandwidth specified by cloud disk vendors. When set to zero, disk IO operations are not limited.
++ Limits the maximum I/O bytes that a server can write to or read from the disk (determined by the `mode` configuration item below) in one second. With this limit is reached, TiKV prefers throttling background operations over foreground ones. The value of this configuration item should be set to the disk's optimal I/O bandwidth, for example, the maximum I/O bandwidth specified by your cloud disk vendor. When this configuration value is set to zero, disk I/O operations are not limited.
 + Default value: `"0MB"`
 
 ### `mode`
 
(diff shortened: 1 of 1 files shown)


tikv-configuration-file.md
```diff
@@ -300,6 +300,20 @@ Configuration items related to the sharing of block cache among multiple RocksDB
...
 + Unit: KB|MB|GB
 
+## storage.io-rate-limit
+
+Configuration items related to the I/O rate limiter.
+
+### `max-bytes-per-sec`
+
++ Maximum I/O bytes that this server can write to or read from disk (determined by mode) in one second. Internally it prefers throttling background operations over foreground ones. This value should be set to the disk's optimal IO bandwidth, e.g. maximum IO bandwidth specified by cloud disk vendors. When set to zero, disk IO operations are not limited.
++ Default value: `"0MB"`
```
yiwu-arbug, line 310: Commented on the other PR, can we use empty value to denote no IO limiter, instead of "0MB"?

yiwu-arbug (commented):
rest looking good

tikv-configuration-file.md
TomShawn: @tabokie Any update?

tabokie: @TomShawn Discussed offline, this change is not needed now.

yiwu-arbug (approved):

tikv-configuration-file.md
```diff
@@ -300,6 +300,21 @@ Configuration items related to the sharing of block cache among multiple RocksDB
...
+
+### `mode`
+
++ Determine which types of IO operations are counted and restrained below threshold. Only write-only mode is supported for now.
```
TomShawn, line 314: Does the "threshold" refer to max-bytes-per-sec?
```suggestion
+ Determines which types of I/O operations are counted and restrained below the threshold. Only the write-only mode is supported for now.
```

tikv-configuration-file.md
```diff
@@ -300,6 +300,21 @@ Configuration items related to the sharing of block cache among multiple RocksDB
...
++ Optional value: `"write-only"`, `"read-only"`, `"all-io"`
```
TomShawn, line 315: If  the write-only mode is supported for now, what about removing `"read-only"` and `"all-io"` for now?
```suggestion
+ Value options: `"write-only"`, `"read-only"`, `"all-io"`
```

tikv-configuration-file.md
```diff
@@ -300,6 +300,21 @@ Configuration items related to the sharing of block cache among multiple RocksDB
 + Default value: 45% of the size of total system memory
 + Unit: KB|MB|GB
 
+## storage.io-rate-limit
+
+Configuration items related to the I/O rate limiter.
+
+### `max-bytes-per-sec`
+
++ Maximum I/O bytes that this server can write to or read from disk (determined by mode) in one second. Internally it prefers throttling background operations over foreground ones. This value should be set to the disk's optimal IO bandwidth, e.g. maximum IO bandwidth specified by cloud disk vendors. When set to zero, disk IO operations are not limited.
```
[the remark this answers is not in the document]
tabokie: This is actually an important feature of the new limiter. Instead of indiscriminately limit all IO types, it will throttle background operations first (e.g. compaction, import).

tikv-configuration-file.md
tabokie: Yes, do we need to substitute it here?

tikv-configuration-file.md
tabokie: I'm not sure. If so this configuration item doesn't make a lot of sense to exist. Might as well remove it? What do you think?

tabokie: And BTW the expression `Optional value` is used several times before in this document. Do we need to change them as well?

TomShawn: + If it affects the user behavior, this config should be documented. If `"read-only"` and `"all-io"` are currently not supported, we should remove them and add them back when available.
+ OK. I've deleted the suggested comment. Thanks for reminding~

tikv-configuration-file.md
TomShawn: Yes, to make it explicit and clear.

tikv-configuration-file.md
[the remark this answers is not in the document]
TomShawn: Thanks for explaining. I've modify my comment. PTAL again~

tikv-configuration-file.md
```diff
@@ -300,6 +300,21 @@ Configuration items related to the sharing of block cache among multiple RocksDB
 + Default value: 45% of the size of total system memory
 + Unit: KB|MB|GB
 
+## storage.io-rate-limit
+
+Configuration items related to the I/O rate limiter.
+
+### `max-bytes-per-sec`
+
++ Limits the maximum I/O bytes that a server can write to or read from the disk (determined by the `mode` configuration item below) in one second. With this limit is reached, TiKV prefers throttling background operations over foreground ones. The value of this configuration item should be set to the disk's optimal I/O bandwidth, for example, the maximum I/O bandwidth specified by your cloud disk vendor. When this configuration value is set to zero, disk I/O operations are not limited.
```
TomShawn, line 309:
```suggestion
+ Limits the maximum I/O bytes that a server can write to or read from the disk (determined by the `mode` configuration item below) in one second. When this limit is reached, TiKV prefers throttling background operations over foreground ones. The value of this configuration item should be set to the disk's optimal I/O bandwidth, for example, the maximum I/O bandwidth specified by your cloud disk vendor. When this configuration value is set to zero, disk I/O operations are not limited.
```

TomShawn (approved):
LGTM

PR #5752, opened 2021-06-03
system-variables: update from generated source

morgo:

This updates the system variables, generated against the tidb source code. i.e. this PR is derived from https://github.com/pingcap/docs/pull/5720 , but the generator source code will be submitted separately.

This allows this PR to potentially be cherry picked and handled like a regular contribution.

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/system-variables.md
+++ b/system-variables.md
@@ -113,6 +113,7 @@ mysql> SELECT * FROM t1;
 - Scope: SESSION | GLOBAL
 - Default value: 28800
 - Range: [1, 31536000]
+- Unit: seconds
 - This variable represents the idle timeout of the interactive user session, which is measured in seconds. Interactive user session refers to the session established by calling [`mysql_real_connect()`](https://dev.mysql.com/doc/c-api/5.7/en/mysql-real-connect.html) API using the `CLIENT_INTERACTIVE` option (for example, MySQL shell client). This variable is fully compatible with MySQL.
 
 ### last_plan_from_binding <span class="version-mark">New in v4.0</span>
@@ -132,6 +133,7 @@ mysql> SELECT * FROM t1;
 - Scope: SESSION | GLOBAL
 - Default value: 0
 - Range: [0, 2147483647]
+- Unit: milliseconds
 - The maximum execution time of a statement in milliseconds. The default value is unlimited (zero).
 
 > **Note:**
@@ -252,7 +254,8 @@ mysql> SELECT * FROM t1;
 - Scope: SESSION | GLOBAL
 - Default value: 104857600
 - Range: [0, 9223372036854775807]
-- If the table size (in the unit of bytes) is less than the value of the variable, the Broadcast Hash Join algorithm is used. Otherwise, the Shuffled Hash Join algorithm is used. The default of 104857600 equals 100 megabytes.
+- Unit: Bytes
+- If the table size is less than the value of the variable, the Broadcast Hash Join algorithm is used. Otherwise, the Shuffled Hash Join algorithm is used.
 
 ### tidb_build_stats_concurrency
 
@@ -389,7 +392,7 @@ Constraint checking is always performed in place for pessimistic transactions (d
 
 > **Note:**
 >
-> The default of `ON` only applies to new clusters. When upgrading from an earlier version of TiDB, the value `OFF` will be used instead.
+> The default value of `ON` only applies to new clusters. if your cluster was upgraded from an earlier version of TiDB, the value `OFF` will be used instead.
 
 - Scope: SESSION | GLOBAL
 - Default value: ON
@@ -418,7 +421,7 @@ Constraint checking is always performed in place for pessimistic transactions (d
 
 > **Note:**
 >
-> The default of `ON` only applies to new clusters. When upgrading from an earlier version of TiDB, the value `OFF` will be used instead.
+> The default value of `ON` only applies to new clusters. if your cluster was upgraded from an earlier version of TiDB, the value `OFF` will be used instead.
 
 - Scope: SESSION | GLOBAL
 - Default value: ON
@@ -461,9 +464,9 @@ Constraint checking is always performed in place for pessimistic transactions (d
 
 - Scope: NONE
 - Default value: OFF
-- This variable indicates if the TiDB server you are connected to has Security Enhanced Mode (SEM) enabled, and can not be changed without restarting the TiDB server.
-- SEM is inspired by the design of systems such as [Security-Enhanced Linux](https://en.wikipedia.org/wiki/Security-Enhanced_Linux). It reduces the capabilities of users with the MySQL `SUPER` privilege, and instead requires `RESTRICTED` fine grained privileges to be granted as a replacement. These include:
-    - `RESTRICTED_TABLES_ADMIN`: The ability to write to system tables in the `mysql` schema, and see sensitive columns on `information_schema` tables.
+- This variable indicates whether the TiDB server you are connected to has the Security Enhanced Mode (SEM) enabled. To change its value, you need to modify the value of `enable-sem` in your TiDB server configuration file and restart the TiDB server.
+- SEM is inspired by the design of systems such as [Security-Enhanced Linux](https://en.wikipedia.org/wiki/Security-Enhanced_Linux). It reduces the abilities of users with the MySQL `SUPER` privilege and instead requires `RESTRICTED` fine-grained privileges to be granted as a replacement. These fine-grained privileges include:
+    - `RESTRICTED_TABLES_ADMIN`: The ability to write data to system tables in the `mysql` schema and to see sensitive columns on `information_schema` tables.
     - `RESTRICTED_STATUS_ADMIN`: The ability to see sensitive variables in the command `SHOW STATUS`.
     - `RESTRICTED_VARIABLES_ADMIN`: The ability to see and set sensitive variables in `SHOW [GLOBAL] VARIABLES` and `SET`.
     - `RESTRICTED_USER_ADMIN`: The ability to prevent other users from making changes or dropping a user account.
@@ -846,7 +849,8 @@ For a system upgraded to v5.0 from an earlier version, if you have not modified 
 - Scope: SESSION | GLOBAL
 - Default value: 33554432
 - Range: [0, 9223372036854775807]
-- This variable is used to set the memory usage threshold of the local cache in the `Apply` operator in bytes.
+- Unit: Bytes
+- This variable is used to set the memory usage threshold of the local cache in the `Apply` operator.
 - The local cache in the `Apply` operator is used to speed up the computation of the `Apply` operator. You can set the variable to `0` to disable the `Apply` cache feature.
 
 ### tidb_mem_quota_query
@@ -854,7 +858,8 @@ For a system upgraded to v5.0 from an earlier version, if you have not modified 
 - Scope: SESSION
 - Default value: 1073741824
 - Range: [-1, 9223372036854775807]
-- This variable is used to set the threshold value of memory quota for a query in bytes.
+- Unit: Bytes
+- This variable is used to set the threshold value of memory quota for a query.
 - If the memory quota of a query during execution exceeds the threshold value, TiDB performs the operation designated by the OOMAction option in the configuration file. The initial value of this variable is configured by [`mem-quota-query`](/tidb-configuration-file.md#mem-quota-query).
 
 ### tidb_memory_usage_alarm_ratio
@@ -1046,7 +1051,8 @@ explain select * from t where age=5;
 - Scope: INSTANCE
 - Default value: 4096
 - Range: [-1, 9223372036854775807]
-- The maximum length of the SQL statement output in bytes. When the output length of a statement is larger than the `tidb_query-log-max-len` value, the statement is truncated to output.
+- Unit: Bytes
+- The maximum length of the SQL statement output. When the output length of a statement is larger than the `tidb_query-log-max-len` value, the statement is truncated to output.
 
 Usage example:
 
@@ -1135,7 +1141,8 @@ Query OK, 0 rows affected, 1 warning (0.00 sec)
 - Scope: INSTANCE
 - Default value: 300
 - Range: [-1, 9223372036854775807]
-- This variable is used to output the threshold value of the time consumed by the slow log, measured in ms. When the time consumed by a query is larger than this value, this query is considered as a slow log and its log is output to the slow query log.
+- Unit: milliseconds
+- This variable is used to output the threshold value of the time consumed by the slow log. When the time consumed by a query is larger than this value, this query is considered as a slow log and its log is output to the slow query log.
 
 Usage example:
 
@@ -1274,7 +1281,8 @@ This variable is an alias for _transaction_isolation_.
 - Scope: SESSION | GLOBAL
 - Default value: 0
 - Range: [0, 31536000]
-- This variable controls the idle timeout of user sessions in seconds. A zero-value means unlimited.
+- Unit: seconds
+- This variable controls the idle timeout of user sessions. A zero-value means unlimited.
 
 ### windowing_use_high_precision
 
--- a/system-variables.md
+++ b/system-variables.md
@@ -252,7 +252,7 @@ mysql> SELECT * FROM t1;
 ### tidb_broadcast_join_threshold_size <span class="version-mark">New in v5.0</span>
 
 - Scope: SESSION | GLOBAL
-- Default value: `104857600`
+- Default value: `104857600` (100 MiB)
 - Range: `[0, 9223372036854775807]`
 - Unit: Bytes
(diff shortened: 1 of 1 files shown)


system-variables.md
```diff
@@ -434,9 +461,9 @@ Constraint checking is always performed in place for pessimistic transactions (d
 
 - Scope: NONE
 - Default value: OFF
-- This variable indicates whether the TiDB server you are connected to has the Security Enhanced Mode (SEM) enabled. To change its value, you need to modify the value of `enable-sem` in your TiDB server configuration file and restart the TiDB server.
-- SEM is inspired by the design of systems such as [Security-Enhanced Linux](https://en.wikipedia.org/wiki/Security-Enhanced_Linux). It reduces the abilities of users with the MySQL `SUPER` privilege and instead requires `RESTRICTED` fine-grained privileges to be granted as a replacement. These fine-grained privileges include:
-    - `RESTRICTED_TABLES_ADMIN`: The ability to write data to system tables in the `mysql` schema and to see sensitive columns on `information_schema` tables.
+- This variable indicates if the TiDB server you are connected to has Security Enhanced Mode (SEM) enabled, and can not be changed without restarting the TiDB server.
```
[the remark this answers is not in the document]
morgo: Yes, I forgot to update the template. This will be easier to spot when the diffs are smaller in future :-)

system-variables.md
```diff
@@ -750,34 +787,37 @@ For a system upgraded to v5.0 from an earlier version, if you have not modified
...
 - Default value: 32
-- Range: 1 - 32
+- Range: [1, 32]
 - This variable is used to set the number of rows for the initial chunk during the execution process.
 
 ### tidb_isolation_read_engines <span class="version-mark">New in v4.0</span>
 
 - Scope: SESSION
-- Default value: tikv, tiflash, tidb
+- Default value: tikv,tiflash,tidb
```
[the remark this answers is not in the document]
morgo: Yes, it's one value. The previous default had spaces in it, but this was updated in master to appear without spaces which is more consistent with enum value handling as in MySQL.

system-variables.md
```diff
@@ -791,27 +831,30 @@ For a system upgraded to v5.0 from an earlier version, if you have not modified
...
 - Default value: 1024
+- Range: [100, 16384]
 - This variable is used to set the maximum number of schema versions (the table IDs modified for corresponding versions) allowed to be cached. The value range is 100 ~ 16384.
 
 ### tidb_mem_quota_apply_cache <span class="version-mark">New in v5.0</span>
 
 - Scope: SESSION | GLOBAL
-- Default value: 32 MB
-- This variable is used to set the memory usage threshold of the local cache in the `Apply` operator.
+- Default value: 33554432
```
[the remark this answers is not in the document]
morgo: Yes I agree. I have added units for the variables where I know them. It is not specifically tracked internally, so it will not be available on all variables.

system-variables.md
```diff
@@ -137,24 +144,26 @@ mysql> SELECT * FROM t1;
...
 - Scope: NONE
-- Default value: ''
+- Default value: ""
 - The local unix socket file that the `tidb-server` is listening on when speaking the MySQL protocol.
 
 ### sql_mode
 
 - Scope: SESSION | GLOBAL
-- Default value: `ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION`
+- Default value: ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION
```
TomShawn, line 159: I prefer adding the back ticks to make it look like one value.
```suggestion
- Default value: `ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION`
```

system-variables.md
```diff
@@ -103,30 +105,35 @@ mysql> SELECT * FROM t1;
...
-- Default value: 0
+- Default value: OFF
 - This variable is used to show whether the execution plan used in the previous `execute` statement is taken directly from the plan cache.
 
 ### max_execution_time
 
 - Scope: SESSION | GLOBAL
 - Default value: 0
+- Range: [0, 2147483647]
+- Unit: milliseconds
```
TomShawn, line 136:
```suggestion
- Unit: Milliseconds
```

system-variables.md
```diff
@@ -103,30 +105,35 @@ mysql> SELECT * FROM t1;
...
 - Default value: 50
+- Range: [1, 1073741824]
 - The lock wait timeout for pessimistic transactions (default) in seconds.
 
 ### interactive_timeout
 
 - Scope: SESSION | GLOBAL
 - Default value: 28800
+- Range: [1, 31536000]
+- Unit: seconds
```
TomShawn, line 116:
```suggestion
- Unit: Seconds
```

system-variables.md
```diff
@@ -364,13 +384,18 @@ Constraint checking is always performed in place for pessimistic transactions (d
...
 - Default value: 0
+- Range: [0, 2147483647]
 - When this value is greater than `0`, TiDB will batch commit statements such as `INSERT` or `LOAD DATA` into smaller transactions. This reduces memory usage and helps ensure that the `txn-total-size-limit` is not reached by bulk modifications.
 - Only the value `0` provides ACID compliance. Setting this to any other value will break the atomicity and isolation guarantees of TiDB.
 
 ### tidb_enable_1pc <span class="version-mark">New in v5.0</span>
 
+> **Note:**
+>
+> The default value of `ON` only applies to new clusters. if your cluster was upgraded from an earlier version of TiDB, the value `OFF` will be used instead.
```
TomShawn, line 395: The default reading sequence is up -> down. So it's a bit weird that this note is presented before Line 398. What about writing it in Line 398, just after "Default value: ON"?

system-variables.md
```diff
@@ -394,8 +419,12 @@ Constraint checking is always performed in place for pessimistic transactions (d
 
 ### tidb_enable_async_commit <span class="version-mark">New in v5.0</span>
 
+> **Note:**
+>
+> The default value of `ON` only applies to new clusters. if your cluster was upgraded from an earlier version of TiDB, the value `OFF` will be used instead.
```
TomShawn, line 424: ditto

system-variables.md
```diff
@@ -643,7 +678,7 @@ For a system upgraded to v5.0 from an earlier version, if you have not modified
 ### tidb_gc_life_time <span class="version-mark">New in v5.0</span>
 
 - Scope: GLOBAL
-- Default: `"10m0s"`
+- Default value: 10m0s
```
TomShawn, line 681: This value was presented as a "string". If so, I think we can keep the "" quotes. The same for similar cases in this PR.

system-variables.md
```diff
@@ -750,34 +790,37 @@ For a system upgraded to v5.0 from an earlier version, if you have not modified
...
 - Default value: 32
-- Range: 1 - 32
+- Range: [1, 32]
 - This variable is used to set the number of rows for the initial chunk during the execution process.
 
 ### tidb_isolation_read_engines <span class="version-mark">New in v4.0</span>
 
 - Scope: SESSION
-- Default value: tikv, tiflash, tidb
+- Default value: tikv,tiflash,tidb
```
TomShawn, line 823:
```suggestion
- Default value: `tikv,tiflash,tidb`
```

system-variables.md
```diff
@@ -1081,7 +1139,9 @@ Query OK, 0 rows affected, 1 warning (0.00 sec)
 ### tidb_slow_log_threshold
 
 - Scope: INSTANCE
-- Default value: 300ms
+- Default value: 300
+- Range: [-1, 9223372036854775807]
+- Unit: milliseconds
```
TomShawn, line 1144:
```suggestion
- Unit: Milliseconds
```

system-variables.md
```diff
@@ -1219,7 +1280,9 @@ This variable is an alias for _transaction_isolation_.
 
 - Scope: SESSION | GLOBAL
 - Default value: 0
-- This variable controls the idle timeout of user sessions in seconds. A zero-value means unlimited.
+- Range: [0, 31536000]
+- Unit: seconds
```
TomShawn, line 1284:
```suggestion
- Unit: Seconds
```

system-variables.md
morgo: I used this style to match the deprecated sysvars, where we have the warning before the variable options. But warning didn't seem appropriate, so I changed it to note.

system-variables.md
morgo: If we change this to a string, then the other cases would need to be a string too. It's not that this is incorrect - it's just that it doesn't add any value (all sysvars are actually intepreted as strings). See: https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_flush_method for example on how mysql does it.

morgo:
/cc @dveeden may I ask for a technical review?

This is generated from code if you'd like to take a look at https://github.com/pingcap/docs/pull/5720 - but we'll treat it as a my contribution for workflow simplicity.

system-variables.md
```diff
@@ -616,34 +643,36 @@ For a system upgraded to v5.0 from an earlier version, if you have not modified
...
 - Scope: GLOBAL
-- Default value: ON
+- Default value: `ON`
 - Enables garbage collection for TiKV. Disabling garbage collection will reduce system performance, as old versions of rows will no longer be purged.
 
 ### tidb_gc_life_time <span class="version-mark">New in v5.0</span>
 
 - Scope: GLOBAL
-- Default: `"10m0s"`
+- Default value: `10m0s`
```
dveeden, line 675:
```suggestion
- Default value: `10m0s`
- Range: `[10m, 2562047h]`
```

```sql
5.7.25-TiDB-v5.2.0-alpha-16-gf81ef5579-dirty 127.0.0.1:4000  SQL  set global tidb_gc_life_time='10m';
Query OK, 0 rows affected (0.0031 sec)
5.7.25-TiDB-v5.2.0-alpha-16-gf81ef5579-dirty 127.0.0.1:4000  SQL  set global tidb_gc_life_time='9m';
ERROR: 1232 (42000): Incorrect argument type to variable 'tidb_gc_life_time'
5.7.25-TiDB-v5.2.0-alpha-16-gf81ef5579-dirty 127.0.0.1:4000  SQL  set global tidb_gc_life_time='9m59s';
ERROR: 1232 (42000): Incorrect argument type to variable 'tidb_gc_life_time'
5.7.25-TiDB-v5.2.0-alpha-16-gf81ef5579-dirty 127.0.0.1:4000  SQL  set global tidb_gc_life_time='9m60s';
Query OK, 0 rows affected (0.0029 sec)
```

system-variables.md
```diff
@@ -749,95 +783,105 @@ For a system upgraded to v5.0 from an earlier version, if you have not modified
...
-- Default value: 1024
+- Default value: `1024`
+- Range: `[100, 16384]`
 - This variable is used to set the maximum number of schema versions (the table IDs modified for corresponding versions) allowed to be cached. The value range is 100 ~ 16384.
 
 ### tidb_mem_quota_apply_cache <span class="version-mark">New in v5.0</span>
 
 - Scope: SESSION | GLOBAL
-- Default value: 32 MB
+- Default value: `33554432`
```
dveeden, line 844: I think `32 MB` is easier to read. Note that `32 MB` is sort-of wrong as it is `32 MiB`. But as `SET GLOBAL...` doesn't accept values like this and `SHOW GLOBAL VARIABLES` doesn't do that either it is probably ok to do this.

```suggestion
- Default value: `33554432` (32 MiB)
```

system-variables.md
```diff
@@ -1031,34 +1081,36 @@ SET tidb_query_log_max_len = 20
...
 - Note that modifying this variable does not affect the old data that has been saved, but applies the corresponding version format only to the newly written data after modifying this variable.
 
 ### tidb_scatter_region
 
 - Scope: GLOBAL
-- Default value: OFF
+- Default value: `OFF`
 - By default, Regions are split for a new table when it is being created in TiDB. After this variable is enabled, the newly split Regions are scattered immediately during the execution of the `CREATE TABLE` statement. This applies to the scenario where data need to be written in batches right after the tables are created in batches, because the newly split Regions can be scattered in TiKV beforehand and do not have to wait to be scheduled by PD. To ensure the continuous stability of writing data in batches, the `CREATE TABLE` statement returns success only after the Regions are successfully scattered. This makes the statement's execution time multiple times longer than that when you disable this variable.
 
 ### tidb_skip_ascii_check <span class="version-mark">New in v5.0</span>
```
dveeden, line 1103: Should we link to the similar setting for UTF-8?

dveeden (approved):

system-variables.md
```diff
@@ -232,40 +245,43 @@ mysql> SELECT * FROM t1;
...
 
 ### tidb_capture_plan_baselines <span class="version-mark">New in v4.0</span>
 
 - Scope: SESSION | GLOBAL
-- Default value: OFF
+- Default value: `OFF`
 - This variable is used to control whether to enable the [baseline capturing](/sql-plan-management.md#baseline-capturing) feature. This feature depends on the statement summary, so you need to enable the statement summary before you use baseline capturing.
 - After this feature is enabled, the historical SQL statements in the statement summary are traversed periodically, and bindings are automatically created for SQL statements that appear at least twice.
 
 ### tidb_check_mb4_value_in_utf8
```
dveeden, line 274: Maybe this should link to `tidb_skip_utf8_check` and/or explain how these are related?

system-variables.md
morgo: Thanks! I will do it in a separate PR. For this I'm aiming for parity through an auto-generator.

system-variables.md
morgo: Thanks! I will do it in a separate PR. For this I'm aiming for parity through an auto-generator.

system-variables.md
morgo: I think this is the right way to do it (human size in brackets). As you note the previous example is problematic because `SET GLOBAL` won't accept it. I will work on it.

system-variables.md
```diff
@@ -232,40 +245,43 @@ mysql> SELECT * FROM t1;
...
+- Default value: `10240`
+- Range: `[0, 9223372036854775807]`
 - The unit of the variable is rows. If the objects of the join operation belong to a subquery, the optimizer cannot estimate the size of the subquery result set. In this situation, the size is determined by the number of rows in the result set. If the estimated number of rows in the subquery is less than the value of this variable, the Broadcast Hash Join algorithm is used. Otherwise, the Shuffled Hash Join algorithm is used.
 
 ### tidb_broadcast_join_threshold_size <span class="version-mark">New in v5.0</span>
 
 - Scope: SESSION | GLOBAL
-- Default value: 104857600 (which equals 100 megabytes)
-- If the table size (in the unit of bytes) is less than the value of the variable, the Broadcast Hash Join algorithm is used. Otherwise, the Shuffled Hash Join algorithm is used.
+- Default value: `104857600`
```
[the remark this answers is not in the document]
morgo: I will use the value in brackets as suggested below.

system-variables.md
```diff
@@ -136,13 +143,14 @@ mysql> SELECT * FROM t1;
...
-- Default value: 4000
+- Default value: `4000`
+- Range: `[0, 65535]`
 - The port that the `tidb-server` is listening on when speaking the MySQL protocol.
 
 ### socket
 
 - Scope: NONE
-- Default value: ''
+- Default value: ""
```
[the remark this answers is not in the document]
morgo: Maybe in a future PR. I am just aiming for parity first.

system-variables.md
morgo: I'm going to take this up with the team responsible before fixing it. The range here is 292 years, which doesn't really make sense so Golang formats this in minutes with duration.String() method.

I think setting it in code to be a maximum of 1 year is the logical way to handle it.

morgo:
@TomShawn can we merge this? It will likely start to get merge conflicts.

TomShawn (approved):
LGTM

system-variables.md
```diff
@@ -1206,7 +1261,7 @@ This variable is an alias for _transaction_isolation_.
 ### version
 
 - Scope: NONE
-- Default value: 5.7.25-TiDB-(tidb version)
+- Default value: `5.7.25-TiDB-`(tidb version)
```
TomShawn, line 1264: \`5.7.25-TiDB-(tidb version)\`

PR #5761, opened 2021-06-09
system-variables: add datadir and license

morgo:

This adds 2 missing system variables that function correctly in TiDB: datadir and license.

The language is intentionally a little vague and says "typically" because in mocktikv/unistore cases it will show a path, but that is for developers to know. I don't believe we ever change the license string for enterprise edition but the language doesn't preclude that.

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/system-variables.md
+++ b/system-variables.md
@@ -92,7 +92,7 @@ mysql> SELECT * FROM t1;
 
 - Scope: NONE
 - Default value: /tmp/tidb
-- This variable indicates the location where data is stored. This can be a local path or point to a PD server in case the data is stored on TiKV.
+- This variable indicates the location where data is stored. This location can be a local path or point to a PD server if the data is stored on TiKV.
 - A value in the format of `ip_address:port` indicates the PD server that TiDB connects to on startup.
 
 ### ddl_slow_threshold
--- a/system-variables.md
+++ b/system-variables.md
@@ -86,7 +86,7 @@ mysql> SELECT * FROM t1;
 - Scope: NONE
 - Default value: /tmp/tidb
 - This variable indicates the location where data is stored. This can be a local path or point to a PD server in case the data is stored on TiKV.
-- Typically the value when be in the format of `ip_address:port`, indicating the PD server that TiDB connected to on startup.
+- A value in the format of `ip_address:port` indicates the PD server that TiDB connects to on startup.
 
 ### ddl_slow_threshold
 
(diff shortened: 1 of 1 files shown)


system-variables.md
```diff
@@ -81,6 +81,13 @@ mysql> SELECT * FROM t1;
 - Default value: ON
 - Controls whether statements should automatically commit when not in an explicit transaction. See [Transaction Overview](/transaction-overview.md#autocommit) for more information.
 
+### datadir
+
+- Scope: NONE
+- Default value: /tmp/tidb
+- This variable indicates the location where data is stored.
+- Typically the value when be in the format of `ip_address:port`, indicating the PD server that TiDB connected to on startup.
```
TomShawn, line 89:
```suggestion
- Typically, the value in the format of `ip_address:port` indicates the PD server that TiDB connects to on startup.
```

system-variables.md
```diff
@@ -81,6 +81,13 @@ mysql> SELECT * FROM t1;
 - Default value: ON
 - Controls whether statements should automatically commit when not in an explicit transaction. See [Transaction Overview](/transaction-overview.md#autocommit) for more information.
 
+### datadir
+
+- Scope: NONE
+- Default value: /tmp/tidb
+- This variable indicates the location where data is stored.
```
dveeden, line 88:
```suggestion
- This variable indicates the location where data is stored. This can be a local path or point to a PD server in case the data is stored on TiKV.
```

dveeden (approved):

system-variables.md
morgo: I made a small edit to your change. Combining with Daniel's feedback we can just say "A value in the format.." since above it clarifies it can be a local path.

xhebox (approved):

system-variables.md
```diff
@@ -88,6 +88,13 @@ mysql> SELECT * FROM t1;
 - Default value：1000
 - Controls the maximum recursion depth in Common Table Expressions.
 
+### datadir
+
+- Scope: NONE
+- Default value: /tmp/tidb
+- This variable indicates the location where data is stored. This can be a local path or point to a PD server in case the data is stored on TiKV.
```
TomShawn, line 95:
```suggestion
- This variable indicates the location where data is stored. This location can be a local path or point to a PD server if the data is stored on TiKV.
```

TomShawn (approved):
Rest LGTM

PR #5796, opened 2021-06-16
Add script for server side help

dveeden:

Add script to create help data for https://github.com/pingcap/tidb/pull/25485

This takes the statement documentation and creates insert statements from that for `mysql.help_topic`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,6 +112,26 @@ jobs:
       - store_artifacts:
           path: /tmp/artifacts
 
+  server_help:
+    docker:
+      - image: golang:1.16
+    working_directory: /go/src/github.com/pingcap/docs
+
+    steps:
+      - checkout
+
+      - run:
+          name: "Install dependencies"
+          command: apt-get update && apt-get install -y pandoc
+
+      - run:
+          name: "Generate Help Files"
+          command: scripts/server_help.sh
+
+      - run:
+          name: "Go tests"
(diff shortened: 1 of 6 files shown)


morgo (approved):

xhebox (approved):
I thought go embed need a real existed file. But looks like we did not generate any file yet.. Maybe add a example file for `help` command itself?

dveeden:
> I thought go embed need a real existed file. But looks like we did not generate any file yet.. Maybe add a example file for `help` statement itself?

Yes, that's a bit of an issue. The files `go embed` uses are created by `scripts/server_help.sh` but are not added as part of the commit as this is excluded in `.gitignore`. We could change that and include it, just like `parser.go` in pingcap/parser.

There is another option: embed the `sql-statements/sql-statement-*md` files and add some golang dependencies for the markdown to text conversion.

My choice would be the first option, were we use pandoc to convert from markdown to plain text.

xhebox:
> My choice would be the first option

I like how it is done now :)

It is just, seems we can not really test the result of this PR in TiDB... But I gave LGTM anyway. There is nothing wrong from my point of view.

.circleci/config.yml
```diff
@@ -112,6 +112,26 @@ jobs:
       - store_artifacts:
           path: /tmp/artifacts
 
+  server_help:
```
TomShawn, line 115: Do we really need to add this to CircleCI check that runs for every `docs` commit?

PR #5520, opened 2021-04-26
ticdc: Include local:// and s3:// sink support

dveeden:
### What is changed, added or deleted? (Required)

- Add documentation about the `local://` and `s3://` sinks in TiCDC
- Add documentation about `br restore cdclog`

### Which TiDB version(s) do your changes apply to? (Required)

- [x] master (the latest development version)
- [x] v5.0 (TiDB 5.0 versions)
- [ ] v4.0 (TiDB 4.0 versions)
- [ ] v3.1 (TiDB 3.1 versions)
- [ ] v3.0 (TiDB 3.0 versions)
- [ ] v2.1 (TiDB 2.1 versions)

### What is the related PR or file link(s)?

- This PR is translated from:
- Other reference link(s):

### Do your changes match any of the following descriptions?

- [ ] Delete files
- [ ] Change aliases
- [ ] Need modification after applied to another branch 
- [ ] Might cause conflicts after applied to another branch

PR #5879, opened 2021-06-29
Remove contents about PiTR feature

IANTHEREAL:

The PiTR function an experiment feature which developed in FY 2020 and there is no any follow-up test and work. It's not under any verification, such as from QA/PM, so I think it's not recommanded to customer, even in PoC. 

It should be better to remove now, and add it back when it's really ready.
This PR reverts https://github.com/pingcap/docs/pull/5520.

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/br/use-br-command-line-tool.md
+++ b/br/use-br-command-line-tool.md
@@ -244,28 +244,6 @@ LAST_BACKUP_TS=`br validate decode --field="end-version" -s local:///home/tidb/b
 
 In the above example, for the incremental backup data, BR records the data changes and the DDL operations during `(LAST_BACKUP_TS, current PD timestamp]`. When restoring data, BR first restores DDL operations and then the data.
 
-### Point-in-time recovery (experimental feature)
-
-Point-in-time recovery (PITR) allows you to restore data to a point in time of your choice.
-
-An example scenario would be to take a full backup every day and take incremental backups every 6 hours and then use TiCDC for PITR. Assume that on one day, the full backup was performed at 00:00 and the first incremental backup was performed at 06:00. If you want to restore the database to the state of 07:16, you can first restore the full backup (taken at 00:00) and the incremental backup (taken at 06:00), and then restore TiCDC logs that fill in the gap between 06:00 and 07:16.
-
-To peform the PITR, you can take the following steps:
-
-1. Restore a full backup using `br restore full`.
-2. (optional) Restore incremental backup(s).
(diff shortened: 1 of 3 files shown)


IANTHEREAL:
@Ishiihara @qiancai @3pointer  PTAL

TomShawn:
@

> @Ishiihara @qiancai @3pointer PTAL

@dveeden is the author of the doc content.

3pointer (approved):

dveeden (changes requested):
I think that adding a warning in the docs instead is better.

I don't think removing it completely is a good step. At least we should leave some info about this being an experimental feature. Maybe also add info about what people should use instead (TiDB Binlog?).

IANTHEREAL:
@dveeden Got you point. Actually I am quite confused about the standard of the experimental function. This feature is initially developed and has not undergone any testing, whether development or QA verification, so I think it is not suitable for release as an experimental feature. 

And for the PiTR feature, I think it is currently not supported now, and there is no suitable alternative, such as the tidb binlog you mentioned. I think we need to advance this feature as soon as possible

The above point of view is mainly based on the fact that we don’t want users to use a feature that is full of problems and may be changed.

dveeden:
@IANTHEREAL I do think people are using TiDB Binlog for this. There is no mention of this being experimental or deprecated on https://docs.pingcap.com/tidb/stable/tidb-binlog-reparo

For now adding a warning to TiCDC PITR similar to this https://docs.pingcap.com/tidb/stable/use-br-command-line-tool#back-up-raw-kv-experimental-feature would be good.

I agree that testing and removing the experimental flag of this feature would be good.

IANTHEREAL:
@dveeden Ok, Let me see

IANTHEREAL:
@dveeden Sorry for the late reply, I have been on a business trip in the last two days. I think we have a deviation from the standard definition of the experimental feature. After all, it has not been tested and is not recommended for any users. If you want to keep these contents, how about we put it in a file separately, in this file we can also describe the use of tidb binlog as an alternative solution

dveeden:
@IANTHEREAL I'm ok with that.

TomShawn:
@IANTHEREAL @dveeden Any update?

dveeden:
@TomShawn No updates from me, I'm waiting for @IANTHEREAL to update this PR

PR #6374, opened 2021-09-02
system-variables: improve noop functions warning

morgo:

The feature 'noop functions' is actually very unsafe. Several internal users I have spoken to recently have misunderstood it.

This changes it from a 'note' to a 'warning' and provides an example for more context of what it does.

The example will need to be changed slightly for TiDB 5.1 and below because transaction read only is not affected by noop funcs yet.

- [x] v5.2 (TiDB 5.2 versions)

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases
- [x] Might cause conflicts after applied to another branch

### DIFF
--- a/system-variables.md
+++ b/system-variables.md
@@ -596,9 +596,9 @@ Constraint checking is always performed in place for pessimistic transactions (d
     * `START TRANSACTION READ ONLY` and `SET TRANSACTION READ ONLY` syntax
     * The `tx_read_only`, `transaction_read_only`, `offline_mode`, `super_read_only` and `read_only` system variables
 
-> **Note:**
+> **Warning:**
 >
-> Only the default value of `OFF` can be considered safe. Setting `tidb_enable_noop_functions=1` might lead to unexpected behaviors in your application, because it permits TiDB to ignore certain syntax without providing an error.
+> Only the default value of `OFF` can be considered safe. Setting `tidb_enable_noop_functions=1` might lead to unexpected behaviors in your application, because it permits TiDB to ignore certain syntax without providing an error. For example, the syntax `START TRANSACTION READ ONLY` will be permitted, but transactions will remain in read-write mode.
 
 ### tidb_enable_parallel_apply <span class="version-mark">New in v5.0</span>
 
--- a/system-variables.md
+++ b/system-variables.md
@@ -598,7 +598,7 @@ Constraint checking is always performed in place for pessimistic transactions (d
 
 > **Warning:**
 >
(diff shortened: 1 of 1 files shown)


dveeden:
I don't think a read only transaction is the best example. This because a read only transaction for me is mostly a performance optimization allowing the server to skip these when scanning transactions in some cases. (see also https://dev.mysql.com/doc/refman/8.0/en/innodb-performance-ro-txn.html )

Maybe this would be a good example?
```
mysql> set global read_only=on;
ERROR 1235 (42000): function READ ONLY has only noop implementation in tidb now, use tidb_enable_noop_functions to enable these functions
mysql> set global tidb_enable_noop_functions=on;

mysql> set global read_only=on;
```

side note:
I expect setting this as a session variable to work, but it didn't:
```
mysql> set global read_only=on;
ERROR 1235 (42000): function READ ONLY has only noop implementation in tidb now, use tidb_enable_noop_functions to enable these functions
mysql> set session tidb_enable_noop_functions=on;

mysql> set global read_only=on;
ERROR 1235 (42000): function READ ONLY has only noop implementation in tidb now, use tidb_enable_noop_functions to enable these functions
```

dveeden (approved):

system-variables.md
```diff
@@ -596,9 +596,9 @@ Constraint checking is always performed in place for pessimistic transactions (d
     * `START TRANSACTION READ ONLY` and `SET TRANSACTION READ ONLY` syntax
     * The `tx_read_only`, `transaction_read_only`, `offline_mode`, `super_read_only` and `read_only` system variables
 
-> **Note:**
+> **Warning:**
 >
-> Only the default value of `OFF` can be considered safe. Setting `tidb_enable_noop_functions=1` might lead to unexpected behaviors in your application, because it permits TiDB to ignore certain syntax without providing an error.
+> Only the default value of `OFF` can be considered safe. Setting `tidb_enable_noop_functions=1` might lead to unexpected behaviors in your application, because it permits TiDB to ignore certain syntax without providing an error. For example, the syntax `START TRANSACTION READ ONLY` will be permitted, but transactions will remain in read-write mode.
```
TomShawn, line 601:
```suggestion
> Only the default value of `OFF` can be considered safe. Setting `tidb_enable_noop_functions=1` might lead to unexpected behaviors in your application, because it permits TiDB to ignore certain syntax without providing an error. For example, the syntax `START TRANSACTION READ ONLY` is permitted, but the transaction remains in read-write mode.
```

TomShawn (approved):

morgo:
> I don't think a read only transaction is the best example. This because a read only transaction for me is mostly a performance optimization allowing the server to skip these when scanning transactions in some cases. (see also https://dev.mysql.com/doc/refman/8.0/en/innodb-performance-ro-txn.html )

I'm not sure the MySQL history is aparent to users. The reason I went with this feature is it was "feature syntax" versus a variable. When I was discussing with users who misunderstood this, they seemed to think the feature had some magic that was compatibility mode aware, which is not the case at all.

PR #7413, opened 2022-01-15
hardware-and-software-requirements: explicitly exclude Ubuntu 21.10

dveeden:

Explictly exclude Ubuntu 21.10 from the list of supported operating systems as current TiFlash versions can't work with new glibc versions.

- [x] v5.4 (TiDB 5.4 versions)
- [x] v5.3 (TiDB 5.3 versions)
- [x] v5.2 (TiDB 5.2 versions)


dveeden:
Note that the latest Ubuntu LTS release is 20.04 and the latest non-LTS release is 21.10. See https://releases.ubuntu.com/
Maybe we should rephrase this to be less confusing.
Note that Ubuntu 22.04 is going to be an LTS release (source: https://ubuntu.com/about/release-cycle ), but would probably also need to be excluded from the list of supported versions.

morgo:
@dveeden Is there a known TiFlash bug we can point to so users will know when this is resolved? (or documenters; since we rarely track supported platforms diligently).

I am an Ubuntu 21.10 user, I was wondering why TiFlash was broken.

dveeden:
> @dveeden Is there a known _public_ TiFlash bug we can point to so users will know when this is resolved?

I'm not aware of any public bugs that cover this.

morgo:
> I'm not aware of any public bugs that cover this.

Can we link to this PR/issue since it has a little more context.

SchrodingerZhu (approved):
LGTM. According to previous tests by @dveeden, it is true that TiFlash has compatibility issues with newer glibcs (for all released stable versions; but not those nightly ones).

dveeden:
With v6.x releases of TiFlash it is no longer needed to exclude Ubuntu 21.10

PR #7880, opened 2022-03-17
Add cached tables documentation

TomShawn:
Add cached tables documentation

- [x] v6.0 (TiDB 6.0 versions)

- This PR is translated from: https://github.com/pingcap/docs-cn/pull/8557
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases


table-cache.md
```diff
@@ -0,0 +1,254 @@
...
+
+## Size limit of table cache
+
+The table cache is only suitable to scenarios with small tables, because TiDB loads the data of an entire table into memory, and the cached data becomes invalid after modification and needs to be reloaded.
+
+Currently, the size limit of the table cache is 64 MB in TiDB. If the table data exceeds 64 MB, executing `ALTER TABLE t CACHE` will fail.
+
+## Compatibility restrictions with other TiDB features
+
+以下是缓存表不支持的功能：
```
tiancaiamao, line 230: Please remove this line

sql-statements/sql-statement-alter-table.md
```diff
@@ -42,6 +42,8 @@ AlterTableSpec ::=
 |   ( 'WITH' | 'WITHOUT' ) 'VALIDATION'
 |   'SECONDARY_LOAD'
 |   'SECONDARY_UNLOAD'
+|   ( 'AUTO_INCREMENT' | 'AUTO_ID_CACHE' | 'AUTO_RANDOM_BASE' | 'SHARD_ROW_ID_BITS' ) EqOpt LengthNum
```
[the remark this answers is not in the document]
TomShawn: Yes. To keep it consistent with the Chinese version: https://github.com/pingcap/docs-cn/blob/master/sql-statements/sql-statement-alter-table.md

table-cache.md
TomShawn: Done

tiancaiamao (approved):

morgo:
I mentioned to @tiancaiamao today - I didn't realize from the [design doc](https://github.com/pingcap/tidb/pull/23673) that we were going to call this 'table cache'.

Unfortunately this name exists in another feature in MySQL, which is the setting for how many copies of table meta data can be in memory (see [Google searches for tuning this](https://www.google.com/search?q=mysql+table+cache&oq=mysql+t&aqs=chrome.0.69i59j69i57j69i59j0i512l2j69i60l3.2438j0j4&sourceid=chrome&ie=UTF-8)). The setting is historical now, but well known by MySQL DBAs.

If we could instead rename it to 'cached tables' or 'read-optimized tables' it would be clearer.

TomShawn:
> I mentioned to @tiancaiamao today - I didn't realize from the [design doc](https://github.com/pingcap/tidb/pull/23673) that we were going to call this 'table cache'.
> 
> Unfortunately this name exists in another feature in MySQL, which is the setting for how many copies of table meta data can be in memory (see [Google searches for tuning this](https://www.google.com/search?q=mysql+table+cache&oq=mysql+t&aqs=chrome.0.69i59j69i57j69i59j0i512l2j69i60l3.2438j0j4&sourceid=chrome&ie=UTF-8)). The setting is historical now, but well known by MySQL DBAs.
> 
> If we could instead rename it to 'cached tables' or 'read-optimized tables' it would be clearer.

Thanks for pointing it out! 
@tiancaiamao What do you think? Shall we change the feature name in English?

dveeden:
> I mentioned to @tiancaiamao today - I didn't realize from the [design doc](https://github.com/pingcap/tidb/pull/23673) that we were going to call this 'table cache'.
> 
> Unfortunately this name exists in another feature in MySQL, which is the setting for how many copies of table meta data can be in memory (see [Google searches for tuning this](https://www.google.com/search?q=mysql+table+cache&oq=mysql+t&aqs=chrome.0.69i59j69i57j69i59j0i512l2j69i60l3.2438j0j4&sourceid=chrome&ie=UTF-8)). The setting is historical now, but well known by MySQL DBAs.
> 
> If we could instead rename it to 'cached tables' or 'read-optimized tables' it would be clearer.

I fully agree that this might lead to problems. However if we rename the feature, but still use the `CACHE` keyword in SQL statements this might get confusing.

We *could* name this something like "In Memory Tables" as that seems to fit. However with the (somewhat) recent hype around [In-memory databases](https://en.wikipedia.org/wiki/In-memory_database) that might not be what we want. And this could also lead to confusion with the [`MEMORY`/`HEAP` storage engine in MySQL](https://dev.mysql.com/doc/refman/8.0/en/memory-storage-engine.html).

tiancaiamao:
> However if we rename the feature, but still use the `CACHE` keyword in SQL statements this might get confusing.

MySQL don't support the `alter table cache` syntax, so the SQL statement will not conflict with it.

> If we could instead rename it to 'cached tables' or 'read-optimized tables' it would be clearer.

"In Memory Tables" may lead to confusion too, as @dveeden mentioned.
I would vote for 'cached tables'...
Because if we change to 'read-optimized tables', the SQL statement looks inconsistent
Then we might need to reconsider the syntax like 'alter table xxx for read' or 'alter table xxx read optimized'.
Considering the code is frozen and 6.0 is about to be released, those changes is not a big deal.

TomShawn:
@morgo @dveeden @tiancaiamao The feature is renamed in this PR as "cached tables". PTAL, thanks~

dveeden (approved):

sql-statements/sql-statement-flush-tables.md
```diff
@@ -51,7 +51,7 @@ ERROR 1105 (HY000): FLUSH TABLES WITH READ LOCK is not supported.  Please use @@
 
 ## MySQL compatibility
 
-* TiDB does not have a concept of table cache as in MySQL. Thus, `FLUSH TABLES` is parsed but ignored in TiDB for compatibility.
+* TiDB does not have a concept of cached tables as in MySQL. Thus, `FLUSH TABLES` is parsed but ignored in TiDB for compatibility.
```
morgo, line 54: This edit is incorrect. It refers to the MySQL table cache feature correctly (not to be confused with cached tables).

cached-tables.md
```diff
@@ -0,0 +1,252 @@
...
+
+```sql
+mysql> ALTER TABLE users ADD INDEX k_id(id);
+```
+
+```sql
+ERROR 8242 (HY000): 'Alter Table' is unsupported on cache tables.
+```
+
+To recover a cached table to a normal table, use `ALTER TABLE t NOCACHE`:
```
TomShawn, line 210:
```suggestion
To revert a cached table to a normal table, use `ALTER TABLE t NOCACHE`:
```

cached-tables.md
```diff
@@ -0,0 +1,252 @@
...
+Cached tables **CANNOT** be used in the following scenarios:
+
+- Setting the system variable `tidb_snapshot` to read historical data.
+- During modification, the cached data becomes invalid until the data is reloaded.
+
+## Compatibility with TiDB ecosystem tools
+
+The cached table is not a standard MySQL feature but a TiDB extension. Only TiDB can recognize the `ALTER TABLE CACHE` statement. TiDB ecosystem tools **DOES NOT** support cached tables, including Backup & Restore (BR), TiCDC, and Dumpling. They treat cached tables as a normal table.
+
+That is to say, when a cached table is backed up and restored, it becomes a normal table. If the downstream cluster is a different TiDB cluster and you want to continue using the cached table feature, you can manually enable cached tables on the downstream cluster by executing `ALTER TABLE CACHE` on the downstream table.
```
TomShawn, line 247:
```suggestion
That is to say, when a cached table is backed up and restored, it becomes a normal table. If the downstream cluster is a different TiDB cluster and you want to continue using the cached table feature, you can manually enable cached tables on the downstream cluster by executing `ALTER TABLE ... CACHE` on the downstream table.
```

sql-statements/sql-statement-flush-tables.md
TomShawn:
```suggestion
* TiDB does not have a concept of table cache as in MySQL. Thus, `FLUSH TABLES` is parsed but ignored in TiDB for compatibility.
```

sql-statements/sql-statement-alter-table.md
```diff
@@ -162,6 +164,8 @@ The following major restrictions apply to `ALTER TABLE` in TiDB:
 
 * Spatial data types are not supported.
 
+* `ALTER TABLE t CACHE | NOCACHE` is not a standard MySQL syntax but a TiDB extension. See [Cached Tables](/cached-tables.md) for more information.
```
morgo, line 167: SQL is a standard, but for MySQL it is only a reference implementation. I think we can improve the readability by just calling it an extension to MySQL syntax:
```suggestion
* `ALTER TABLE t CACHE | NOCACHE` is a TiDB extension to MySQL syntax. See [Cached Tables](/cached-tables.md) for more information.
```

cached-tables.md
```diff
@@ -0,0 +1,252 @@
...
+
+```sql
+Query OK, 0 rows affected (0.00 sec)
+```
+
+## Size limit of cached tables
+
+Cached tables are only suitable to scenarios with small tables, because TiDB loads the data of an entire table into memory, and the cached data becomes invalid after modification and needs to be reloaded.
+
+Currently, the size limit of a cached table is 64 MB in TiDB. If the table data exceeds 64 MB, executing `ALTER TABLE t CACHE` will fail.
```
en-jin19, line 226:
```suggestion
Currently, the size limit of a cached table is 64 MB in TiDB. If the table data is large than 64 MB, `ALTER TABLE t CACHE` cannot be executed.
```

cached-tables.md
```diff
@@ -0,0 +1,252 @@
+---
+title: Cached Tables
+summary: Learn the cached table feature in TiDB, which is used for rarely-updated small hotspot tables to improve read performance.
+---
+
+# Cached Tables
+
+In v6.0.0, TiDB introduces the cached table feature for frequently accessed but rarely updated small hotspot tables. When this feature is used, the data of an entire table is loaded into the memory of the TiDB server, and TiDB directly gets the table data from the memory without accessing TiKV, which improves the read performance.
+
+This document describes the usage scenarios of cached tables, the examples, and the compatibility restrictions with other TiDB features.
```
en-jin19, line 10:
```suggestion
This document describes the usage scenarios, the examples, and the compatibility restrictions with other TiDB features of cached tables.
```

cached-tables.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Usage scenarios
+
+The cached table feature is suitable for tables with the following characteristics:
+
+- The data volume of the table is small.
+- The table is read-only or rarely updated.
+- The table is frequently accessed, and you expect a better read performance.
+
+When the data volume of the table is small but the data is frequently accessed, the data is concentrated on a Region in TiKV and makes it a hotspot Region, which affects the performance. Therefore, the typical usage scenarios of cached tables are as follows:
```
en-jin19, line 20:
```suggestion
When the data volume of the table is small but the data is frequently accessed, the data is concentrated on a Region in TiKV and makes it a hotspot Region, which affects the performance of TiKV. Therefore, the typical usage scenarios of cached tables are as follows:
```

cached-tables.md
```diff
@@ -0,0 +1,252 @@
...
+
+- Configuration tables, from which applications read the configuration information.
+- The tables of exchange rates in the financial sector. These tables are updated only once a day but not in real-time.
+- Bank branch or network information tables, which are rarely updated.
+
+Taking configuration tables as an example, when the application restarts, the configuration information is loaded in all connections, which causes a high read latency. You can solve this problem by using the cached tables feature.
```
en-jin19, line 26:
```suggestion
Taking configuration tables as an example, when the application restarts, the configuration information is loaded in all connections, which causes a high read latency. In this case, you can solve this problem by using the cached tables feature.
```

cached-tables.md
```diff
@@ -0,0 +1,252 @@
...
+> set @@global.tidb_table_cache_lease = 10;
+> ```
+>
+> The write latency of cached tables is high, because the cached table feature is implemented with a complex mechanism that requires a lease to be set for each cache. When there are multiple TiDB instances, one instance does not know whether the other instances have cached data. If an instance modifies the table data directly, the other instances read the old cache data. To ensure correctness, the cached table implementation uses a lease mechanism to ensure that the data is not modified before the lease expires. That is why the write latency is high.
+
+### Revert a cached table to a normal table
+
+> **Note:**
+>
+> Executing DDL statements on a cached table will fail. Before executing DDL statements on a cached table, you need to remove the cache attribute first and set the cached table back to a normal table.
```
en-jin19, line 188:
```suggestion
> You cannot execute DDL statements on a cached table. Before executing DDL statements on a cached table, you need to remove the cache attribute first and set the cached table back to a normal table.
```

en-jin19 (approved):
rest LGTM

cached-tables.md
TomShawn: Can't agree with the edit.

cached-tables.md
TomShawn: Not sure the performance of TiKV or of multiple components as a whole.

tiancaiamao (approved):

PR #8440, opened 2022-04-29
add docs for GET_LOCK

morgo:

This updates the docs for GET_LOCK, RELEASE_LOCK, RELEASE_ALL_LOCKS. A feature I worked on for TiDB 6.1.

- [ ] v6.0 (TiDB 6.0 versions)

https://github.com/pingcap/tidb/pull/33947

- [ ] Delete files
- [ ] Change aliases


functions-and-operators/locking-functions.md
```diff
@@ -0,0 +1,26 @@
+---
+title: Locking Functions
+summary: Learn about user-level locking functions in TiDB.
+---
+
+# Locking Functions
+
+TiDB supports most of the user-level [locking functions](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html) available in MySQL 5.7.
```
dveeden, line 8:
```suggestion
TiDB supports locking functions to allow users to explicitly control what things are allowed to execute at the same time.
```

I think this should be a general description of the functionality, especially for users that may not have used MySQL before.

functions-and-operators/locking-functions.md
```diff
@@ -0,0 +1,26 @@
...
+
+| Name                                                                                                                 | Description                                       |
+|:---------------------------------------------------------------------------------------------------------------------|:--------------------------------------------------|
+| [`GET_LOCK(lockName, timeout)`](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_get-lock)    | Acquire an advisory lock                          |
+| [`RELEASE_LOCK(lockName)`](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_release-lock)     | Release a previously acquired lock                |
+| [`RELEASE_ALL_LOCKS()`](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_release-all-locks)   | Release all locks held by the current session     |
+
+## MySQL compatibility
+
+* The maximum timeout permitted by TiDB is 1 hour. This differs from MySQL, where the timeout is unlimited.
```
dveeden, line 20:
```suggestion
* TiDB supports most of the user-level [locking functions](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html) available in MySQL 5.7.
* The maximum timeout permitted by TiDB is 1 hour. This differs from MySQL, where the timeout is unlimited.
```

functions-and-operators/locking-functions.md
```diff
@@ -0,0 +1,26 @@
...
+* TiDB does not automatically detect deadlocks in user-level locks. Deadlocked sessions will timeout after a maximum of 1 hour, but can also be manually resolved by using `KILL` on one of the affected sessions. Deadlocks can also be prevented by always acquiring user-locks in the same order.
```
dveeden, line 21:
```suggestion
* TiDB does not automatically detect deadlocks in user-level locks. Deadlocked sessions will timeout after a maximum of 1 hour, but can also be manually resolved by using `KILL` on one of the affected sessions. Deadlocks can also be prevented by always acquiring user-locks in the same order.
* Locks are visible on all the TiDB nodes in the cluster.
```

With MySQL Cluster (NDB) or InnoDB Cluster (aka Group Replication) the locks are local to the node (which I think is not desirable). But this is an important difference for users.

dveeden (approved):

functions-and-operators/locking-functions.md
morgo: What do you think about saying something to "locks are scoped to all TiDB nodes in the cluster", instead of locks are visible? Visibility is ambiguous.

functions-and-operators/locking-functions.md
morgo: I copied the format from other function-reference pages, which links to the MySQL docs here. For example: https://docs.pingcap.com/tidb/stable/string-functions and https://docs.pingcap.com/tidb/stable/numeric-functions-and-operators

I think your suggestion is an improvement, but in the interest of consistency I think we should keep it as it is.

functions-and-operators/locking-functions.md
```diff
@@ -0,0 +1,26 @@
...
+
+# Locking Functions
+
+TiDB supports most of the user-level [locking functions](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html) available in MySQL 5.7.
+
+## Supported functions
+
+| Name                                                                                                                 | Description                                       |
+|:---------------------------------------------------------------------------------------------------------------------|:--------------------------------------------------|
+| [`GET_LOCK(lockName, timeout)`](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_get-lock)    | Acquire an advisory lock                          |
```
seiya-annie, line 14: should we add the restriction information of the lock name in this doc?
like: the length of lock name should <= 64 characters

morgo: I don't think it's necessary. 64 chars is the same limit in mysql for lock name, and the same limit for column / table names.

dveeden: I think we should document this <=64 chars requirement for people that don't have a MySQL background. Or do we expect them to check the docs on mysql.com?

morgo:
@dveeden @seiya-annie PTAL again. Thanks!

dveeden (approved):

functions-and-operators/locking-functions.md
```diff
@@ -0,0 +1,27 @@
...
+
+# Locking Functions
+
+TiDB supports most of the user-level [locking functions](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html) available in MySQL 5.7.
+
+## Supported functions
+
+| Name                                                                                                                 | Description                                                           |
+|:---------------------------------------------------------------------------------------------------------------------|:----------------------------------------------------------------------|
+| [`GET_LOCK(lockName, timeout)`](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_get-lock)    | Acquire an advisory lock. LockName must be <= 64 characters.          |
```
TomShawn, line 14:
```suggestion
| [`GET_LOCK(lockName, timeout)`](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_get-lock)    | Acquires an advisory lock. The `lockName` parameter must be NO longer than 64 characters.          |
```

functions-and-operators/locking-functions.md
```diff
@@ -0,0 +1,27 @@
...
+| [`RELEASE_LOCK(lockName)`](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_release-lock)     | Release a previously acquired lock LockName must be <= 64 characters. |
```
TomShawn, line 15:
```suggestion
| [`RELEASE_LOCK(lockName)`](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_release-lock)     | Releases a previously acquired lock. The `lockName` parameter must be NO longer than 64 characters. |
```

functions-and-operators/locking-functions.md
```diff
@@ -0,0 +1,27 @@
...
+| [`RELEASE_ALL_LOCKS()`](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_release-all-locks)   | Release all locks held by the current session.                        |
```
TomShawn, line 16:
```suggestion
| [`RELEASE_ALL_LOCKS()`](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_release-all-locks)   | Releases all locks held by the current session.                        |
```

functions-and-operators/locking-functions.md
```diff
@@ -0,0 +1,27 @@
...
+
+## MySQL compatibility
+
+* The minimum timeout permitted by TiDB is 1 second, and the maximum timeout is 1 hour. This differs from MySQL, where both zero second and unlimited timeouts are permitted. TiDB will automatically convert out of range values to the nearest permitted value.
```
TomShawn, line 20:
```suggestion
* The minimum timeout permitted by TiDB is 1 second, and the maximum timeout is 1 hour. This differs from MySQL, where both 0 second and unlimited timeouts are permitted. TiDB will automatically convert out-of-range values to the nearest permitted value.
```

functions-and-operators/locking-functions.md
```diff
@@ -0,0 +1,27 @@
...
+* TiDB does not automatically detect deadlocks in user-level locks. Deadlocked sessions will timeout after a maximum of 1 hour, but can also be manually resolved by using `KILL` on one of the affected sessions. Deadlocks can also be prevented by always acquiring user-locks in the same order.
```
TomShawn, line 21: I am confused about "by using `KILL` on one of the affected sessions". Does it mean users need to use `KILL` only on one affected session out of multiple affected sessions? 
```suggestion
* TiDB does not automatically detect deadlocks caused by user-level locks. Deadlocked sessions will timeout after a maximum of 1 hour, but can also be manually resolved by using `KILL` on one of the affected sessions. You can also prevent deadlocks by constantly acquiring user-level locks in the same order.
```

functions-and-operators/locking-functions.md
```diff
@@ -0,0 +1,27 @@
...
+* Locks are scoped to all TiDB servers in the cluster. This differs from MySQL Cluster and Group Replication where locks are local to a single server.
```
TomShawn, line 22:
```suggestion
* Locks take effect on all TiDB servers in the cluster. This differs from MySQL Cluster and Group Replication where locks are local to a single server.
```

TomShawn (commented):
mostly minor comments

functions-and-operators/locking-functions.md
```diff
@@ -0,0 +1,27 @@
...
+
+# Locking Functions
+
+TiDB supports most of the user-level [locking functions](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html) available in MySQL 5.7.
+
+## Supported functions
+
+| Name                                                                                                                 | Description                                                           |
+|:---------------------------------------------------------------------------------------------------------------------|:----------------------------------------------------------------------|
+| [`GET_LOCK(lockName, timeout)`](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_get-lock)    | Acquires an advisory lock. The `lockName` parameter must be NO longer than 64 characters.          |
```
mjonss, line 14: Is characters meant to mean *bytes* here or any utf character? I just tested the master branch and a name "ö" cannot be more than 32 characters (2 byte character in utf8mb4).

mjonss, line 14:
```suggestion
| [`GET_LOCK(lockName, timeout)`](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_get-lock)    | Acquires an advisory lock. The `lockName` parameter must be NO longer than 64 characters. Waits maximum `timeout` seconds before timing out and returns a failure.         |
```

functions-and-operators/locking-functions.md
```diff
@@ -0,0 +1,27 @@
...
+| [`RELEASE_LOCK(lockName)`](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_release-lock)     | Releases a previously acquired lock. The `lockName` parameter must be NO longer than 64 characters. |
+| [`RELEASE_ALL_LOCKS()`](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_release-all-locks)   | Releases all locks held by the current session.                        |
+
+## MySQL compatibility
+
+* The minimum timeout permitted by TiDB is 1 second, and the maximum timeout is 1 hour. This differs from MySQL, where both 0 second and unlimited timeouts are permitted. TiDB will automatically convert out-of-range values to the nearest permitted value.
```
mjonss, line 20:
```suggestion
* The minimum timeout permitted by TiDB is 1 second, and the maximum timeout is 1 hour (3600 seconds). This differs from MySQL, where both 0 second and unlimited timeouts are permitted. TiDB will automatically convert out-of-range values to the nearest permitted value.
```

functions-and-operators/locking-functions.md
mjonss:
```
tidb> select get_lock(repeat("ä",33),1);
ERROR 3057 (HY000): Incorrect user-level lock name 'äääääääääääääääääääääääääääääääää'.
```

morgo: No, I think this is a bug. The check for length is checking based on byte length, but in MySQL the max is expected to be measured in char-length:

```sql
mysql [localhost:8028] {msandbox} (test) > select get_lock(repeat("ä",33),1);
+-----------------------------+
| get_lock(repeat("ä",33),1)  |
+-----------------------------+
|                           1 |
+-----------------------------+
1 row in set (0.00 sec)

mysql [localhost:8028] {msandbox} (test) > select get_lock(repeat("ä",65),1);
ERROR 3057 (42000): Incorrect user-level lock name 'äääääääääääääääääääääääääääääääääääääääääääääääääääääääääääääääää'.
```

morgo: Reported as https://github.com/pingcap/tidb/issues/34639

functions-and-operators/locking-functions.md
TomShawn: changed back to "always"

TomShawn (approved):
LGTM

PR #9481, opened 2022-07-04
We should never recommend slow storage for TiKV

dveeden:

Change the placement policy example where HDD and SSD is used to use SSD and NVMe as running TiKV is not a recommended setup.

- [ ] v6.2 (TiDB 6.2 versions)
- [x] v5.4 (TiDB 5.4 versions)

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/placement-rules-in-sql.md
+++ b/placement-rules-in-sql.md
@@ -222,30 +222,30 @@ The placement options `PRIMARY_REGION`, `REGIONS`, and `SCHEDULE` meet the basic
 For example, to set constraints that data must reside on a TiKV store where the label `disk` must match a value:
 
 ```sql
-CREATE PLACEMENT POLICY storeonfastssd CONSTRAINTS="[+disk=ssd]";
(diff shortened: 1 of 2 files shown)


TomShawn (approved):
LGTM

placement-rules-in-sql.md
```diff
@@ -15,7 +15,7 @@ The detailed user scenarios are as follows:
 
 - Merge multiple databases of different applications to reduce the cost on database maintenance
 - Increase replica count for important data to improve the application availability and data reliability
-- Store new data into SSDs and store old data into HHDs to lower the cost on data archiving and storage
+- Store new data into NVMe storage and store old data into SSDs to lower the cost on data archiving and storage
```
TomShawn, line 18:
```suggestion
- Store new data into NVMe storage and store old data into SSDs to lower the cost on data archiving and storage 
```

PR #10943, opened 2022-10-20
json: add document for json_storage_free

YangKeao:
Signed-off-by: YangKeao <yangkeao@chunibyo.icu>

Add a new function `json_storage_free`.

- This PR is translated from: 
- Other reference link(s): https://github.com/pingcap/tidb/issues/38517

### DIFF
--- a/functions-and-operators/json-functions.md
+++ b/functions-and-operators/json-functions.md
@@ -60,7 +60,7 @@ TiDB supports most of the JSON functions that shipped with the GA release of MyS
 | Function Name                     | Description |
 | --------------------------------- | ----------- |
 | [JSON_PRETTY(json_doc)][json_pretty] | Pretty formatting of a JSON document |
-| [JSON_STORAGE_FREE(json_doc)][json_storage_free] | Returns how much storage space was freed in the binary representation of the JSON value after it was updated in place. As TiDB has different storage architecture from MySQL, this function always returns 0 for a valid JSON value, and it's implemented for compatibility with MySQL 8.0. |
+| [JSON_STORAGE_FREE(json_doc)][json_storage_free] | Returns how much storage space was freed in the binary representation of the JSON value after it was updated in place. As TiDB has different storage architecture from MySQL, this function always returns 0 for a valid JSON value, and it is implemented for compatibility with MySQL 8.0. |
 | [JSON_STORAGE_SIZE(json_doc)][json_storage_size] | Returns an approximate size of bytes required to store the json value. As the size does not account for TiKV using compression, the output of this function is not strictly compatible with MySQL. |
 
 ## Aggregate Functions
--- a/functions-and-operators/json-functions.md
+++ b/functions-and-operators/json-functions.md
@@ -60,6 +60,7 @@ TiDB supports most of the JSON functions that shipped with the GA release of MyS
 | Function Name                     | Description |
 | --------------------------------- | ----------- |
 | [JSON_PRETTY(json_doc)][json_pretty] | Pretty formatting of a JSON document |
+| [JSON_STORAGE_FREE(json_doc)][json_storage_free] | Returns how much storage space was freed in the binary representation of the JSON value after it was updated in place. As TiDB has different storage architecture from MySQL, this function always returns 0 for a valid JSON value, and it's implemented for compatibility with MySQL 8.0. |
 | [JSON_STORAGE_SIZE(json_doc)][json_storage_size] | Returns an approximate size of bytes required to store the json value. As the size does not account for TiKV using compression, the output of this function is not strictly compatible with MySQL. |
 
 ## Aggregate Functions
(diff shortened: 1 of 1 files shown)


functions-and-operators/json-functions.md
```diff
@@ -132,4 +133,6 @@ TiDB supports most of the JSON functions that shipped with the GA release of MyS
 
 [json_pretty]: https://dev.mysql.com/doc/refman/5.7/en/json-utility-functions.html#function_json-pretty
 
+[json_storage_free]: https://dev.mysql.com/doc/refman/8.0/en/json-utility-functions.html#function_json-storage-free
```
dveeden, line 136: I think this link is fine as this function was introduced in MySQL 8.0 and isn't in the MySQL 5.7 docs. Not sure if we should explicitly list this as MySQL 8.0 compatibility or not. Also might be good to consider updating all links to MySQL 8.0 at some point, but that's not for this PR

dveeden (approved):

functions-and-operators/json-functions.md
```diff
@@ -60,6 +60,7 @@ TiDB supports most of the JSON functions that shipped with the GA release of MyS
 | Function Name                     | Description |
 | --------------------------------- | ----------- |
 | [JSON_PRETTY(json_doc)][json_pretty] | Pretty formatting of a JSON document |
+| [JSON_STORAGE_FREE(json_doc)][json_storage_free] | Returns how much storage space was freed in the binary representation of the json value after it was updated in place. As TiDB doesn't support update json value in place, the output of this function is always 0 for valid json value. |
```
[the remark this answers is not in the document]
YangKeao: Fixed.

functions-and-operators/json-functions.md
YangKeao: > Not sure if we should explicitly list this as MySQL 8.0 compatibility or not.

@qiancai PTAL. I think it would be better to tag as MySQL 8.0 feature, but I don't know whether other docs have similar problem / solution? :thinking: .

> Also might be good to consider updating all links to MySQL 8.0 at some point, but that's not for this PR

As I know, the support for MySQL 8.0 JSON features are still ongoing, and some references can be updated while some are not (e.g. I am working on supporting 8.0 json path syntax this month).

functions-and-operators/json-functions.md
```diff
@@ -60,6 +60,7 @@ TiDB supports most of the JSON functions that shipped with the GA release of MyS
 | Function Name                     | Description |
 | --------------------------------- | ----------- |
 | [JSON_PRETTY(json_doc)][json_pretty] | Pretty formatting of a JSON document |
+| [JSON_STORAGE_FREE(json_doc)][json_storage_free] | Returns how much storage space was freed in the binary representation of the JSON value after it was updated in place. As TiDB has different storage architecture with MySQL, this function always returns 0 for a valid JSON value, and it's implemented for compatibility. |
```
qiancai, line 63:
```suggestion
| [JSON_STORAGE_FREE(json_doc)][json_storage_free] | Returns how much storage space was freed in the binary representation of the JSON value after it was updated in place. As TiDB has different storage architecture from MySQL, this function always returns 0 for a valid JSON value, and it's implemented for compatibility with MySQL 8.0. |
```

functions-and-operators/json-functions.md
qiancai: Because L9 in this document mentions that "TiDB supports most of the JSON functions that shipped with the GA release of MySQL 5.7.", users would assume that the functions mentioned in this doc are for v5.7 by default. In this case, do you think we can highlight v8.0 as follows?
image

functions-and-operators/json-functions.md
YangKeao: Fixed.

functions-and-operators/json-functions.md
```diff
@@ -60,6 +60,7 @@ TiDB supports most of the JSON functions that shipped with the GA release of MyS
 | Function Name                     | Description |
 | --------------------------------- | ----------- |
 | [JSON_PRETTY(json_doc)][json_pretty] | Pretty formatting of a JSON document |
+| [JSON_STORAGE_FREE(json_doc)][json_storage_free] | Returns how much storage space was freed in the binary representation of the JSON value after it was updated in place. As TiDB has different storage architecture from MySQL, this function always returns 0 for a valid JSON value, and it's implemented for compatibility with MySQL 8.0. |
```
qiancai, line 63:
```suggestion
| [JSON_STORAGE_FREE(json_doc)][json_storage_free] | Returns how much storage space was freed in the binary representation of the JSON value after it was updated in place. As TiDB has different storage architecture from MySQL, this function always returns 0 for a valid JSON value, and it is implemented for compatibility with MySQL 8.0. |
```

qiancai (approved):
LGTM

Oreoxmt (approved):
LGTM

PR #11077, opened 2022-10-28
Fix an error in importing sample data

### LINKED ISSUE #6647: Import Example Database does not work with 'tiup client'
## Change Request

1. Describe what you find is inappropriate or missing in the existing docs.

This error was reported by @alkaagr81

On https://docs.pingcap.com/tidb/stable/import-example-data

The step to `LOAD DATA INFILE` does not work with the client `tiup client` which is recommended on https://docs.pingcap.com/tidb/stable/quick-start-with-tidb

The following error is returned:

```sql
my:root@127.0.0.1:4000=> SET tidb_dml_batch_size = 20000;
SET
my:root@127.0.0.1:4000=> LOAD DATA LOCAL INFILE '2017Q1-capitalbikeshare-tripdata.csv' INTO TABLE trips
my:root@127.0.0.1:4000->   FIELDS TERMINATED BY ',' ENCLOSED BY '"'
my:root@127.0.0.1:4000->   LINES TERMINATED BY '\r\n'
my:root@127.0.0.1:4000->   IGNORE 1 LINES
my:root@127.0.0.1:4000-> (duration, start_date, end_date, start_station_number, start_station,
my:root@127.0.0.1:4000(> end_station_number, end_station, bike_number, member_type);
error: mysql: local file '2017Q1-capitalbikeshare-tripdata.csv' is not registered
```

2. Describe your suggestion or addition.

This is because the `tiup client` does not support the file transfer parts of the MySQL protocol. It might be helpful to include a note on this page saying that it requires the real mysql client, or alternatively an example could be given with lightning as well.

(Ideally lightning wouldn't require so many setup steps, since that makes it hard to use as a quickstart.)

3. Provide some reference materials (documents, websites, etc) if you could.

hfxsd:

The step to LOAD DATA INFILE does not work with the client tiup client. Changed it to another method.

 close #6647

- [x] v6.3 (TiDB 6.3 versions)

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases


dveeden:
I'll review this next week. I think some more changes are needed.

dveeden:
I think we should consider the following:

1. There are two documents about this with different procedures

* https://docs.pingcap.com/tidbcloud/import-sample-data
* https://docs.pingcap.com/tidb/dev/import-example-data

It might be good to consider if we need to combine these or not. My take on it is that we should not do this as the procedure for TiDB Cloud uses the cloud UI to import and uses data that was already converted to dumpling format instead of the original CSV data.  We could use the S3 hosted files for the onprem example as well as `tidb-lightning` supports S3.

2. Client strategy

The referenced issue links to https://docs.pingcap.com/tidb/stable/quick-start-with-tidb on which there are examples for "MySQL Client" and `tiup client`. On TiDB Cloud we have WebSQL which is based on https://github.com/xo/usql just like `tiup client`. However there isn't an example for `tiup client` on the TiDB Cloud connect page. It also doesn't look like `tiup client` was intended to be used for anything except for local playgrounds. On this dialog we also list https://github.com/dbcli/mycli. I'm often using "MySQL Shell", which is listed on https://docs.pingcap.com/tidb/stable/dev-guide-connect-to-tidb#mysql-shell as well. There are also many GUI clients that people use like https://github.com/dbeaver/dbeaver and https://github.com/mysql/mysql-workbench. We should make sure the method that we're choosing here aligns with the long term plans for support of clients.

image

Side note: Maybe we should add a note about `tiup client` on https://docs.pingcap.com/tidb/stable/sql-statement-load-data

3. Audience

People that are just starting with TiDB are likely to come from a MySQL background. This means that they might already be familiar with "MySQL Client", `LOAD DATA...`, etc. but not have experience with `tidb-lightning`.

While showing the functionality of Lightning it might also be good to not overwhelm them with new tools.

Maybe we should list both the `tidb-lightning` method and the `LOAD DATA...` method.

4. TiDB Cloud

If we use `LOAD DATA...` or `tidb-lightning` with `tidb` backend then this procedure should also work on TiDB Cloud. Let's try to keep all of this working with TiDB Cloud and document if anything special is needed for that.

5. Using a schema file.

The example uses `no-schema = true` and manually creates the schema. We could put the SQL in a file and let `tidb-lightning` take care of this.

6. The goal

I think the goal of importing the sample data is:
- Giving people some data in TiDB to allow them to play with this with SQL.
  - Maybe add some example queries?
  - What about adding indexes?
  - Using the data in dumpling format from S3 might be faster
- Having people learn how to import data into TiDB
  - Maybe link to https://docs.pingcap.com/tidb/stable/migrate-small-mysql-to-tidb ?
  - The S3 data is in dumpling format, which may be similar to what they eventually use if they use Dumpling to dump data from MySQL.

7. Newer data

https://s3.amazonaws.com/capitalbikeshare-data/index.html has data from 2010 until last month (2022-09). Maybe we should use the newer data. Note that the newer files have different fields.

8. Data amount

To see the full benefits of TiDB it might be needed to load more years of data than the example uses. Maybe list the commands to import all data as well.

9. TiFlash.

Might be good to give an example about TiFlash with this data (adding replica and explain, etc)

10. OSS Insight.

OSS Insight is a full example as it has the data and an opensource application. Maybe we want to either use OSS Insight data here or create a small example frontend for the bikeshare data to give a more complete example.

If we do this we could use some of the example queries from the TiDB Cloud Playground functionality.



Maybe of these points are probably out of scope for this specific PR. For now I think for this PR we should:

1. Keep both methods (`LOAD DATA...` and `tidb-lightning`)
2. Consider putting the SQL in a file for `tidb-lightning`.
3. Add some example queries. Maybe add an index.
4. Load newer data
5. List optional steps to load more data

import-example-data.md
```diff
@@ -21,49 +21,54 @@ unzip \*-tripdata.zip
...
+    no-schema = true
+    data-source-dir = "/home/dvaneeden/bikeshare-data"
+
+    [mydumper.csv]
+    header = true
+
+    [tidb]
+    host = "127.0.0.1"
+    port = 4000
+    user = "root"
```
dveeden, line 67:
```suggestion
    user = "root"
    password = "very_secret"
```

import-example-data.md
```diff
@@ -21,49 +21,54 @@ unzip \*-tripdata.zip
...
+
+3. Create a tidb-lightning.toml file as follows:
+
+    ```toml
+    [tikv-importer]
+    backend = "tidb"
+
+    [mydumper]
+    no-schema = true
+    data-source-dir = "/home/dvaneeden/bikeshare-data"
```
dveeden, line 59:
```suggestion
    data-source-dir = "~/bikeshare-data"
```

Might need testing to see if the `~` gets expanded correctly.

import-example-data.md
```diff
@@ -21,48 +21,55 @@ unzip \*-tripdata.zip
...
+      `start station number` INT,
+      `start station` VARCHAR(255),
+      `end station number` INT,
+      `end station` VARCHAR(255),
+      `bike number` VARCHAR(255),
+      `member type` VARCHAR(255)
+    );
+    ```
+
+3. Create a tidb-lightning.toml file as follows:
```
dveeden, line 51:
```suggestion
3. Create a `tidb-lightning.toml` file as follows:
```

dveeden (approved):
This seems fine. However if I recall correctly not all CSV files had the same set of columns as that changed at some point in time. Does this work correctly for all CSV files?

alastori:
> This seems fine. However if I recall correctly not all CSV files had the same set of columns as that changed at some point in time. Does this work correctly for all CSV files?

Yes. I did a quick test using the procedure and I could import all the files:
```plaintext
#Total files
% ls -1 *.csv | wc -l
      26

# Total lines      
% wc -l *.csv | awk '/total/ {print $1}'
19117669
```

```sql
mysql> SELECT COUNT(*) - 26 AS adjusted_row_count FROM bikeshare.trips;
+--------------------+
| adjusted_row_count |
+--------------------+
|           19117617 |
+--------------------+
1 row in set (0.01 sec)
```

PR #11078, opened 2022-10-28
add MariaDB CLI

zhangyangyu:

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/tidb-cloud/secure-connections-to-serverless-tier-clusters.md
+++ b/tidb-cloud/secure-connections-to-serverless-tier-clusters.md
@@ -37,15 +37,15 @@ TiDB Cloud provides some connection examples in the **Connect** dialog. You can 
 
 Generally, enabling TLS and offering a CA root path to authenticate the server is a good practice to prevent a man-in-the-middle attack. Different clients have different operations in the TLS connection. Enable TLS and verify the server according to your actual use of the client.
 
-The following examples show the connection string in MySQL CLI client, MyCLI client, Java, Python, Go, and Node.js:
+The following examples show the connection string in MySQL CLI client, MariaDB CLI client, MyCLI client, Java, Python, Go, and Node.js:
 
 <SimpleTab>
-<div label="MySQL CLI client">
+<div label="MySQL CLI">
 
 MySQL CLI client attempts to establish TLS connection by default. When you connect to TiDB Serverless Tier clusters, you should set `ssl-mode` and `ssl-ca`.
 
 ```shell
-mysql --connect-timeout 15 -u <username> -h <host> -P 4000 --ssl-mode=VERIFY_IDENTITY --ssl-ca=<CA_root_path> -D test -p
+mysql -u <username> -h <host> -P 4000 --ssl-mode=VERIFY_IDENTITY --ssl-ca=<CA_root_path> -D test -p
 ```
 
 - With `--ssl-mode=VERIFY_IDENTITY`, MySQL CLI client forces to enable TLS and validate TiDB Serverless Tier clusters.
@@ -53,7 +53,20 @@ mysql --connect-timeout 15 -u <username> -h <host> -P 4000 --ssl-mode=VERIFY_IDE
 
 </div>
 
-<div label="MyCLI client">
+<div label="MariaDB CLI">
+
+[MariaDB CLI client](https://mariadb.com/kb/en/mysql-command-line-client/) automatically enables TLS when using TLS related parameters. When you connect to TiDB Serverless Tier clusters, you need to set `ssl-ca` and `ssl-verify-server-cert`.
+
+ ```shell
+mysql -u <username> -h <host> -P 4000 --ssl-ca=<CA_root_path> --ssl-verify-server-cert -D test -p
+ ```
+
+ - Use `--ssl-ca=<CA_root_path>` to set the CA root path on your system.
+ - With `--ssl-verify-server-cert` to validate TiDB Serverless Tier clusters.
+
+ </div>
+
+<div label="MyCLI">
 
 [MyCLI](https://www.mycli.net/) automatically enables TLS when using TLS related parameters. When you connect to TiDB Serverless Tier clusters, you need to set `ssl-ca` and `ssl-verify-server-cert`.
 
--- a/tidb-cloud/secure-connections-to-serverless-tier-clusters.md
+++ b/tidb-cloud/secure-connections-to-serverless-tier-clusters.md
@@ -42,7 +42,7 @@ The following examples show the connection string in MySQL CLI client, MariaDB C
 <SimpleTab>
 <div label="MySQL CLI">
 
-MySQL CLI client attempts to establish TLS connection by default. When you connect to TiDB Serverless Tier clusters, you should set `ssl-mode` and `ssl-ca`.
+[MySQL CLI client](https://dev.mysql.com/doc/refman/8.0/en/mysql.html) attempts to establish TLS connection by default. When you connect to TiDB Serverless Tier clusters, you should set `ssl-mode` and `ssl-ca`.
 
 ```shell
 mysql -u <username> -h <host> -P 4000 --ssl-mode=VERIFY_IDENTITY --ssl-ca=<CA_root_path> -D test -p
@@ -127,7 +127,7 @@ db, err := sql.Open("mysql", "<usename>:<your_password>@tcp(<host>:4000)/test?tl
 
 <div label="Node.js">
 
-[Mysql2](https://www.npmjs.com/package/mysql2)'s TLS connection configurations are used here as an example.
+[mysql2](https://www.npmjs.com/package/mysql2)'s TLS connection configurations are used here as an example.
 
 ```
 host: '<host>', port: 4000,user: '<username>', password: '<your_password>', database: 'test', ssl: {minVersion: 'TLSv1.2', rejectUnauthorized: true}


tidb-cloud/secure-connections-to-serverless-tier-clusters.md
```diff
@@ -37,23 +37,36 @@ TiDB Cloud provides some connection examples in the **Connect** dialog. You can
...
 
 - With `--ssl-mode=VERIFY_IDENTITY`, MySQL CLI client forces to enable TLS and validate TiDB Serverless Tier clusters.
 - Use `--ssl-ca=<CA_root_path>` to set the CA root path on your system.
 
 </div>
 
-<div label="MyCLI client">
+<div label="MariaDB CLI">
+
+[MariaDB CLI client](https://mariadb.com/kb/en/mysql-command-line-client/) automatically enables TLS when using TLS related parameters. When you connect to TiDB Serverless Tier clusters, you need to set `ssl-ca` and `ssl-verify-server-cert`.
```
Daemonxiao, line 58: How about changing `need to `to `should`?  Are these parameters necessary？

zhangyangyu: I think we could remove `need to` or `should`, just plain `set xxxx`.

ran-huang: A side note: Google documentation style guide suggests [avoiding "should"](https://developers.google.cn/style/word-list#should).

ran-huang (approved):

tidb-cloud/secure-connections-to-serverless-tier-clusters.md
```diff
@@ -37,23 +37,36 @@ TiDB Cloud provides some connection examples in the **Connect** dialog. You can
 
 Generally, enabling TLS and offering a CA root path to authenticate the server is a good practice to prevent a man-in-the-middle attack. Different clients have different operations in the TLS connection. Enable TLS and verify the server according to your actual use of the client.
 
-The following examples show the connection string in MySQL CLI client, MyCLI client, Java, Python, Go, and Node.js:
+The following examples show the connection string in MySQL CLI client, MariaDB CLI client, MyCLI client, Java, Python, Go, and Node.js:
```
dveeden, line 40:
```suggestion
The following examples show the connection string in MySQL Client, MariaDB Client, MyCLI client, Java, Python, Go, and Node.js:
```

tidb-cloud/secure-connections-to-serverless-tier-clusters.md
```diff
@@ -37,23 +37,36 @@ TiDB Cloud provides some connection examples in the **Connect** dialog. You can
...
 
 <SimpleTab>
-<div label="MySQL CLI client">
+<div label="MySQL CLI">
 
-MySQL CLI client attempts to establish TLS connection by default. When you connect to TiDB Serverless Tier clusters, you should set `ssl-mode` and `ssl-ca`.
+[MySQL CLI client](https://dev.mysql.com/doc/refman/8.0/en/mysql.html) attempts to establish TLS connection by default. When you connect to TiDB Serverless Tier clusters, you should set `ssl-mode` and `ssl-ca`.
```
dveeden, line 45:
```suggestion
[MySQL Client](https://dev.mysql.com/doc/refman/8.0/en/mysql.html) attempts to establish TLS connection by default. When you connect to TiDB Serverless Tier clusters, you should set `ssl-mode` and `ssl-ca`.
```

tidb-cloud/secure-connections-to-serverless-tier-clusters.md
```diff
@@ -37,23 +37,36 @@ TiDB Cloud provides some connection examples in the **Connect** dialog. You can
...
 
 ```shell
-mysql --connect-timeout 15 -u <username> -h <host> -P 4000 --ssl-mode=VERIFY_IDENTITY --ssl-ca=<CA_root_path> -D test -p
+mysql -u <username> -h <host> -P 4000 --ssl-mode=VERIFY_IDENTITY --ssl-ca=<CA_root_path> -D test -p
```
dveeden, line 48:
```suggestion
mysql --comments -u <username> -h <host> -P 4000 --ssl-mode=VERIFY_IDENTITY --ssl-ca=<CA_root_path> -D test -p
```

tidb-cloud/secure-connections-to-serverless-tier-clusters.md
```diff
@@ -37,23 +37,36 @@ TiDB Cloud provides some connection examples in the **Connect** dialog. You can
...
 ```
 
 - With `--ssl-mode=VERIFY_IDENTITY`, MySQL CLI client forces to enable TLS and validate TiDB Serverless Tier clusters.
 - Use `--ssl-ca=<CA_root_path>` to set the CA root path on your system.
 
 </div>
 
-<div label="MyCLI client">
+<div label="MariaDB CLI">
```
dveeden, line 56:
```suggestion
<div label="MariaDB Client">
```

dveeden (commented):
I don't think we should recommend people to use the MariaDB Client. They should use MySQL Client or mycli. Personally I use MySQL Shell instead of MySQL Client, but I think we decided not to support that either.

As MySQL and MariaDB continue to divert I don't think this is a good choice for the long term.

dveeden:
If we decide to add more clients we may want to consider adding instructions for [DBeaver](https://github.com/dbeaver/dbeaver), [DataGrip](https://www.jetbrains.com/datagrip/) or [MySQL Workbench](https://github.com/mysql/mysql-workbench) instead.

dveeden:
Shouldn't this be merged into `master` and then be backported to v6.1 and newer?

tidb-cloud/secure-connections-to-serverless-tier-clusters.md
zhangyangyu: I deliberately changed to this since it's official called MySQL Command-Line Client. I think CLI could make it more clear. Search MySQL Client on Google will also lead to results like MySQL Shell.

zhangyangyu:
@dveeden This page currently corresponds to the connect tab on TiDB Cloud. So the GUI clients you mentioned are not considered. And as for MariaDB Client, currently it still claims to be compatible with MySQL. And I don't think what we list here is what we recommend. We just want the users who use these clients on their own wishes to quickly find how to connect. Why do I add MariaDB? Because when we are testing Serverless Tier, some colleagues find the mysql client on their OS is actually MariaDB client. And their arguments are not same.

As for the branch, TiDB Cloud documentation resides in release-6.1, not master. It's different.

tidb-cloud/secure-connections-to-serverless-tier-clusters.md
dveeden: Changing it from "MySQL Client" to "MySQL CLI Client" to avoid getting "MySQL Shell" results from Google doesn't seem to work:

image

Looks like "MySQL Client" is the common way of naming this: https://trends.google.com/trends/explore?q=%22MySQL%20Client%22,%22MySQL%20CLI%20Client%22

dveeden:
> @dveeden This page currently corresponds to the connect tab on TiDB Cloud. So the GUI clients you mentioned are not considered. 

Why are the GUI clients not considered?

> And as for MariaDB Client, currently it still claims to be compatible with MySQL. And I don't think what we list here is what we recommend. We just want the users who use these clients on their own wishes to quickly find how to connect. Why do I add MariaDB? Because when we are testing Serverless Tier, some colleagues find the mysql client on their OS is actually MariaDB client. And their arguments are not same.

My own preference would still be to tell people to use the MySQL Client instead of MariaDB. However I don't object to add MariaDB like this PR does. 

Note that it might be good to add MariaDB specific content to other pages that use options that aren't compatible with MariaDB Client:
```
$ git grep -E 'mysql.*--ssl-mode'
tidb-cloud/migrate-from-aurora-bulk-import.md:    mysqldump -h ${export_endpoint} -u ${export_username} -p --ssl-mode=DISABLED -d${export_database} >db.sql
tidb-cloud/secure-connections-to-serverless-tier-clusters.md:mysql --connect-timeout 15 -u <username> -h <host> -P 4000 --ssl-mode=VERIFY_IDENTITY --ssl-ca=<CA_root_path> -D test -p
```

> As for the branch, TiDB Cloud documentation resides in release-6.1, not master. It's different.

Ok, thanks.

tidb-cloud/secure-connections-to-serverless-tier-clusters.md
zhangyangyu: I would like to also change our connect tab to be consistent with this. So I follow PlanetScale and Cockroach Serverless:
image
image
Both "Client" and "CLI" are okay to me.

zhangyangyu:
> Why are the GUI clients not considered?

I don't mean they are totally ignored. Just not considered in this page, or this PR.

> My own preference would still be to tell people to use the MySQL Client instead of MariaDB. However I don't object to add MariaDB like this PR does.
>
> Note that it might be good to add MariaDB specific content to other pages that use options that aren't compatible with MariaDB Client:

The intention to add MariaDB Client is just want users already using MariaDB Client to easily find how to connect, like some of our colleagues.

As for other pages, it sound reasonable but seems will make the wording complex. :-( Not sure we should mention it every time there is some difference.

dveeden:
@zhangyangyu While I still don't really like this, I would be fine to approve this PR if you add the `--comments`.

tidb-cloud/secure-connections-to-serverless-tier-clusters.md
dveeden: @zhangyangyu Could you add the `--comments`?

dveeden:
If we still want to add MariaDB Client then we should be aware of this:
```
/usrbin/mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
/usr/bin/mysql from 11.2.2-MariaDB, client 15.2 for linux-systemd (x86_64) using readline 5.1
```

PR #11091, opened 2022-10-31
add system variable tidb_opt_prefix_index_single_scan

TomShawn:
add system variable tidb_opt_prefix_index_single_scan

- This PR is translated from: https://github.com/pingcap/docs-cn/pull/11713
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/system-variables.md
+++ b/system-variables.md
@@ -2764,6 +2764,69 @@ explain select * from t where age=5;
 3 rows in set (0.00 sec)
 ```
 
+### tidb_opt_prefix_index_single_scan <span class="version-mark">New in v6.4.0</span>
+
+- Scope: SESSION | GLOBAL
+- Persists to cluster: Yes
+- Default value: `ON`
+- This variable controls whether the TiDB optimizer pushes down some filter conditions to the prefix index to avoid unnecessary table lookup and to improve query performance.
+- When this variable value is set to `ON`, the filter conditions are pushed down to the prefix index. Suppose that the `col` column is the index prefix column in a table. The `col is null` or `col is not null` condition in the query is handled as a filter condition on the index instead of a filter condition for the table lookup, so that unnecessary table lookup is avoided.
+
+<details>
+<summary>Usage example of <code>tidb_opt_prefix_index_single_scan</code></summary>
+
+Create a table with prefix index:
+
+```sql
+CREATE TABLE t (a INT, b VARCHAR(10), c INT, INDEX idx_a_b(a, b(5)));
+```
+
+Disable `tidb_opt_prefix_index_single_scan`:
+
+```sql
+SET tidb_opt_prefix_index_single_scan = 'OFF';
+```
+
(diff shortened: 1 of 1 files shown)


system-variables.md
```diff
@@ -2764,6 +2764,69 @@ explain select * from t where age=5;
...
+- Scope: SESSION | GLOBAL
+- Persists to cluster: Yes
+- Default value: `ON`
+- This variable controls whether the TiDB optimizer pushes down some filter conditions to the prefix index to avoid unnecessary table lookup and to improve query performance.
+- When this variable value is set to `ON`, the filter conditions are pushed down to the prefix index. Suppose that the `col` column is the index prefix column in a table. The `col is null` or `col is not null` condition in the query is handled as a filter condition on the index instead of a filter condition for the table lookup, so that unnecessary table lookup is avoided.
+
+<details>
+<summary>Usage example of <code>tidb_opt_prefix_index_single_scan</code></summary>
+
+Create a table with prefix index:
```
hfxsd, line 2778:
```suggestion
Create a table with a prefix index:
```

system-variables.md
```diff
@@ -2764,6 +2764,69 @@ explain select * from t where age=5;
 3 rows in set (0.00 sec)
 ```
 
+### tidb_opt_prefix_index_single_scan <span class="version-mark">New in v6.4.0</span>
+
+- Scope: SESSION | GLOBAL
+- Persists to cluster: Yes
+- Default value: `ON`
+- This variable controls whether the TiDB optimizer pushes down some filter conditions to the prefix index to avoid unnecessary table lookup and to improve query performance.
+- When this variable value is set to `ON`, the filter conditions are pushed down to the prefix index. Suppose that the `col` column is the index prefix column in a table. The `col is null` or `col is not null` condition in the query is handled as a filter condition on the index instead of a filter condition for the table lookup, so that unnecessary table lookup is avoided.
```
hfxsd, line 2773: Both 'index prefix' and 'prefix index' are used. Are they the same term?

hfxsd (approved):
Rest LGTM

system-variables.md
TomShawn: See https://github.com/pingcap/docs-cn/pull/11713#discussion_r1013546552. It is expected.

xuyifangreeneyes, line 2773:
```suggestion
- When this variable value is set to `ON`, some filter conditions are pushed down to the prefix index. Suppose that the `col` column is the index prefix column in a table. The `col is null` or `col is not null` condition in the query is handled as a filter condition on the index instead of a filter condition for the table lookup, so that unnecessary table lookup is avoided.
```
Maybe we need to address that not all filter conditions can be pushed down to prefix column. Only some of them such as `col is null`.

xuyifangreeneyes: 'the filter conditions' refers to 'some filter conditions' in the last sentence? It's ok then.

xuyifangreeneyes (approved):

PR #12275, opened 2023-01-29
add foreign key doc

Oreoxmt:

Translate from: https://github.com/pingcap/docs-cn/pull/11974

TiDB v6.6.0 supports foreign key constrains.

- Update the note/compatibility of foreign key constraints
- Update system-variables: foreign_key_checks and tidb_enable_foreign_key
- Add a new doc to introduce how to use FOREIGN KEY in TiDB

- [x] v6.6 (TiDB 6.6 versions)

- This PR is translated from: https://github.com/pingcap/docs-cn/pull/11974
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases


sql-statements/sql-statement-foreign-key.md
```diff
@@ -0,0 +1,255 @@
...
+
+The effect of disabling [`foreign_key_checks`](/system-variables.md#foreign_key_checks) is as follows:
+
+- When deleting a parent table referenced by a foreign key, the deletion can succeed only when the foreign key constraint check is disabled.
+- When importing data to a database, the order of creating tables might be different from the foreign key dependency order, which might cause the creation of tables to fail. Only when the foreign key constraint check is disabled can the tables be created successfully. In addition, when importing data, disabling the foreign key constraint check can also speed up data import.
+- When importing data to a database, if the data of the child table is imported first, an error will be reported. Only when the foreign key constraint check is disabled can the data of the child table be imported successfully.
+- When executing `ALTER TABLE` operations related to foreign keys, the foreign key constraint check must be enabled to execute successfully.
```
shichun-0415, line 161:
```suggestion
- If an `ALTER TABLE` operation to be executed involves change of the foreign key, this operation succeeds only when the foreign key constraint check is disabled.
```
It seems "关于" in Chinese should be "关闭".

sql-statements/sql-statement-foreign-key.md
```diff
@@ -0,0 +1,255 @@
...
+
+When the foreign key constraint check is disabled, the foreign key constraint check and reference operation are not executed, except for the following scenarios:
+
+- When executing `ALTER TABLE`, if the definition of the foreign key is incorrect, an error is still reported.
```
shichun-0415, line 165: Inconsistent with the Chinese. Please confirm. If we follow the Chinese, the translation should be:
> If the execution of  `ALTER TABLE` might result in wrong definition of the foreign key, an error is still reported during the execution.

sql-statements/sql-statement-foreign-key.md
```diff
@@ -0,0 +1,255 @@
...
+- When deleting the index required by the foreign key, the foreign key must be deleted first. Otherwise, an error is reported when deleting the foreign key.
```
shichun-0415, line 166:
```suggestion
- When deleting the index required by the foreign key, you should delete the foreign key first. Otherwise, an error is reported.
```

sql-statements/sql-statement-foreign-key.md
```diff
@@ -0,0 +1,255 @@
...
+- When creating a foreign key, if it does not meet the conditions or restrictions of the foreign key, an error is reported.
```
shichun-0415, line 167:
```suggestion
- When you create a foreign key that does not meet related conditions or restrictions, an error is reported.
```

sql-statements/sql-statement-foreign-key.md
```diff
@@ -0,0 +1,255 @@
...
+| test         | child         | pid              | fk_1            |
+| test         | product_order | product_category | fk_1            |
+| test         | product_order | product_id       | fk_1            |
+| test         | product_order | customer_id      | fk_2            |
++--------------+---------------+------------------+-----------------+
+```
+
+## View the execution plan with foreign keys
+
+You can use the `EXPLAIN` statement to view the execution plan. The `Foreign_Key_Check` operator is the operator that performs the foreign key constraint check when executing DML statements.
```
shichun-0415, line 205:
```suggestion
## View execution plans with foreign keys

You can use the `EXPLAIN` statement to view execution plans. The `Foreign_Key_Check` operator performs the foreign key constraint check on DML statements that are executed.
```

sql-statements/sql-statement-foreign-key.md
```diff
@@ -0,0 +1,255 @@
...
+mysql> explain insert into child values (1,1);
++-----------------------+---------+------+---------------+-------------------------------+
+| id                    | estRows | task | access object | operator info                 |
++-----------------------+---------+------+---------------+-------------------------------+
+| Insert_1              | N/A     | root |               | N/A                           |
+| └─Foreign_Key_Check_3 | 0.00    | root | table:parent  | foreign_key:fk_1, check_exist |
++-----------------------+---------+------+---------------+-------------------------------+
+```
+
+You can use the `EXPLAIN ANALYZE` statement to view the execution of the foreign key reference behavior. The `Foreign_Key_Cascade` operator is the operator that performs the foreign key reference behavior when executing DML statements.
```
shichun-0415, line 217:
```suggestion
You can use the `EXPLAIN ANALYZE` statement to view the execution of the foreign key reference behavior. The `Foreign_Key_Cascade` operator performs foreign key referencing for DML statements that are executed.
```

sql-statements/sql-statement-foreign-key.md
```diff
@@ -0,0 +1,255 @@
...
++----------------------------------+---------+---------+-----------+---------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------------------+-----------+------+
+```
+
+## Compatibility
+
+### Compatibility between TiDB versions
+
+Before v6.6.0, TiDB supported the syntax of creating foreign keys, but the created foreign keys are ineffective. If you upgrade a TiDB cluster created before v6.6.0 to v6.6.0 or later, the foreign keys created before the upgrade are still ineffective. Only the foreign keys created in v6.6.0 or later versions are effective. You can use the `SHOW CREATE TABLE` statement to check whether the foreign keys are effective. The invalid foreign key has a `/* FOREIGN KEY INVALID */` comment.
```
shichun-0415, line 238:
```suggestion
Before v6.6.0, TiDB supports the syntax of creating foreign keys, but the created foreign keys are ineffective. If you upgrade a TiDB cluster created before v6.6.0 to v6.6.0 or later, the foreign keys created before the upgrade are still ineffective. Only the foreign keys created in v6.6.0 or later versions are effective. You can use the `SHOW CREATE TABLE` statement to check whether the foreign keys are effective. An invalid foreign key has a `/* FOREIGN KEY INVALID */` comment.
```

TOC.md
```diff
@@ -690,6 +690,7 @@
       - [`FLUSH PRIVILEGES`](/sql-statements/sql-statement-flush-privileges.md)
       - [`FLUSH STATUS`](/sql-statements/sql-statement-flush-status.md)
       - [`FLUSH TABLES`](/sql-statements/sql-statement-flush-tables.md)
+      - [`FOREIGN KEY`](/sql-statements/sql-statement-foreign-key.md)
```
dveeden, line 693: This looks wrong to me. I don't think there is a `FOREIGN KEY` statement, I think Foreign Keys are part of `CREATE TABLE` statements etc isn't it?

system-variables.md
```diff
@@ -270,8 +270,8 @@ For more possible values of this variable, see [Authentication plugin status](/s
 - Scope: SESSION | GLOBAL
 - Persists to cluster: Yes
 - Type: Boolean
-- Default value: `OFF`
-- For compatibility, TiDB returns foreign key checks as `OFF`.
+- Default value: `ON`
```
dveeden, line 273:
```suggestion
- Default value: `ON` (since v6.6.0)
```

Or is there a better way of showing when this changed? Now we only mark new variables with a version, but not important variable changes.

system-variables.md
```diff
@@ -270,8 +270,8 @@ For more possible values of this variable, see [Authentication plugin status](/s
...
+- Indicates whether foreign key constraint checking is enabled.
```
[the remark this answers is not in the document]
dveeden:
```suggestion
- Indicates whether foreign key constraint checking is enabled or not.
```

Same for the suggestion from @shichun-0415

dveeden (commented):
This also needs updating:
- https://github.com/pingcap/docs/blob/master/dm/dm-precheck.md?plain=1#L55

sql-statements/sql-statement-foreign-key.md
```diff
@@ -0,0 +1,255 @@
...
+summary: An overview of the usage of FOREIGN KEY for the TiDB database.
+---
+
+# FOREIGN KEY
+
+Starting from v6.6.0, TiDB supports the foreign key feature, which allows cross-table referencing of related data, and foreign key constraints to maintain data consistency.
+
+The foreign key is defined in the child table. The syntax is as follows:
+
+```sql
```
dveeden, line 12: I think we should use `ebnf+diagram` here?

sql-statements/sql-statement-foreign-key.md
```diff
@@ -0,0 +1,255 @@
...
+- When importing data to a database, if the data of the child table is imported first, an error will be reported. Only when the foreign key constraint check is disabled can the data of the child table be imported successfully.
+- When executing `ALTER TABLE` operations related to foreign keys, the foreign key constraint check must be enabled to execute successfully.
+
+When the foreign key constraint check is disabled, the foreign key constraint check and reference operation are not executed, except for the following scenarios:
+
+- When executing `ALTER TABLE`, if the definition of the foreign key is incorrect, an error is still reported.
+- When deleting the index required by the foreign key, the foreign key must be deleted first. Otherwise, an error is reported when deleting the foreign key.
+- When creating a foreign key, if it does not meet the conditions or restrictions of the foreign key, an error is reported.
+
+## Lock
```
dveeden, line 169:
```suggestion
## Locking
```

sql-statements/sql-statement-foreign-key.md
```diff
@@ -0,0 +1,255 @@
...
+       Table: child
+Create Table: CREATE TABLE `child` (
+  `id` int(11) DEFAULT NULL,
+  `pid` int(11) DEFAULT NULL,
+  KEY `idx_pid` (`pid`),
+  CONSTRAINT `fk_1` FOREIGN KEY (`pid`) REFERENCES `test`.`parent` (`id`) ON DELETE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
+```
+
+You can get information about foreign keys from the `INFORMATION_SCHEMA.KEY_COLUMN_USAGE` system table. The following is an example query:
```
dveeden, line 189:
```suggestion
You can get information about foreign keys from the [`INFORMATION_SCHEMA.KEY_COLUMN_USAGE`](information-schema/information-schema-key-column-usage.md) system table. The following is an example query:
```

sql-statements/sql-statement-foreign-key.md
```diff
@@ -0,0 +1,255 @@
...
+  `id` int(11) DEFAULT NULL,
+  `pid` int(11) DEFAULT NULL,
+  KEY `idx_pid` (`pid`),
+  CONSTRAINT `fk_1` FOREIGN KEY (`pid`) REFERENCES `test`.`parent` (`id`) ON DELETE CASCADE /* FOREIGN KEY INVALID */
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
+```
+
+### Compatibility with MySQL
+
+- [TiDB Binlog](/tidb-binlog/tidb-binlog-overview.md) does not support foreign keys.
```
dveeden, line 254: This doesn't below under a "Compatibility with MySQL" header. Maybe put this on a different header or rename the section

dveeden: Maybe also document compatibility with Dumpling, Lightning, DM, CDC and sync-diff-inspector etc.

dveeden (changes requested):
The only real issue I see with this is that this is listed as a statement, while it isn't a statement.

dveeden:
Does `ADMIN CHECK TABLE` validate foreign keys? If not, should we list this as limitation?

sql-statements/sql-statement-foreign-key.md
```diff
@@ -0,0 +1,255 @@
...
+- When deleting the index required by the foreign key, the foreign key must be deleted first. Otherwise, an error is reported when deleting the foreign key.
+- When creating a foreign key, if it does not meet the conditions or restrictions of the foreign key, an error is reported.
+
+## Lock
+
+When `INSERT` or `UPDATE` a child table, the foreign key constraint checks whether the corresponding foreign key value exists in the parent table, and locks the row in the parent table to avoid the foreign key value being deleted by other operations violating the foreign key constraint. The locking behavior is equivalent to performing a `SELECT FOR UPDATE` operation on the row where the foreign key value is located in the parent table. Because TiDB currently does not support `LOCK IN SHARE MODE`, if there are a lot of concurrent writes to the child table and most of the referenced foreign key values are the same, there might be serious lock conflicts. It is recommended to disable [`foreign_key_checks`](/system-variables.md#foreign_key_checks) when writing a large number of child table data.
+
+## Definition and metadata of foreign keys
```
dveeden, line 173: Looks like `information_schema.table_constraints` should also be mentioned.

```
sql> SELECT * FROM information_schema.table_constraints WHERE CONSTRAINT_TYPE='FOREIGN KEY'\G
*************************** 1. row ***************************
CONSTRAINT_CATALOG: def
 CONSTRAINT_SCHEMA: test
   CONSTRAINT_NAME: fk_1
      TABLE_SCHEMA: test
        TABLE_NAME: t2
   CONSTRAINT_TYPE: FOREIGN KEY
1 row in set (0.0025 sec)
```

And `information_schema.referential_constraints`:
```
sql> TABLE information_schema.referential_constraints\G
*************************** 1. row ***************************
       CONSTRAINT_CATALOG: def
        CONSTRAINT_SCHEMA: test
          CONSTRAINT_NAME: fk_1
UNIQUE_CONSTRAINT_CATALOG: def
 UNIQUE_CONSTRAINT_SCHEMA: test
   UNIQUE_CONSTRAINT_NAME: PRIMARY
             MATCH_OPTION: NONE
              UPDATE_RULE: NO ACTION
              DELETE_RULE: CASCADE
               TABLE_NAME: t2
    REFERENCED_TABLE_NAME: t1
1 row in set (0.0021 sec)
```

dveeden:
https://github.com/pingcap/docs/blob/a7f44a289af83a4f9081374abf0eab941b983d6a/information-schema/information-schema-referential-constraints.md?plain=1#L8 needs to be updated as well

dveeden:
Looks like the `ebnf+diagram` of `sql-statements/sql-statement-create-table.md` is missing `FOREIGN KEY`

dveeden:
Would replicating between two TiDB clusters with TiCDC be a problem with foreign keys?

dveeden:
When upgrading there could be previously created foreign keys.

1. Would these referencing invalid tables be an issue?
2. Is there something like `foreign_key_checks=ONLY_WARNINGS` or some other way to not have foreign key checks disabled but to find out where there might be problems?

sql-statements/sql-statement-foreign-key.md
crazycs520: I already change the Chinese version, and this is lgtm.

TOC.md
crazycs520: Great, Maybe use `FOREIGN KEY Constraints` is better, just like the MySQL name. https://dev.mysql.com/doc/refman/8.0/en/create-table-foreign-keys.html

sql-statements/sql-statement-foreign-key.md
crazycs520: Great, Nice catch, I will add those in https://github.com/pingcap/docs-cn/pull/11974 first.

TOC.md
dveeden: Yes, they document foreign key constraints as part of the `CREATE TABLE` statement.

crazycs520:
> Looks like the `ebnf+diagram` of `sql-statements/sql-statement-create-table.md` is missing `FOREIGN KEY`

nice catch, I already added in https://github.com/pingcap/docs-cn/pull/11974/commits/22fa1ab91f8b307285e70805e4a963a1a0525c56

> Would replicating between two TiDB clusters with TiCDC be a problem with foreign keys?

TiCDC v6.6.0 is compatible with foreign key. And I already added `与 TiDB 工具的兼容性` part in https://github.com/pingcap/docs-cn/pull/11974/files#diff-9e0ae4af12c0d4f83009d64785384b1c6cd6eac10092ccedfa02aeed2853d2b5R288 , It will then be added to this PR.

>When upgrading there could be previously created foreign keys.
>
>1. Would these referencing invalid tables be an issue?

As mentioned here: https://github.com/pingcap/docs/pull/12275/files#diff-9e0ae4af12c0d4f83009d64785384b1c6cd6eac10092ccedfa02aeed2853d2b5R236 , the previously created foreign keys are invalid, TiDB will ignore the invalid foreign key.

> 2. Is there something like foreign_key_checks=ONLY_WARNINGS or some other way to not have foreign key checks disabled but to find out where there might be problems?

No. Do you mean checking whether the foreign key constraint is satisfied? Currently, I don't provide any syntax to check it since MySQL doesn't either.

sql-statements/sql-statement-foreign-key.md
```diff
@@ -0,0 +1,312 @@
...
+Table        | child
+Create Table | CREATE TABLE `child` (
+  `id` int(11) DEFAULT NULL,
+  `pid` int(11) DEFAULT NULL,
+  KEY `idx_pid` (`pid`),
+  CONSTRAINT `fk_1` FOREIGN KEY (`pid`) REFERENCES `test`.`parent` (`id`) ON DELETE CASCADE /* FOREIGN KEY INVALID */
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
+```
+
+### Compatibility with TiDB tools
```
Oreoxmt, line 290: @qiancai PTAL

TOC-tidb-cloud.md
```diff
@@ -297,6 +297,7 @@
       - [`FLUSH PRIVILEGES`](/sql-statements/sql-statement-flush-privileges.md)
       - [`FLUSH STATUS`](/sql-statements/sql-statement-flush-status.md)
       - [`FLUSH TABLES`](/sql-statements/sql-statement-flush-tables.md)
+      - [`FOREIGN KEY Constraints`](/sql-statements/sql-statement-foreign-key.md)
```
Oreoxmt, line 300: @qiancai PTAL

dveeden:
> > When upgrading there could be previously created foreign keys.
> > 
> > 1. Would these referencing invalid tables be an issue?
> 
> As mentioned here: https://github.com/pingcap/docs/pull/12275/files#diff-9e0ae4af12c0d4f83009d64785384b1c6cd6eac10092ccedfa02aeed2853d2b5R236 , the previously created foreign keys are invalid, TiDB will ignore the invalid foreign key.

So when upgrading and then looking at the `SHOW CREATE TABLE` output the foreign key won't be active if `/* FOREIGN KEY INVALID */` is shown?

This isn't really matching with what we do for other functionality:
```
sql> SHOW CREATE TABLE t1\G
*************************** 1. row ***************************
       Table: t1
Create Table: CREATE TABLE `t1` (
  `id` bigint(20) NOT NULL /*T![auto_rand] AUTO_RANDOM(5) */,
  PRIMARY KEY (`id`) /*T![clustered_index] CLUSTERED */
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
1 row in set (0.0031 sec)
```

Here the `/*T![...] ... */` comments are used for feature dependent things.

We now have this:
```
  CONSTRAINT `fk_1` FOREIGN KEY (`pid`) REFERENCES `test`.`parent` (`id`)
    ON DELETE CASCADE /* FOREIGN KEY INVALID */
```

Something like this might have been better:
```
  CONSTRAINT `fk_1` FOREIGN KEY (`pid`) REFERENCES `test`.`parent` (`id`)
    ON DELETE CASCADE /*T![foreignkey] FOREIGN KEY NOT ENFORCED */
```

1. When copy pasting `SHOW CREATE TABLE` output from TiDB v6.5.0 or earlier to TiDB v6.6.0 or newer this would keep the same behaviour.
2. This could help to bring a feature that MySQL doesn't have but is often requested: foreign keys that are not enforced. This would help tools like MySQL Workbench to find relations between tables and be a way to document relations between tables.

However I don't expect that we're going to change this now.

constraints.md
```diff
@@ -384,7 +384,7 @@ ALTER TABLE orders ADD FOREIGN KEY fk_user_id (user_id) REFERENCES users(id);
 
 ### Notes
 
-* TiDB supports foreign keys to avoid errors caused by this syntax when you migrate data from other databases to TiDB.
+* Before v6.6.0, TiDB supports foreign keys to avoid errors caused by this syntax when you migrate data from other databases to TiDB.
 
     However, TiDB does not perform constraint checking on foreign keys in DML statements. For example, even if there is no record with id=123 in the users table, the following transactions can be submitted successfully.
```
qiancai, line 389: Is line 389 applicable to v6.6.0? If not, can we remove the line 385~389 for v6.6.0 docs?

dveeden (changes requested):
The filename (`sql-statement-foreign-key.md`) is still showing this as if it were a statement, which it isn't. This needs to be changed. Otherwise I don't see any real issues with this PR.

crazycs520:
@dveeden Thanks for the Patient review.

> So when upgrading and then looking at the `SHOW CREATE TABLE` output the foreign key won't be active if `/* FOREIGN KEY INVALID */` is shown?
> 
> This isn't really matching with what we do for other functionality:
> 
> ```
> sql> SHOW CREATE TABLE t1\G
> *************************** 1. row ***************************
>        Table: t1
> Create Table: CREATE TABLE `t1` (
>   `id` bigint(20) NOT NULL /*T![auto_rand] AUTO_RANDOM(5) */,
>   PRIMARY KEY (`id`) /*T![clustered_index] CLUSTERED */
> ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
> 1 row in set (0.0031 sec)
> ```
> 
> Here the `/*T![...] ... */` comments are used for feature dependent things.
> 
> We now have this:
> 
> ```
>   CONSTRAINT `fk_1` FOREIGN KEY (`pid`) REFERENCES `test`.`parent` (`id`)
>     ON DELETE CASCADE /* FOREIGN KEY INVALID */
> ```
> 
> Something like this might have been better:
> 
> ```
>   CONSTRAINT `fk_1` FOREIGN KEY (`pid`) REFERENCES `test`.`parent` (`id`)
>     ON DELETE CASCADE /*T![foreignkey] FOREIGN KEY NOT ENFORCED */
> ```
> 
> 1. When copy pasting `SHOW CREATE TABLE` output from TiDB v6.5.0 or earlier to TiDB v6.6.0 or newer this would keep the same behaviour.
> 2. This could help to bring a feature that MySQL doesn't have but is often requested: foreign keys that are not enforced. This would help tools like MySQL Workbench to find relations between tables and be a way to document relations between tables.
> 

The comment format may need more discuss with PM. But now that the release is approaching, v6.6.0 has no time to change


> What's the way to make these invalid foreign keys from pre-upgrade to become active? Is there something like `ALTER TABLE ... MODIFY CONSTRAINT `fk_1` ENABLE`?

We did not provide any way. The only way is drop the invalid foreign key first and then add the foreign key again.

>  Or do we have a tool/script to loop over all tables/foreign keys and drop and add them to make them active?

I have some doubts about whether this tool is necessary. If the user is use old TiDB cluster and then upgrade to v6.6.0 TiDB cluster and have some `invailid foreign key`, the user should not want to make the `invalid foreign key` to be valid, 
since the user application must already handle the foreign key invalid situation, if convert the `invalid foreign key` to be valid, the user application must meet something wrong.
So, for the `invalid foreign key`, it must be confirmed by the user before handling, and cannot be simply handled with tools or scripts

dveeden:
@crazycs520 Yes I understand that we can't change the comment format at this point. 

I think a tools that goes over invalid foreign keys and checks if they reference an existing table/column and have any pre-existing fk violations would be nice together with a warning only setting. But that's by no means required.

Could I create new tables on v6.6.0 with a invalid foreign key by using the same comment as the output from `SHOW CREATE TABLE` uses?

crazycs520:
@dveeden 
> Could I create new tables on v6.6.0 with a invalid foreign key by using the same comment as the output from `SHOW CREATE TABLE` uses?

No, you can't create a invalid foreign key. the comment will be ignore.

constraints.md
Oreoxmt: Done https://github.com/pingcap/docs/pull/12275/commits/0f5109e2a71f49e25023f14a69042350c848b5f8

Oreoxmt:
> The filename (sql-statement-foreign-key.md) is still showing this as if it were a statement, which it isn't. This needs to be changed.

@dveeden The file is renamed to `foreign-key.md` and is moved to the same level as **Views** https://github.com/pingcap/docs/commit/0f5109e2a71f49e25023f14a69042350c848b5f8

dveeden (approved):

sql-statements/sql-statement-create-table.md
```diff
@@ -225,8 +225,6 @@ mysql> DESC t1;
 * The `COMMENT` attribute does not support the `WITH PARSER` option.
 * TiDB supports at most 512 columns in a single table. The corresponding number limit in InnoDB is 1017, and the hard limit in MySQL is 4096. For details, see [TiDB Limitations](/tidb-limitations.md).
 * For partitioned tables, only Range, Hash and Range Columns (single column) are supported. For details, see [partitioned table](/partitioned-table.md).
-* `CHECK` constraints are parsed but ignored (MySQL 5.7 compatible behavior). For details, see [Constraints](/constraints.md).
```
crazycs520, line 228: Should keep this.

sql-statements/sql-statement-create-table.md
```diff
@@ -225,8 +225,6 @@ mysql> DESC t1;
 * The `COMMENT` attribute does not support the `WITH PARSER` option.
```
crazycs520, line 225: Also need to update ebnf, see https://github.com/pingcap/docs-cn/pull/11974/files#diff-05d7fa31873281b59e53a89cec973d1c909ee9fe91fde46ce677bb4700cd4cd7R62

crazycs520 (approved):

foreign-key.md
```diff
@@ -0,0 +1,312 @@
...
+- When deleting the index required by the foreign key, you should delete the foreign key first. Otherwise, an error is reported.
+- When you create a foreign key but it does not meet related conditions or restrictions, an error is reported.
+
+## Locking
+
+When `INSERT` or `UPDATE` a child table, the foreign key constraint checks whether the corresponding foreign key value exists in the parent table, and locks the row in the parent table to avoid the foreign key value being deleted by other operations violating the foreign key constraint. The locking behavior is equivalent to performing a `SELECT FOR UPDATE` operation on the row where the foreign key value is located in the parent table.
+
+Because TiDB currently does not support `LOCK IN SHARE MODE`, if there are a lot of concurrent writes to the child table and most of the referenced foreign key values are the same, there might be serious locking conflicts. It is recommended to disable [`foreign_key_checks`](/system-variables.md#foreign_key_checks) when writing a large number of child table data.
```
shichun-0415, line 178:
```suggestion
Because TiDB currently does not support `LOCK IN SHARE MODE`, if a child table accepts a large number of concurrent writes and most of the referenced foreign key values are the same, there might be serious locking conflicts. It is recommended to disable [`foreign_key_checks`](/system-variables.md#foreign_key_checks) when writing a large number of child table data.
```

foreign-key.md
```diff
@@ -0,0 +1,312 @@
...
+
+## Definition and metadata of foreign keys
+
+To view the definition of a foreign key constraint, use the `SHOW CREATE TABLE` statement:
```
shichun-0415, line 182:
```suggestion
To view the definition of a foreign key constraint, execute the [`SHOW CREATE TABLE`](/sql-statements/sql-statement-show-create-table.md) statement:
```

foreign-key.md
```diff
@@ -0,0 +1,312 @@
...
+*************************** 1. row ***************************
+       Table: child
+Create Table: CREATE TABLE `child` (
+  `id` int(11) DEFAULT NULL,
+  `pid` int(11) DEFAULT NULL,
+  KEY `idx_pid` (`pid`),
+  CONSTRAINT `fk_1` FOREIGN KEY (`pid`) REFERENCES `test`.`parent` (`id`) ON DELETE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
+```
+
```
shichun-0415, line 195:
```suggestion

You can also get information about foreign keys using either of the following system table:

- [`INFORMATION_SCHEMA.KEY_COLUMN_USAGE`](/information-schema/information-schema-key-column-usage.md)
- [`INFORMATION_SCHEMA.TABLE_CONSTRAINTS`](/information-schema/information-schema-table-constraints.md)
- [`INFORMATION_SCHEMA.TABLE_CONSTRAINTS`](/information-schema/information-schema-table-constraints.md)

The following provides examples:

```

foreign-key.md
```diff
@@ -0,0 +1,312 @@
...
+You can get information about foreign keys from the [`INFORMATION_SCHEMA.KEY_COLUMN_USAGE`](/information-schema/information-schema-key-column-usage.md) system table. The following is an example:
```
shichun-0415, line 196:
```suggestion
Get information about foreign keys from the `INFORMATION_SCHEMA.KEY_COLUMN_USAGE` system table:
```

foreign-key.md
```diff
@@ -0,0 +1,312 @@
...
+| TABLE_SCHEMA | TABLE_NAME    | COLUMN_NAME      | CONSTRAINT_NAME |
++--------------+---------------+------------------+-----------------+
+| test         | child         | pid              | fk_1            |
+| test         | product_order | product_category | fk_1            |
+| test         | product_order | product_id       | fk_1            |
+| test         | product_order | customer_id      | fk_2            |
++--------------+---------------+------------------+-----------------+
+```
+
+You can get information about foreign keys from the [`INFORMATION_SCHEMA.TABLE_CONSTRAINTS`](/information-schema/information-schema-table-constraints.md) system table. The following is an example:
```
shichun-0415, line 210:
```suggestion
Get information about foreign keys from the `INFORMATION_SCHEMA.TABLE_CONSTRAINTS` system table:
```

foreign-key.md
```diff
@@ -0,0 +1,312 @@
...
+***************************[ 1. row ]***************************
+CONSTRAINT_CATALOG | def
+CONSTRAINT_SCHEMA  | test
+CONSTRAINT_NAME    | fk_1
+TABLE_SCHEMA       | test
+TABLE_NAME         | child
+CONSTRAINT_TYPE    | FOREIGN KEY
+```
+
+You can get information about foreign keys from the [`INFORMATION_SCHEMA.REFERENTIAL_CONSTRAINTS`](/information-schema/information-schema-referential-constraints.md) system table. The following is an example:
```
shichun-0415, line 223:
```suggestion
Get information about foreign keys from the `INFORMATION_SCHEMA.REFERENTIAL_CONSTRAINTS` system table:
```

shichun-0415: Please update the Chinese as well if these suggestions are accepted.

foreign-key.md
```diff
@@ -0,0 +1,312 @@
...
+  CONSTRAINT `fk_1` FOREIGN KEY (`pid`) REFERENCES `test`.`parent` (`id`) ON DELETE CASCADE /* FOREIGN KEY INVALID */
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
+```
+
+### Compatibility with TiDB tools
+
+<CustomContent platform="tidb">
+
+- [TiDB Binlog](/tidb-binlog/tidb-binlog-overview.md) does not support foreign keys.
+- [DM](/dm/dm-overview.md) v6.6.0 disables the [`foreign_key_checks`](/system-variables.md#foreign_key_checks) of the downstream TiDB when replicating data to TiDB. Therefore, the cascading operations caused by foreign keys are not replicated from the upstream to the downstream, which might cause inconsistent data. Because TiDB does not support foreign keys before v6.6.0, this behavior is consistent with the previous DM versions.
```
shichun-0415, line 295:
```suggestion
- [DM](/dm/dm-overview.md) v6.6.0 disables the [`foreign_key_checks`](/system-variables.md#foreign_key_checks) of the downstream TiDB when replicating data to TiDB. Therefore, the cascading operations caused by foreign keys are not replicated from the upstream to the downstream, which might cause data inconsistency. Because TiDB does not support foreign keys before v6.6.0, this behavior is consistent with the previous DM versions.
```

foreign-key.md
```diff
@@ -0,0 +1,312 @@
...
+- [TiCDC](/ticdc/ticdc-overview.md) v6.6.0 is compatible with foreign keys. The previous versions of TiCDC might report an error when replicating tables with foreign keys. It is recommended to disable the `foreign_key_checks` of the downstream TiDB cluster when using a TiCDC version earlier than v6.6.0.
+- [br](/br/backup-and-restore-overview.md) v6.6.0 is compatible with foreign keys. The previous versions of br might report an error when restoring tables with foreign keys to a v6.6.0 or later cluster. It is recommended to disable the `foreign_key_checks` of the downstream TiDB cluster before restoring the cluster when using a br version earlier than v6.6.0.
```
shichun-0415, line 297:
```suggestion
- [BR](/br/backup-and-restore-overview.md) v6.6.0 is compatible with foreign keys. The previous versions of BR might report an error when restoring tables with foreign keys to a v6.6.0 or later cluster. It is recommended to disable the `foreign_key_checks` of the downstream TiDB cluster before restoring the cluster when using a BR earlier than v6.6.0.
```

foreign-key.md
```diff
@@ -0,0 +1,320 @@
...
+       Table: child
+Create Table: CREATE TABLE `child` (
+  `id` int(11) DEFAULT NULL,
+  `pid` int(11) DEFAULT NULL,
+  KEY `idx_pid` (`pid`),
+  CONSTRAINT `fk_1` FOREIGN KEY (`pid`) REFERENCES `test`.`parent` (`id`) ON DELETE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
+```
+
+You can also get information about foreign keys using either of the following system table:
```
shichun-0415, line 196:
```suggestion
You can also get information about foreign keys using either of the following system tables:
```

foreign-key.md
Oreoxmt: @shichun-0415 The docs-cn PR is updated: https://github.com/pingcap/docs-cn/pull/11974/commits/52cbed4cd9c05805798c6074ebf5d9492bb26922

shichun-0415 (approved):

PR #12536, opened 2023-02-16
explain for connection: fix the compatibility with MySQL

Oreoxmt:

`EXPLAIN FOR CONNECTION` is used to get the execution plan of the currently executed SQL query or the last executed SQL query in a connection.

Fix the compatibility to make it consistent with https://github.com/pingcap/tidb/pull/10030

```
Differences with MySQL EXPLAIN FOR

If one connection is sleeping, query for query plan for that connection will get nothing in MySQL. However, in this PR, query play for last query will be returned.
Side effect of 1., EXPLAIN FOR CONNECTION <id of this connection> will get error in MySQL, but will get query plan of last query in this PR.
If one without ProcessPriv queries plan for another user's connection, MySQL won't return error if that connection does not running any queries, however, in this PR, an error will always returns, even if that connection does not run any queries yet.
```

- [ ] v6.6 (TiDB 6.6 versions)

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/sql-statements/sql-statement-explain.md
+++ b/sql-statements/sql-statement-explain.md
@@ -340,9 +340,8 @@ In the output, `id`, `estRows`, `taskType`, `accessObject`, and `operatorInfo` h
 
 ### `EXPLAIN FOR CONNECTION`
 
-`EXPLAIN FOR CONNECTION` is used to get the execution plan of the currently executed SQL query or the last executed SQL query in a connection. The output format is the same as that of `EXPLAIN`. However, the implementation of `EXPLAIN FOR CONNECTION` in TiDB is different from that in MySQL. Their differences (apart from the output format) are listed as follows:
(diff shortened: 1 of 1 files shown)


sql-statements/sql-statement-explain.md
```diff
@@ -342,7 +342,8 @@ In the output, `id`, `estRows`, `taskType`, `accessObject`, and `operatorInfo` h
 
 `EXPLAIN FOR CONNECTION` is used to get the execution plan of the currently executed SQL query or the last executed SQL query in a connection. The output format is the same as that of `EXPLAIN`. However, the implementation of `EXPLAIN FOR CONNECTION` in TiDB is different from that in MySQL. Their differences (apart from the output format) are listed as follows:
 
-- MySQL returns the query plan that is **being executing**, while TiDB returns the **last executed** query plan.
+- If the connection is sleeping, MySQL returns an empty result, while TiDB returns the last executed query plan.
```
hfxsd, line 345: sleeping 还是 idle？

Oreoxmt: - https://github.com/pingcap/tidb/pull/10030 introduces this feature and uses "sleeping" when describing compatibility.

- To check MySQL connection status, we can use ```show status where `variable_name`=`threads_connected``` and "sleeping" is also used in this MySQL doc https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html

hfxsd (approved):

PR #14480, opened 2023-08-10
Clarify swap descriptions 

hfxsd:

Clarify confusing descriptions for disabling system swap

- [x] v7.3 (TiDB 7.3 versions)
- [x] v7.2 (TiDB 7.2 versions)
- [x] v7.1 (TiDB 7.1 versions)
- [ ] v7.0 (TiDB 7.0 versions)

- This PR is translated from: https://github.com/pingcap/docs-cn/pull/14662
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/check-before-deployment.md
+++ b/check-before-deployment.md
@@ -112,21 +112,25 @@ Take the `/dev/nvme0n1` data disk as an example:
 
 ## Check and disable system swap
 
-TiDB needs sufficient memory space for operation. When memory is insufficient, using swap as a buffer might degrade performance. Therefore, it is recommended to disable the system swap permanently by executing the following commands:
+TiDB needs sufficient memory space for operation. If you want to maintain stable performance, it is recommended that you permanently disable the system swap, but it might trigger OOM issues when there is insufficient memory. If you want to avoid such OOM issues, you can just decrease the swap priority, instead of permanently disabling it.
 
-{{< copyable "shell-regular" >}}
+- Enabling and using swap might introduce performance jitter issues. It is recommended that you permanently disable the operating system tier swap for low-latency and stability-critical database services. To permanently disable swap, you can use the following method:
 
-```bash
-echo "vm.swappiness = 0">> /etc/sysctl.conf
-swapoff -a && swapon -a
-sysctl -p
-```
+    - During the initialization phase of the operating system, do not partition the swap partition disk separately.
+    - If you have already partitioned a separate swap partition disk during the initialization phase of the operating system and enabled swap, run the following command to disable it:
 
-> **Note:**
->
-> - Executing `swapoff -a` and then `swapon -a` is to refresh swap by dumping data to memory and cleaning up swap. If you drop the swappiness change and execute only `swapoff -a`, swap will be enabled again after you restart the system.
->
-> - `sysctl -p` is to make the configuration effective without restarting the system.
+        ```bash
+        echo "vm.swappiness = 0">> /etc/sysctl.conf
+        swapoff -a
+        sysctl -p
+        ```
+
+- If the host memory is insufficient, disabling the system swap might be more likely to trigger OOM issues. You can run the following command to decrease the swap priority instead of disabling it permanently:
+
+    ```bash
+    echo "vm.swappiness = 0">> /etc/sysctl.conf
+    sysctl -p
+    ```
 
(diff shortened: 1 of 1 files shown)


bb7133 (approved):
LGTM

check-before-deployment.md
```diff
@@ -112,21 +112,25 @@ Take the `/dev/nvme0n1` data disk as an example:
 
 ## Check and disable system swap
 
-TiDB needs sufficient memory space for operation. When memory is insufficient, using swap as a buffer might degrade performance. Therefore, it is recommended to disable the system swap permanently by executing the following commands:
+TiDB needs sufficient memory space for operation. If you want to maintain stable performance, it is recommended that you permanently disable the system swap, but it might trigger OOM issues when there is insufficient memory. If you want to avoid such OOM issues, you can just decrease the swap priority, instead of permanently disabling it.
```
dveeden, line 115:
```suggestion
TiDB needs a sufficient amount of memory for operation. If the memory that TiDB uses gets swapped out and later gets swapped back in this can cause latency spikes. If you want to maintain stable performance, it is recommended that you permanently disable the system swap, but it might trigger OOM issues when there is insufficient memory. If you want to avoid such OOM issues, you can just decrease the swap priority, instead of permanently disabling it.
```

check-before-deployment.md
```diff
@@ -112,21 +112,25 @@ Take the `/dev/nvme0n1` data disk as an example:
...
+    - If you have already partitioned a separate swap partition disk during the initialization phase of the operating system and enabled swap, run the following command to disable it:
 
-> **Note:**
->
-> - Executing `swapoff -a` and then `swapon -a` is to refresh swap by dumping data to memory and cleaning up swap. If you drop the swappiness change and execute only `swapoff -a`, swap will be enabled again after you restart the system.
->
-> - `sysctl -p` is to make the configuration effective without restarting the system.
+        ```bash
+        echo "vm.swappiness = 0">> /etc/sysctl.conf
+        swapoff -a
```
dveeden, line 124: This is not sufficient. The `swapoff -a` command only disables the swap until the next reboot of the system. One also needs to remove this from `/etc/fstab` of mask the systemd unit for swap.

dveeden (approved):
Could we configure this on a per-process basis? e.g. [`MemorySwapMax`](https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html#MemorySwapMax=bytes) in systemd unit specifications? This might allow TiDB to not swap out but allow other things to be swapped out.

check-before-deployment.md
hfxsd: @bb7133 PTAL

qiancai:
Removed the `needs-cherry-pick-release-7.2` label because the v7.2 docs have been archived at https://docs-archive.pingcap.com/tidb/v7.2 and will no longer receive new updates.

qiancai:
Removed the needs-cherry-pick-release-7.3 label because the v7.3 docs have been archived at https://docs-archive.pingcap.com/tidb/v7.3 and will no longer receive new updates.

check-before-deployment.md
dveeden: https://rageagainstshell.com/2019/12/disable-swap-in-systemd/ has a good description for systemd based systems

dveeden: For now I would suggest to keep it closer to the original statement:

```bash
echo "vm.swappiness = 0">> /etc/sysctl.conf
sysctl -p
swapoff -a && swapon -a
```

This sets `swappiness` to 0 and then disables/enables swap to for things that have already been swapped out to be swapped back in.

If we leave swap disabled, but don't configure this in a way that survives a reboot this is like leaving a booby trap.

hfxsd, line 124:
```suggestion
        echo "vm.swappiness = 0">> /etc/sysctl.conf
        sysctl -p
        swapoff -a && swapon -a
```

check-before-deployment.md
```diff
@@ -112,21 +112,25 @@ Take the `/dev/nvme0n1` data disk as an example:
...
+        sysctl -p
```
hfxsd, line 125:
```suggestion
```

lilin90 (approved):
/lgtm

PR #15443, opened 2023-11-23
Update keywords

dveeden:

1. The `scripts/check-keywords.sh` script didn't check they "TiDBKeyword" section of "parser.y"
2. The entries in the docs from the "TiDBKeyword" section were labeled as reserved, while that doesn't seem to be the case.
3. Various missing keywords.

This shows the behavior for `TIDB` and `DESC` which are both listed as reserved keyword in the docs:

```
sql> CREATE TABLE TIDB(id INT);
Query OK, 0 rows affected (0.2196 sec)

sql> CREATE TABLE DESC(id INT);
ERROR: 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your TiDB version for the right syntax to use line 1 column 17 near "DESC(id INT)" 
```

I would expect both to fail, but it looks like `TIDB` (from the "TiDBKeyword" list) isn't actually reserved.

- [x] v7.5 (TiDB 7.5 versions)
- [ ] v7.4 (TiDB 7.4 versions)
- [ ] v7.3 (TiDB 7.3 versions)
- [x] v7.1 (TiDB 7.1 versions)

- This PR is translated from:
- Other reference link(s):

- https://github.com/pingcap/tidb/pull/48807
- https://github.com/pingcap/tidb/pull/48851
- 

- [ ] Delete files
- [ ] Change aliases


dveeden:
> @dveeden Could you help confirm if this PR also apply to release-7.5?

Yes it does.

mjonss (approved):

PR #16216, opened 2024-01-23
Revert "add multi value index case from cnf and dnf combination"

ran-huang:
Reverts pingcap/docs#15923


qiancai:
Removed the needs-cherry-pick-release-7.6 label because the v7.6 docs have been archived at https://docs-archive.pingcap.com/tidb/v7.6 and will no longer receive new updates.

qiancai:
Close this PR as it is not needed again according to @Arenatlx

PR #16271, opened 2024-01-24
tidbcloud: Update limits

dveeden:

The "Number of data replicas" could be understood as having a limit of 3 TiKV nodes. This isn't the case it is 3 replicas for every data region. 

And the Cluster Quotas are quite low. This should get a note that customers can raise a support ticket to increase these. (just like the note for Cluster Limits)

- [ ] v7.6 (TiDB 7.6 versions)
- [x] v7.5 (TiDB 7.5 versions)
- [ ] v7.4 (TiDB 7.4 versions)
- [ ] v7.3 (TiDB 7.3 versions)

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/tidb-cloud/limitations-and-quotas.md
+++ b/tidb-cloud/limitations-and-quotas.md
@@ -15,7 +15,7 @@ TiDB Cloud limits how many of each kind of component you can create in a [TiDB D
 
 | Component | Limit |
 |:-|:-|
-| Number of data replicas | 3 |
+| Number of copies for each [data region](/tidb-cloud/tidb-cloud-glossary.md#region) | 3 |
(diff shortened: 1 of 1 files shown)


mjonss (commented):
LGTM, but I strongly think we should change the name of 'region' to something else, like 'raft group' (which is the same!) or 'data range', 'range region', 'data stripe' or something else. By changing the name we will also reduce the name conflict with Cloud Region (as in AWS naming).

mjonss (approved):

sunxiaoguang (approved):
I also agree with the suggestions from @mjonss

qiancai (approved):

PR #18137, opened 2024-07-05
test vale

Oreoxmt:


dveeden:
@Oreoxmt 

This might be useful:
```
git config alias.vale '!vale $(git diff upstream/master --name-only)'
```

And then:
```
git vale
```

This runs vale (which needs to be installed) on the files that have changed between upstream/master and the branch you're on. Not sure if you have also named your upstream `upstream` or not...

PR #19067, opened 2024-10-10
Update timezone docs

dveeden:

- [ ] v8.2 (TiDB 8.2 versions)
- [ ] v8.0 (TiDB 8.0 versions)
- [ ] v7.6 (TiDB 7.6 versions)

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/configure-time-zone.md
+++ b/configure-time-zone.md
@@ -6,7 +6,7 @@ aliases: ['/docs/dev/configure-time-zone/','/docs/dev/how-to/configure/time-zone
 
 # Time Zone Support
 
-The time zone in TiDB is decided by the global [`time_zone`](/system-variables.md#time_zone) system variable and the session `time_zone` system variable. The default value of `time_zone` is `SYSTEM`. The actual time zone corresponding to `SYSTEM` is configured when the TiDB cluster bootstrap is initialized. The detailed logic is as follows:
+The time zone in TiDB is decided by the [`time_zone`](/system-variables.md#time_zone) system variable, which can be set at the session or global level. The default value of `time_zone` is `SYSTEM`. The actual time zone corresponding to `SYSTEM` is configured when the TiDB cluster bootstrap is initialized. The detailed logic is as follows:
 
 - Prioritize the use of the `TZ` environment variable.
 - If the `TZ` environment variable fails, extract the time zone from the actual soft link address of `/etc/localtime`.
@@ -18,7 +18,7 @@ You can use the following statement to set the global server `time_zone` value a
 SET GLOBAL time_zone = 'UTC';
 ```
 
-Examples values are named zones like `UTC`, `Africa/Nairobi`, `Asia/Jakarta` and UTC offsets like `+02:00` and `-05:00`.
+Valid values for `time_zone` include named zones such as `UTC`, `Africa/Nairobi`, and `Asia/Jakarta`, as well as UTC offsets like `+02:00` and `-05:00`.
 
 Each client has its own time zone setting, given by the session `time_zone` variable. Initially, the session variable takes its value from the global `time_zone` variable, but the client can change its own time zone with this statement:
 
@@ -38,7 +38,7 @@ To set the format of the value of the `time_zone`:
 - The value can be given as a string indicating an offset from UTC, such as '+10:00' or '-6:00'.
 - The value can be given as a named time zone, such as 'Europe/Helsinki', 'US/Eastern', or 'MET'.
 
-The current session time zone setting affects the display and interpretation of time values that are zone-sensitive. This includes the values displayed by functions such as [`NOW()`](/functions-and-operators/date-and-time-functions.md) or `CURTIME()`. The `UTC_TIMESTAMP()` function can be used to get a timestamp based on UTC, which can be useful to avoid time zone specific issues. The `CONVERT_TZ()` function can be used to convert between timezones.
+The current session time zone setting affects the display and interpretation of time values that are zone-sensitive, such as the values returned by [`NOW()`](/functions-and-operators/date-and-time-functions.md) and `CURTIME()` functions. To get a timestamp based on UTC, use the `UTC_TIMESTAMP()` function function, which helps avoid time zone-related issues. To convert between time zones, use the `CONVERT_TZ()` function.
 
 > **Note:**
 >
@@ -95,9 +95,9 @@ In this example, no matter how you adjust the value of the time zone, the value 
 >
 > - Time zone is involved during the conversion of the value of Timestamp and Datetime, which is handled based on the current `time_zone` of the session.
 > - For data migration, you need to pay special attention to the time zone setting of the primary database and the secondary database.
-> - To get correct timestamps it is strongly recommended to setup a reliable clock with NTP or PTP. See [system configuration check: NTP](/check-before-deployment.md#check-and-install-the-ntp-service) for how to check this.
-> - When using a timezone that observes daylight saving this might lead to ambigious timestamps and/or nonexistent timestamps, especially if calculations are done with the timestamp.
-> - MySQL uses `mysql_tzinfo_to_sql` to convert the timezone database of the OS to tables in the database. This isn't used or needed with TiDB as TiDB relies on timezone functions that directly read the data files from the OS timezone database that are part of the Go programming language.
+> - To get accurate timestamps, it is strongly recommended that you configure a reliable clock using Network Time Protocol (NTP) or Precision Time Protocol (PTP) services. For information about how to check NTP services, see [Check and install the NTP service](/check-before-deployment.md#check-and-install-the-ntp-service).
+> - Be aware that using time zones that observe daylight saving time can result in ambiguous or nonexistent timestamps, especially when performing calculations with those timestamps.
+> - MySQL uses `mysql_tzinfo_to_sql` to convert the time zone database of the operating system into tables in the `mysql` database. In contrast, TiDB directly reads the time zone data files from the time zone database of the operating system, which leverages the built-in time zone handling capabilities of the Go programming language.
 
 ## See also
 
--- a/configure-time-zone.md
+++ b/configure-time-zone.md
@@ -15,15 +15,15 @@ The time zone in TiDB is decided by the global [`time_zone`](/system-variables.m
 You can use the following statement to set the global server `time_zone` value at runtime:
 
 ```sql
-SET GLOBAL time_zone = timezone;
+SET GLOBAL time_zone = 'UTC';
 ```
 
-Examples values are named zones like `Africa/Nairobi` and `Asia/Jakarta` and UTC offsets like `+02:00` and `-05:00`.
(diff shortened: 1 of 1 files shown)


configure-time-zone.md
```diff
@@ -42,14 +38,12 @@ To set the format of the value of the `time_zone`:
 - The value can be given as a string indicating an offset from UTC, such as '+10:00' or '-6:00'.
 - The value can be given as a named time zone, such as 'Europe/Helsinki', 'US/Eastern', or 'MET'.
 
-The current session time zone setting affects the display and storage of time values that are zone-sensitive. This includes the values displayed by functions such as `NOW()` or `CURTIME()`.
+The current session time zone setting affects the display and storage of time values that are zone-sensitive. This includes the values displayed by functions such as [`NOW()`](/functions-and-operators/date-and-time-functions.md) or `CURTIME()`. The `UTC_TIMESTAMP()` function can be used to get a timestamp based on UTC, which can be useful to avoid time zone specific issues. The `CONVERT_TZ()` function can be used to convert between timezones.
```
mjonss, line 41: I would not say that it affects the "storage" of time values, but rather the interpretation and display of TIMESTAMP column type values, as well as some of the mentioned time and date functions. DATETIME/TIME stores the time as given (no conversion) and TIMESTAMP stores the time in UTC (conversion before storing and when retrieving).

configure-time-zone.md
```diff
@@ -109,3 +95,11 @@ In this example, no matter how you adjust the value of the time zone, the value
 >
 > - Time zone is involved during the conversion of the value of Timestamp and Datetime, which is handled based on the current `time_zone` of the session.
 > - For data migration, you need to pay special attention to the time zone setting of the primary database and the secondary database.
+> - To get correct timestamps it is strongly recommended to setup a reliable clock with NTP or PTP. See [system configuration check: NTP](/check-before-deployment.md#check-and-install-the-ntp-service) for how to check this.
+> - When using a timezone that observes daylight saving this might lead to ambigious timestamps and/or nonexistent timestamps, especially if calculations are done with the timestamp.
+> - MySQL uses `mysql_tzinfo_to_sql` to convert the timezone database of the OS to tables in the database. This isn't used or needed with TiDB as TiDB relies on timezone data that's part of the Go programming language.
```
mjonss, line 100: Do we need to also say that depending on the environment it will use the time zone info files from the OS, as golang's time package says "use the file <time_zone> in the system timezone directory"?

dveeden: This is quite a big advantage. Where [`CONVERT_TZ`](https://dev.mysql.com/doc/refman/8.4/en/date-and-time-functions.html#function_convert-tz) in MySQL might return NULL because of missing timezone data that never happens with TiDB.

In addition to this the data in TiDB doesn't need to get updated when the [Olson database](https://en.wikipedia.org/wiki/Tz_database) gets updated. Note that MySQL doesn't record a version or timestamp, which makes it impossible to see if the data is outdated and needs to be loaded again. See also https://bugs.mysql.com/bug.php?id=60675

dveeden:
@mjonss should we mention weird things like the `right/` zones (e.g. `right/Universal`)  and `posix/` (e.g. `posix/Chile/EasterIsland`) ?

mjonss:
> @mjonss should we mention weird things like the `right/` zones (e.g. `right/Universal`) and `posix/` (e.g. `posix/Chile/EasterIsland`) ? Or should we remove support for them maybe?
> ...

I think we don't need to mention it, since it is platform/OS/installation dependent, I think we can ignore it and just keep what golang implements. As long as we are MySQL compatible too :)

configure-time-zone.md
```diff
@@ -42,14 +38,12 @@ To set the format of the value of the `time_zone`:
 - The value can be given as a string indicating an offset from UTC, such as '+10:00' or '-6:00'.
 - The value can be given as a named time zone, such as 'Europe/Helsinki', 'US/Eastern', or 'MET'.
```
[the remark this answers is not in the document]
dveeden: Not sure, I think keeping this might be best.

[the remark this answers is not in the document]
qiancai: I took a look at the doc again. Yes, I agree, keeping it makes more sense. Btw, the current structure of this doc might be challenging for users. Would it be helpful to introduce the time_zone format first, followed by guidance on using it to set the time zone? If it is OK with you, I’d be glad to submit a commit in this PR for this update.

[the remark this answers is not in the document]
dveeden: Yes that sounds good. Please send the commit.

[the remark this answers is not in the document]
qiancai: Here it is: [`702e7e8` (#19067)](https://github.com/pingcap/docs/pull/19067/commits/702e7e808f6c11252fabbee3fb55f5d988d978b8). Could you please take a look at it?

qiancai (approved):

configure-time-zone.md
```diff
@@ -6,49 +6,61 @@ aliases: ['/docs/dev/configure-time-zone/','/docs/dev/how-to/configure/time-zone
 
 # Time Zone Support
 
-The time zone in TiDB is decided by the global `time_zone` system variable and the session `time_zone` system variable. The default value of `time_zone` is `SYSTEM`. The actual time zone corresponding to `System` is configured when the TiDB cluster bootstrap is initialized. The detailed logic is as follows:
+The time zone in TiDB is decided by the [`time_zone`](/system-variables.md#time_zone) system variable, which can be set at the session or global level. The default value of `time_zone` is `SYSTEM`. The actual time zone corresponding to `SYSTEM` is configured when the TiDB cluster bootstrap is initialized. The detailed logic is as follows:
```
hfxsd, line 9:
```suggestion
The time zone in TiDB is decided by the [`time_zone`](/system-variables.md#time_zone) system variable. You can set it at the session or global level. The default value of `time_zone` is `SYSTEM`. The actual time zone corresponding to `SYSTEM` is configured when the TiDB cluster bootstrap is initialized. The detailed logic is as follows:
```

configure-time-zone.md
```diff
@@ -6,49 +6,61 @@ aliases: ['/docs/dev/configure-time-zone/','/docs/dev/how-to/configure/time-zone
...
 
-The current session time zone setting affects the display and storage of time values that are zone-sensitive. This includes the values displayed by functions such as `NOW()` or `CURTIME()`.
+    ```sql
+    SET time_zone = 'US/Pacific';
+    ```
 
-> **Note:**
->
-> Only the values of the Timestamp data type is affected by time zone. This is because the Timestamp data type uses the literal value + time zone information. Other data types, such as Datetime/Date/Time, do not have time zone information, thus their values are not affected by the changes of time zone.
+## Functions and data types affected by time zone setting
```
hfxsd, line 57:
```suggestion
## Functions and data types affected by time zone settings
```

configure-time-zone.md
```diff
@@ -103,9 +107,17 @@ select * from t;
...
+
+## Important considerations for time zone settings
+
+- Time zone is involved during the conversion of the values of `TIMESTAMP` and `DATETIME`, which is handled based on the `time_zone` of the current session.
+- For data migration, you need to pay special attention to the time zone setting of the primary database and the secondary database.
+- To get accurate timestamps, it is strongly recommended that you configure a reliable clock using Network Time Protocol (NTP) or Precision Time Protocol (PTP) services. For information about how to check NTP services, see [Check and install the NTP service](/check-before-deployment.md#check-and-install-the-ntp-service).
+- Be aware that using time zones that observe daylight saving time can result in ambiguous or nonexistent timestamps, especially when performing calculations with those timestamps.
+- MySQL uses `mysql_tzinfo_to_sql` to convert the time zone database of the operating system into tables in the `mysql` database. In contrast, TiDB directly reads the time zone data files from the time zone database of the operating system, which leverages the built-in time zone handling capabilities of the Go programming language.
```
hfxsd, line 118:
```suggestion
- MySQL uses [`mysql_tzinfo_to_sql`](https://dev.mysql.com/doc/refman/8.4/en/mysql-tzinfo-to-sql.html) to convert the time zone database of the operating system into tables in the `mysql` database. In contrast, TiDB directly reads the time zone data files from the time zone database of the operating system, which leverages the built-in time zone handling capabilities of the Go programming language.
```

configure-time-zone.md
```diff
@@ -6,49 +6,61 @@ aliases: ['/docs/dev/configure-time-zone/','/docs/dev/how-to/configure/time-zone
...
 
-The time zone in TiDB is decided by the global `time_zone` system variable and the session `time_zone` system variable. The default value of `time_zone` is `SYSTEM`. The actual time zone corresponding to `System` is configured when the TiDB cluster bootstrap is initialized. The detailed logic is as follows:
+The time zone in TiDB is decided by the [`time_zone`](/system-variables.md#time_zone) system variable, which can be set at the session or global level. The default value of `time_zone` is `SYSTEM`. The actual time zone corresponding to `SYSTEM` is configured when the TiDB cluster bootstrap is initialized. The detailed logic is as follows:
 
-- Prioritize the use of the `TZ` environment variable.
-- If the `TZ` environment variable fails, extract the time zone from the actual soft link address of `/etc/localtime`.
-- If both of the above methods fail, use `UTC` as the system time zone.
+- TiDB prioritizes the use of the `TZ` environment variable.
+- If the `TZ` environment variable fails, TiDB reads the time zone from the soft link at `/etc/localtime`.
+- If both of the preceding methods fail, TiDB uses `UTC` as the system time zone.
```
hfxsd, line 13:
```suggestion
1. TiDB prioritizes the use of the `TZ` environment variable.
2. If the `TZ` environment variable fails, TiDB reads the time zone from the soft link at `/etc/localtime`.
3. If both of the preceding methods fail, TiDB uses `UTC` as the system time zone.
```

hfxsd (approved):

configure-time-zone.md
```diff
@@ -103,9 +107,17 @@ select * from t;
 1 row in set (0.00 sec)
 ```
 
-In this example, no matter how you adjust the value of the time zone, the value of the Datetime data type is not affected. But the displayed value of the Timestamp data type changes if the time zone information changes. In fact, the value that is stored in the storage does not change, it's just displayed differently according to different time zone setting.
+In this example, no matter how you adjust the time zone value, the value of the `DATETIME` data type is not affected. However, the displayed value of the `TIMESTAMP` data type reflects changes in the time zone. In fact, the `TIMESTAMP` value stored in the database remains unchanged, but it is displayed differently according to different time zone settings.
+
+## Important considerations for time zone settings
+
+- Time zone is involved during the conversion of the values of `TIMESTAMP` and `DATETIME`, which is handled based on the `time_zone` of the current session.
+- For data migration, you need to pay special attention to the time zone setting of the primary database and the secondary database.
```
qiancai, line 115:
```suggestion
- For data migration, you need to pay special attention to whether the time zone settings of the primary database and the secondary database are consistent.
```

PR #19391, opened 2024-11-13
system tables: add a note for DATA_LENGTH, INDEX_LENGTH, and TABLE_SIZE

qiancai:

 Add a note for DATA_LENGTH, INDEX_LENGTH, and TABLE_SIZE to explain they are logical estimates instead of actual physical sizes.

- [ ] v8.4 (TiDB 8.4 versions)
- [ ] v8.3 (TiDB 8.3 versions)
- [ ] v8.2 (TiDB 8.2 versions)

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/information-schema/information-schema-table-storage-stats.md
+++ b/information-schema/information-schema-table-storage-stats.md
@@ -59,5 +59,5 @@ Fields in the `TABLE_STORAGE_STATS` table are described as follows:
 * `PEER_COUNT`: The number of replicas of the table.
 * `REGION_COUNT`: The number of Regions.
 * `EMPTY_REGION_COUNT`: The number of Regions that do not contain data in this table.
-* `TABLE_SIZE`: The total size of the table, in the unit of MiB.
+* `TABLE_SIZE`: The total size of the table, in the unit of MiB. Note that this value is a logical estimate based on table statistics and does not represent the actual compressed physical size in storage.
 * `TABLE_KEYS`: The total number of records in the table.
--- a/information-schema/information-schema-tables.md
+++ b/information-schema/information-schema-tables.md
@@ -102,9 +102,9 @@ The description of columns in the `TABLES` table is as follows:
 * `ROW_FORMAT`: The row format. The value is currently `Compact`.
 * `TABLE_ROWS`: The number of rows in the table in statistics.
 * `AVG_ROW_LENGTH`: The average row length of the table. `AVG_ROW_LENGTH` = `DATA_LENGTH` / `TABLE_ROWS`.
-* `DATA_LENGTH`: Data length. `DATA_LENGTH` = `TABLE_ROWS` \* the sum of storage lengths of the columns in the tuple. The replicas of TiKV are not taken into account.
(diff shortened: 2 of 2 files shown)


information-schema/information-schema-table-storage-stats.md
```diff
@@ -59,5 +59,5 @@ Fields in the `TABLE_STORAGE_STATS` table are described as follows:
 * `PEER_COUNT`: The number of replicas of the table.
 * `REGION_COUNT`: The number of Regions.
 * `EMPTY_REGION_COUNT`: The number of Regions that do not contain data in this table.
-* `TABLE_SIZE`: The total size of the table, in the unit of MiB.
+* `TABLE_SIZE`: The total size of the table, in the unit of MiB. Note that this value is a logical estimate based on table statistics and does not represent the actual compressed physical size in storage.
```
dveeden, line 62: The fact that this is based on statistics is good. However that it doesn't represent the "actual compressed physical size in storage" is not because the statistics aren't precise enough. This is because the size in storage and is measuring something different.

dveeden (commented):
I think this should:

1. Link to the FAQ about storage size
2. Explain the storage size (possibly on another page): 
    2a. The size of the table as it is currently. 
    2b. The size of the table including deleted/updated rows. (before they are removed by GC). Maybe link to `garbage-collection-overview.md`.
    2c. The size on disk of various levels in RocksDB, including compression. Maybe link to `compression-per-level`.
3. Link from I_S.tables to `TABLE_STORAGE_STATS`
4. Explain how TiFlash storage size affects this
5. Explain how PD's `max-replicas` causes up to 3 copies of the data and how this influences I_S.tables.DATA_LENGTH
6. As storage efficiency is a major advantage of our product, maybe list this on `overview.md`
7. Maybe explain why DATA_FREE is 0 for TiKV. For MySQL/InnoDB tablespaces never shrink and one needs to rebuild to freeup space. This is also a advantage of our product.

dveeden:
@mjonss I assume that partitioning and global index work as expected when it comes to the storage size explained here?

dveeden:
`TIKV_REGION_STATUS` seems to depend on Prometheus being available. However this isn't noted on https://docs.pingcap.com/tidb/stable/information-schema-tikv-region-status 

The result is that size calculation would fail with this error:
```
ERROR 1105 (HY000): query metric error: [domain:9009]Prometheus address is not set in PD and etcd
```

If a custom Prometheus is used then the address might not be configured correctly in pd/etcd. If there is no Prometheus we may want to see if there are any alternatives available

AilinKid (approved):

information-schema/information-schema-table-storage-stats.md
mjonss: So it represent the statistical size of the uncompressed data? With or without possible changed rows or tombstones?

qiancai:
Removed the `needs-cherry-pick-release-8.4` label because the v8.4 docs have been archived at https://docs-archive.pingcap.com/tidb/v8.4 and will no longer receive new updates.

information-schema/information-schema-table-storage-stats.md
mjonss: Since it is based on column statistics, it never has size info from Vector data types, and by default not of JSON/BLOB/TEXT, and fixed size columns, like INT, are not taking NULLs into account, see [this comment](https://github.com/pingcap/tidb/pull/69955#issuecomment-5199256882).

PR #20734, opened 2025-04-10
docs: clarify AUTO_RANDOM_BASE usage and explicit insert behavior

### LINKED ISSUE #20732: need to clarify the example in AUTO_RANDOM
## Error Report

Feedback in Chinese:

https://docs.pingcap.com/tidb/stable/auto-random/#user-scenario 文档中的代码示例存在误导，执行ALTER TABLE t AUTO_RANDOM_BASE=0;后确实会输出Can't reset AUTO_INCREMENT to 0 without FORCE option, using xxxx instead，但这并没有起效。需要使用force关键字执行设置，但直觉会让我参考ALTER TABLE t AUTO_RANDOM_BASE=0;就会去执行ALTER TABLE t force AUTO_RANDOM_BASE=0;又会报错不能将下一个ID设置为0。探索后发现需要设置成非0数字。

English translation:

In this doc, https://docs.pingcap.com/tidb/stable/auto-random/#user-scenario, the code example is misleading. After executing `ALTER TABLE t AUTO_RANDOM_BASE=0;`, it does output `Can't reset AUTO_INCREMENT to 0 without FORCE option, using xxxx instead`, but the setting does not actually take effect. You need to use the `FORCE` keyword to apply the change. However, intuitively, referencing `ALTER TABLE t AUTO_RANDOM_BASE=0;` might lead one to try `ALTER TABLE t FORCE AUTO_RANDOM_BASE=0;`, which results in an error saying that the next ID cannot be set to 0. After some exploration, I found that it must be set to a non-zero number.

We need to clarify that in the doc.

alastori:

Close issue: #20732

This PR updates the "Clear the auto-increment ID cache" section in `auto-random.md` to address confusion reported in the issue above. Specifically, the changes:

- Clarify that `ALTER TABLE ... AUTO_RANDOM_BASE` without the `FORCE` keyword does not actually change the value, despite the warning message.
- Emphasize that the `FORCE` keyword is mandatory to modify `AUTO_RANDOM_BASE`.
- Explicitly state that attempting to set `AUTO_RANDOM_BASE` to `0`, even with `FORCE`, will result in an error.
- Clarify that a non-zero positive integer must be used with `FORCE`.
- Improve the explanation of how explicit inserts can potentially collide with the internal auto-incrementing part of `AUTO_RANDOM` IDs, potentially leading to duplicate key errors.
- Split a longer paragraph into smaller ones for better readability.

- [ ] v9.0 (TiDB 9.0 versions)
- [ ] v8.4 (TiDB 8.4 versions)
- [ ] v8.3 (TiDB 8.3 versions)

- Closes #20732
- File modified: `auto-random.md`

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/auto-random.md
+++ b/auto-random.md
@@ -168,7 +168,7 @@ TiDB implicitly allocates values to `AUTO_RANDOM` columns similarly to `AUTO_INC
 
 When you insert data with explicit values into an `AUTO_RANDOM` column in a deployment with multiple TiDB server instances, potential ID collisions can occur, similar to an `AUTO_INCREMENT` column. If explicit inserts happen to use ID values that conflict with the internal counter TiDB uses for automatic generation, this can lead to errors. 
 
-Here's how the collision can happen: each `AUTO_RANDOM` ID contains an auto-incrementing part alongside random bits. TiDB uses an internal counter for this auto-incrementing part. If you explicitly insert an ID where this part matches the counter's next value, TiDB might later try to generate the same ID automatically, leading to a duplicate key error. See [AUTO_INCREMENT Uniqueness](/auto-increment.md#uniqueness) for details.
+Here's how the collision can happen: each `AUTO_RANDOM` ID consists of random bits and an auto-incrementing part. TiDB uses an internal counter for this auto-incrementing part. If you explicitly insert an ID where the auto-incrementing part matches the counter's next value, a duplicate key error might occur when TiDB later attempts to generate the same ID automatically. For more details, see [AUTO_INCREMENT Uniqueness](/auto-increment.md#uniqueness).
 
 With a single TiDB instance, this issue doesn't occur because the node automatically adjusts its internal counter when processing explicit insertions, preventing any future collisions. In contrast, with multiple TiDB nodes, each node maintains its own cache of IDs, which needs to be cleared to prevent collisions after explicit insertions. To clear these unallocated cached IDs and avoid potential collisions, you have two options: 
 
@@ -178,7 +178,7 @@ With a single TiDB instance, this issue doesn't occur because the node automatic
 ALTER TABLE t AUTO_RANDOM_BASE=0;
 ```
 
-This command will automatically determine an appropriate base value. Although it produces a warning message like `Can't reset AUTO_INCREMENT to 0 without FORCE option, using XXX instead`, the value **will** change and you can safely ignore this warning.
+This statement automatically determines an appropriate base value. Although it produces a warning message similar to `Can't reset AUTO_INCREMENT to 0 without FORCE option, using XXX instead`, the base value **will** change and you can safely ignore this warning.
 
 > **Note:**
 >
@@ -192,11 +192,11 @@ If you need to set a specific base value (for example, `1000`), use the `FORCE` 
 ALTER TABLE t FORCE AUTO_RANDOM_BASE = 1000;
 ```
 
-This approach is less convenient as it requires you to determine an appropriate value yourself.
+This approach is less convenient because it requires you to determine an appropriate base value yourself.
 
 > **Note:**
 >
-> * You must use a non-zero positive integer value when using `FORCE`.
+> * When using `FORCE`, you must specify a non-zero positive integer.
 
 Both commands modify the starting point for the auto-increment bits used in subsequent `AUTO_RANDOM` value generations across all TiDB nodes. They do not affect already allocated IDs.
 
--- a/auto-random.md
+++ b/auto-random.md
@@ -166,30 +166,31 @@ TiDB implicitly allocates values to `AUTO_RANDOM` columns similarly to `AUTO_INC
 
 ## Clear the auto-increment ID cache
 
-Explicitly inserting data into an `AUTO_RANDOM` column behaves the same as with an `AUTO_INCREMENT` column, so you also need to clear the auto-increment ID cache. For more details, see [Clear the auto-increment ID cache](/auto-increment.md#clear-the-auto-increment-id-cache).
+When you insert data with explicit values into an `AUTO_RANDOM` column, it works similarly to an `AUTO_INCREMENT` column regarding potential ID collisions. If explicit inserts happen to use ID values that conflict with the internal counter TiDB uses for automatic generation, this can lead to errors.
 
-You can run the `ALTER TABLE` statement to set `AUTO_RANDOM_BASE=0` to clear the auto-increment ID cache on all TiDB nodes in the cluster. For example:
+Here's how the collision can happen: each `AUTO_RANDOM` ID contains an auto-incrementing part alongside random bits. TiDB uses an internal counter for this auto-incrementing part. If you explicitly insert an ID where this part matches the counter's next value, TiDB might later try to generate the same ID automatically, leading to a duplicate key error.
+
+Although `AUTO_RANDOM` doesn't keep track of a single specific "next ID" like `AUTO_INCREMENT` does, frequent explicit inserts that cause such conflicts might still require you to adjust the starting point (`AUTO_RANDOM_BASE`) for the auto-incrementing part of future automatically generated IDs to avoid these potential errors.
+
+To change the base value (`AUTO_RANDOM_BASE`) used for the auto-increment part of future ID generations on an existing table, you must use the `FORCE` keyword.
+
+> **Note:**
+>
+> * If you try to alter `AUTO_RANDOM_BASE` without the `FORCE` keyword, the value will not change. The command completes, but the base value stays unchanged. A subsequent `SHOW WARNINGS` reveals the specific warning `Can't reset AUTO_INCREMENT to 0 without FORCE option, using 1 instead`.
+> * You cannot set `AUTO_RANDOM_BASE` to `0`, even with the `FORCE` keyword. Attempting this results in an error `Cannot force rebase the next global ID to '0'`.
+> * You must use a non-zero positive integer value when using `FORCE`.
+
+To set a new base value (for example, `1000`) for the auto-increment part of future implicitly generated IDs for table `t`, use the following statement:
 
 ```sql
-ALTER TABLE t AUTO_RANDOM_BASE=0;
+ALTER TABLE t FORCE AUTO_RANDOM_BASE = 1000;
 ```
 
 ```
-Query OK, 0 rows affected, 1 warning (0.52 sec)
+Query OK, 0 rows affected (0.XX sec)
 ```
 
-```sql
-SHOW WARNINGS;
-```
-
-```
-+---------+------+-------------------------------------------------------------------------+
-| Level   | Code | Message                                                                 |
-+---------+------+-------------------------------------------------------------------------+
-| Warning | 1105 | Can't reset AUTO_INCREMENT to 0 without FORCE option, using 101 instead |
-+---------+------+-------------------------------------------------------------------------+
-1 row in set (0.00 sec)
-```
+This command modifies the starting point for the auto-increment bits used in subsequent `AUTO_RANDOM` value generations across all TiDB nodes. It does not affect already allocated IDs.
 
 ## Restrictions
 
--- a/auto-random.md
+++ b/auto-random.md
@@ -166,31 +166,39 @@ TiDB implicitly allocates values to `AUTO_RANDOM` columns similarly to `AUTO_INC
 
 ## Clear the auto-increment ID cache
 
-When you insert data with explicit values into an `AUTO_RANDOM` column in a deployment with multiple TiDB server instances, potential ID collisions can occur, similar to an `AUTO_INCREMENT` column. If explicit inserts happen to use ID values that conflict with the internal counter TiDB uses for automatic generation, this can lead to errors.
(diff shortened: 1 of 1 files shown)


alastori:
@frank945946

alastori:
To confirm the behavior described in the documentation updates, I ran the following tests on a local TiDB instance:

**1. Create Table:**
```sql
CREATE TABLE t (a BIGINT PRIMARY KEY AUTO_RANDOM, b VARCHAR(255));
```
```
Query OK, 0 rows affected, 1 warning (0.07 sec)
```

**2. Attempt `AUTO_RANDOM_BASE=0` without `FORCE`:**
```sql
ALTER TABLE t AUTO_RANDOM_BASE=0;
SHOW WARNINGS;
```
```
Query OK, 0 rows affected, 1 warning (0.07 sec)
+---------+------+-----------------------------------------------------------------------+
| Level   | Code | Message                                                               |
+---------+------+-----------------------------------------------------------------------+
| Warning | 1105 | Can't reset AUTO_INCREMENT to 0 without FORCE option, using 1 instead |
+---------+------+-----------------------------------------------------------------------+
1 row in set (0.01 sec)
```
*(Confirms the command completes with a warning but doesn't reset to 0.)*

**3. Attempt `FORCE AUTO_RANDOM_BASE=0`:**
```sql
ALTER TABLE t FORCE AUTO_RANDOM_BASE=0;
```
```
ERROR 1467 (HY000): Cannot force rebase the next global ID to '0'
```
*(Confirms forcing the base to 0 results in an error.)*

**4. Attempt `FORCE AUTO_RANDOM_BASE = 1000` (non-zero):**
```sql
ALTER TABLE t FORCE AUTO_RANDOM_BASE = 1000;
```
```
Query OK, 0 rows affected (0.04 sec)
```
*(Confirms forcing the base to a non-zero value succeeds.)*

**5. Insert data after forcing base to 1000:**
```sql
INSERT INTO t (b) VALUES ('test1'), ('test2');
SELECT * FROM t WHERE b LIKE 'test%';
```
```
Query OK, 2 rows affected (0.01 sec)
Records: 2  Duplicates: 0  Warnings: 0

+---------------------+-------+
| a                   | b     |
+---------------------+-------+
| 4899916394579100648 | test1 | -- Auto-increment part is 1000
| 4899916394579100649 | test2 | -- Auto-increment part is 1001
+---------------------+-------+
2 rows in set (0.00 sec)
```
*(Confirms the auto-increment part starts from the forced base value.)*

These results align with the behavior described in the updated documentation.

auto-random.md
```diff
@@ -166,31 +166,32 @@ TiDB implicitly allocates values to `AUTO_RANDOM` columns similarly to `AUTO_INC
 
 ## Clear the auto-increment ID cache
 
-Explicitly inserting data into an `AUTO_RANDOM` column behaves the same as with an `AUTO_INCREMENT` column, so you also need to clear the auto-increment ID cache. For more details, see [Clear the auto-increment ID cache](/auto-increment.md#clear-the-auto-increment-id-cache).
+When you insert data with explicit values into an `AUTO_RANDOM` column, it works similarly to an `AUTO_INCREMENT` column regarding potential ID collisions. If explicit inserts happen to use ID values that conflict with the internal counter TiDB uses for automatic generation, this can lead to errors.
 
-You can run the `ALTER TABLE` statement to set `AUTO_RANDOM_BASE=0` to clear the auto-increment ID cache on all TiDB nodes in the cluster. For example:
+Here's how the collision can happen: each `AUTO_RANDOM` ID contains an auto-incrementing part alongside random bits. TiDB uses an internal counter for this auto-incrementing part. If you explicitly insert an ID where this part matches the counter's next value, TiDB might later try to generate the same ID automatically, leading to a duplicate key error.
```
tangenta, line 171: If there is only one TiDB, the collision will not happen because it will rebase for the explicit insertion.

See https://docs.pingcap.com/tidb/stable/auto-increment/#uniqueness

auto-random.md
```diff
@@ -166,31 +166,32 @@ TiDB implicitly allocates values to `AUTO_RANDOM` columns similarly to `AUTO_INC
...
 
-```sql
-ALTER TABLE t AUTO_RANDOM_BASE=0;
-```
+Although `AUTO_RANDOM` doesn't keep track of a single specific "next ID" like `AUTO_INCREMENT` does, frequent explicit inserts that cause such conflicts might still require you to adjust the starting point (`AUTO_RANDOM_BASE`) for the auto-incrementing part of future automatically generated IDs to avoid these potential errors.
 
-```
-Query OK, 0 rows affected, 1 warning (0.52 sec)
-```
+To change the base value (`AUTO_RANDOM_BASE`) used for the auto-increment part of future ID generations on an existing table, you must use the `FORCE` keyword.
```
tangenta, line 175: The purpose of this part ("Clear the auto-increment ID cache") is not about setting `AUTO_RANDOM_BASE` to a specific value. The purpose is to clear the unallocated cached IDs to avoid potential collision. There are two ways:
- Set `AUTO_RANDOM_BASE` to a value with `FORCE` keyword (I don't recommend, because it requires user to get the minimum ID value from the table, which is cumbersome).
- Rebase `AUTO_RANDOM_BASE` automatically (`ALTER TABLE t AUTO_RANDOM_BASE=0;`).

auto-random.md
```diff
@@ -166,31 +166,32 @@ TiDB implicitly allocates values to `AUTO_RANDOM` columns similarly to `AUTO_INC
...
+
+> **Note:**
+>
+> * If you try to alter `AUTO_RANDOM_BASE` without the `FORCE` keyword, the value will not change. The command completes, but the base value stays unchanged. A subsequent `SHOW WARNINGS` reveals the specific warning `Can't reset AUTO_INCREMENT to 0 without FORCE option, using 1 instead`.
```
tangenta, line 179: > If you try to alter `AUTO_RANDOM_BASE` without the `FORCE` keyword,

the value WILL ~not~ change even if the warning is reported. We should suggest ignoring the warning instead of using `FORCE` (see my comment above)

auto-random.md
alastori: Thanks for pointing that out. I've mentioned "in a deployment with multiple TiDB server instances" and also added a clear explanation about how single TiDB instances handle this differently: "With a single TiDB instance, this issue doesn't occur because the node automatically adjusts its internal counter when processing explicit insertions, preventing any future collisions."
I've also incorporated the link to the uniqueness documentation you shared. This should help users understand when they need to worry about this issue.

auto-random.md
alastori: Thank you for clarifying the purpose of this section! I've completely restructured it to focus on clearing unallocated cached IDs rather than just setting a specific value. I've presented both approaches with the automatic rebasing method as the recommended option, and explained the manual approach as an alternative. The documentation now clearly states the purpose is to avoid potential collisions in multi-node deployments.

auto-random.md
alastori: @tangenta I appreciate this critical correction. I've updated the documentation to clearly state that when using `AUTO_RANDOM_BASE=0` without the `FORCE` keyword, "the value will change and you can safely ignore this warning." This is an important distinction that users need to know.
The warning message itself (`Can't reset AUTO_INCREMENT to 0 without FORCE option, using XXX instead`) is a bit misleading since it mentions `AUTO_INCREMENT` instead of `AUTO_RANDOM` and suggests the change isn't happening when it actually is. Would updating this warning message be something to consider for a future release?

alastori: Suggestion just in case you are considering updating the warning message: "AUTO_RANDOM_BASE has been updated to XXX. You specified '0' which triggered automatic rebasing to prevent ID collisions."

tangenta: https://github.com/pingcap/tidb/pull/60616

auto-random.md
```diff
@@ -166,30 +166,39 @@ TiDB implicitly allocates values to `AUTO_RANDOM` columns similarly to `AUTO_INC
...
 
-Explicitly inserting data into an `AUTO_RANDOM` column behaves the same as with an `AUTO_INCREMENT` column, so you also need to clear the auto-increment ID cache. For more details, see [Clear the auto-increment ID cache](/auto-increment.md#clear-the-auto-increment-id-cache).
+When you insert explicit values into an `AUTO_RANDOM` column in a multi-instance TiDB deployment, ID collisions might occur, similar to an `AUTO_INCREMENT` column. Errors occur when the explicitly inserted ID conflicts with the internal counter that TiDB uses for automatic ID generation. 
 
-You can run the `ALTER TABLE` statement to set `AUTO_RANDOM_BASE=0` to clear the auto-increment ID cache on all TiDB nodes in the cluster. For example:
+The collision happens as follows: each `AUTO_RANDOM` ID contains an auto-incrementing part alongside random bits. TiDB uses an internal counter for this auto-incrementing part. If you explicitly insert an ID where the auto-incrementing part matches the counter's next value, a duplicate key error might occur when TiDB later attempts to generate the same ID automatically. For more details, see [AUTO_INCREMENT Uniqueness](/auto-increment.md#uniqueness).
```
qiancai, line 171:
```suggestion
The collision happens as follows: each `AUTO_RANDOM` ID consists of random bits and an auto-incrementing part. TiDB uses an internal counter for this auto-incrementing part. If you explicitly insert an ID where the auto-incrementing part matches the counter's next value, a duplicate key error might occur when TiDB later attempts to generate the same ID automatically. For more details, see [AUTO_INCREMENT Uniqueness](/auto-increment.md#uniqueness).
```

auto-random.md
```diff
@@ -166,30 +166,39 @@ TiDB implicitly allocates values to `AUTO_RANDOM` columns similarly to `AUTO_INC
...
+### Option 1: Automatically rebase (Recommended)
 
 ```sql
 ALTER TABLE t AUTO_RANDOM_BASE=0;
 ```
 
-```
-Query OK, 0 rows affected, 1 warning (0.52 sec)
-```
+This statement automatically determines an appropriate base value. Although it produces a warning message similar to `Can't reset AUTO_INCREMENT to 0 without FORCE option, using XXX instead`, the value **will** change and you can safely ignore this warning.
```
qiancai, line 181:
```suggestion
This statement automatically determines an appropriate base value. Although it produces a warning message similar to `Can't reset AUTO_INCREMENT to 0 without FORCE option, using XXX instead`, the base value **will** change and you can safely ignore this warning.
```

auto-random.md
```diff
@@ -166,30 +166,39 @@ TiDB implicitly allocates values to `AUTO_RANDOM` columns similarly to `AUTO_INC
...
 
-```
-+---------+------+-------------------------------------------------------------------------+
-| Level   | Code | Message                                                                 |
-+---------+------+-------------------------------------------------------------------------+
-| Warning | 1105 | Can't reset AUTO_INCREMENT to 0 without FORCE option, using 101 instead |
-+---------+------+-------------------------------------------------------------------------+
-1 row in set (0.00 sec)
-```
+This approach is less convenient because it requires you to determine an appropriate value yourself.
```
qiancai, line 195:
```suggestion
This approach is less convenient because it requires you to determine an appropriate base value yourself.
```

auto-random.md
```diff
@@ -166,30 +166,39 @@ TiDB implicitly allocates values to `AUTO_RANDOM` columns similarly to `AUTO_INC
...
+
+> **Note:**
+>
+> You must use a non-zero positive integer value when using `FORCE`.
```
qiancai, line 199:
```suggestion
> When using `FORCE`, you must specify a non-zero positive integer.

**Review by qiancai — APPROVED** [review]
Rest LGTM

File: auto-random.md (hunk 4 above)
Comment (alastori) [reply auto-random.md]: Thanks! Applied this suggestion in commit [89585ab4](https://github.com/pingcap/docs/commit/89585ab43).

File: auto-random.md (hunk 6 above)
Comment (alastori) [reply auto-random.md]: Thanks! Applied this suggestion in commit [89585ab4](https://github.com/pingcap/docs/commit/89585ab43).

File: auto-random.md (hunk 7 above)
Comment (alastori) [reply auto-random.md]: Thanks! Applied this suggestion in commit [89585ab4](https://github.com/pingcap/docs/commit/89585ab43).

Comment (alastori) [reply auto-random.md]: Thanks! Applied this suggestion in commit [89585ab4](https://github.com/pingcap/docs/commit/89585ab43).

File: auto-random.md [hunk 8]
```diff
@@ -166,30 +166,39 @@ TiDB implicitly allocates values to `AUTO_RANDOM` columns similarly to `AUTO_INC
...
-+---------+------+-------------------------------------------------------------------------+
-| Warning | 1105 | Can't reset AUTO_INCREMENT to 0 without FORCE option, using 101 instead |
-+---------+------+-------------------------------------------------------------------------+
-1 row in set (0.00 sec)
-```
+This approach is less convenient because it requires you to determine an appropriate base value yourself.
+
+> **Note:**
+>
+> * When using `FORCE`, you must specify a non-zero positive integer.
```
hfxsd, line 199:
```suggestion
> When using `FORCE`, you must specify a non-zero positive integer.
```

hfxsd (approved):

PR #21843, opened 2025-09-26
enable-tls-between-clients-and-servers: what variable to check

dveeden:

- [x] I've signed the [**Contributor License Agreement**](https://cla.pingcap.net/pingcap/docs), which is required for the repository owners to accept my contribution.

- [x] v9.0 (TiDB 9.0 versions)

- This PR is translated from:
- Other reference link(s):

- [ ] Delete files
- [ ] Change aliases

### DIFF
--- a/enable-tls-between-clients-and-servers.md
+++ b/enable-tls-between-clients-and-servers.md
@@ -121,6 +121,8 @@ SHOW STATUS LIKE "Ssl%";
 6 rows in set (0.0062 sec)
 ```
 
+The connection is secure when the value of `Ssl_cipher` is not empty.
+
 For the official MySQL client, you can also use the `STATUS` or `\s` statement to view the connection status:
 
 ```
--- a/enable-tls-between-clients-and-servers.md
+++ b/enable-tls-between-clients-and-servers.md
@@ -121,7 +121,7 @@ SHOW STATUS LIKE "Ssl%";
 6 rows in set (0.0062 sec)
 ```
 
-The connection is secure when the value of `Ssl_cipher` is not empty.
+A non-empty `Ssl_cipher` value indicates that your connection is secure.
 
 For the official MySQL client, you can also use the `STATUS` or `\s` statement to view the connection status:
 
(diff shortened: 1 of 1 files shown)


dveeden:
@Oreoxmt @qiancai I think things like "Check whether the current connection uses encryption" might need to be changed to "Check whether the current connection is secure" or "Check whether the current connection uses TLS".

This is because TLS does more than just encryption:
- Authentication of the server
- Authentication of the client (optional, aka mTLS)
- Encryption of the connection
- Integrity protection
- Negotiation of algorithms

However, a UNIX socket connection is also considered secure, so either we should include that or make this specific to TLS or say something like "secure network connection" (UNIX sockets are not network connections).

What do you think?

qiancai:
> @Oreoxmt @qiancai I think things like "Check whether the current connection uses encryption" might need to be changed to "Check whether the current connection is secure" or "Check whether the current connection uses TLS".
> 
> This is because TLS does more than just encryption:
> 
> * Authentication of the server
> * Authentication of the client (optional, aka mTLS)
> * Encryption of the connection
> * Integrity protection
> * Negotiation of algorithms
> 
> However, a UNIX socket connection is also considered secure, so either we should include that or make this specific to TLS or say something like "secure network connection" (UNIX sockets are not network connections).
> 
> What do you think?

Hi @dveeden, thanks a lot for the suggestion. Yes, “Check whether the current connection uses TLS” is more accurate in the context of `SHOW STATUS LIKE "%Ssl%";`.

enable-tls-between-clients-and-servers.md
```diff
@@ -121,6 +121,8 @@ SHOW STATUS LIKE "Ssl%";
 6 rows in set (0.0062 sec)
 ```
 
+A non-empty `Ssl_cipher` value indicates that your connection is secure.
```
Oreoxmt:
```suggestion
If the `Ssl_cipher` value is not empty, the connection is encrypted.
```

Oreoxmt (approved):
Rest LGTM

hfxsd (approved):

PR #22562, opened 2026-03-13
tidb-performance-tuning-config: remove pessimistic lock memory limit recommendations

dbsid:
## What's changed
- remove `in-memory-peer-size-limit` and `in-memory-instance-size-limit` from the TiKV config example in `tidb-performance-tuning-config.md`
- remove the corresponding recommendation row from the configuration table

## Why
These settings were originally introduced as a targeted workaround for an issue where oversized Raft entries could be generated during leader transfer. If the configuration values were too large, accumulated in-memory pessimistic lock entries might exceed the `raft-entry-max-size` limit when the Region leader was transferred. In this case, the lock entries could fail to be persisted to disk through Raft, which might result in the loss of pessimistic locks.

Because these settings were designed to mitigate this specific scenario rather than to improve general system performance, they should not be considered or documented as general performance tuning recommendations.

## Testing
- docs only

### DIFF
--- a/tidb-performance-tuning-config.md
+++ b/tidb-performance-tuning-config.md
@@ -91,10 +91,6 @@ concurrent-send-snap-limit = 64
 concurrent-recv-snap-limit = 64
 snap-io-max-bytes-per-sec = "400MiB"
 
-[pessimistic-txn]
-in-memory-peer-size-limit = "32MiB"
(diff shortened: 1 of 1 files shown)


mayjiang0203 (approved):
/LGTM

PR #23165, opened 2026-06-29
fix: remove merge conflict artifact in clinic-user-guide-for-tiup.md

yahonda:
Remove a stray line with conflict markers in clinic-user-guide-for-tiup.md that was left behind from a merge conflict resolution.


qiancai:
@yahonda It looks like the English file has the stray line too, so the stray line removed from PR might not be caused by merge conflict resolution.

https://github.com/pingcap/docs/blob/master/clinic/clinic-user-guide-for-tiup.md?plain=1

image

yahonda:
Thanks for the review. and my bad. Closing this is not an merge conflict.

PR #23551, opened 2026-08-18
i18n(ja): remove stray Japanese quote marks around English UI labels

yahonda:
## What is changed

Corpus-wide sweep for a defect pattern first identified and scoped in PR #23519: a bolded English UI label wrapped in Japanese kagi-kakko quotes (`**「Label」**`) even though EN's own doc has no such quote decoration at that position (plain `**Label**`). PR #23519 only fixed the sites it directly touched; this covers the remaining 52 files found via a fresh corpus-wide scan (94 instances).

Each instance was verified against the `release-8.5` (English) source before fixing:

```
**「OK」**  →  **OK**
**「Amazon S3」**  →  **Amazon S3**
**「MySQL Connections」**  →  **MySQL Connections**
```

Quotes wrapping genuinely Japanese-translated UI references (e.g. `「接続」`, `「データソース」`) were left untouched — that quoting is normal and unrelated to this defect class.

Also fixed a handful of adjacent defects found while touching these lines:
- particles (を/の) trapped inside a bold span's closing `**` instead of sitting outside it
- two straddling-bracket cases where an opening `「` landed outside the `**` span while its closing `」` landed inside

## Verification

- Bracket/`**`-parity check across the full diff: 0 issues.
- Every fix confirmed against the EN source at the exact sentence position before applying.


qiancai (approved):

yahonda:
Let me resolve the conflict tomorrow.

qiancai (approved):

PR #23848, opened 2026-09-14
i18n(ja): unify TiKV/PD region-split notation

yahonda:

Unifies the notation for the TiKV/PD "Region Split" concept (splitting a
Region into two, including the Load Base Split feature) across the Japanese
docs. It was previously rendered inconsistently:

- `分割` — dominant, ~50+ sites
- `スプリット` — only in `configure-load-base-split.md`'s own H1/body
  and a couple of cross-references

Region Merge is already overwhelmingly rendered as `マージ`
(リージョンマージ 49 vs リージョン結合 11 sites). A symmetric split/merge
pair reads more consistently than mixing two different renderings for two
complementary operations, so this unifies split to `スプリット` to match
merge, rather than the reverse.

Alternative renderings for merge were considered and rejected before
settling on this direction:
- `結合` already means SQL JOIN throughout this corpus (100s of occurrences:
  テーブル結合, 内部結合, 外部結合, ハッシュ結合, etc.) — reusing it for
  "merge" would collide with an unrelated, deeply established meaning.
- `統合` already means third-party "integration" (738 occurrences, mostly
  monitoring-integration docs) — same collision problem.

Since merge is best left as `マージ`, split is unified to match it.

**Scope discipline**: every site was individually checked against its
context to distinguish the TiKV/PD Region Split feature from unrelated
uses of 分割/"split" in the same files, which were intentionally left
unchanged:
- SQL `PARTITION` splitting/partitioning (a separate, already-settled term
  decision from a previous PR)
- TiFlash column-storage split, SST-file split, RocksDB log split,
  backup-file split (storage-engine-internal splitting, not the Region
  Split scheduling feature)
- TiCDC event/table/transaction splitting (a different mechanism)
- TiDB Lightning/Dumpling data-chunk splitting for import (a different,
  tool-specific mechanism)

Also fixed in passing: a stale `TOC.md` link label that didn't match its
own target page's actual title (EN "Partitioned Raft KV") — a separate
naming mismatch unrelated to this split/merge decision.

- [x] i18n-ja-release-8.5 (TiDB Japanese documentation for TiDB 8.5 versions)

- This PR is translated from:
- Other reference link(s):

### AI agent involvement

- [x] The changes in this PR were primarily made by an AI agent on behalf of the PR author.

- [ ] Delete files
- [ ] Change aliases

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yahonda:
Withdrawing this approach — unifying 分割 to スプリット corpus-wide reads worse than expected. Reverting to a narrower fix (see follow-up PR).

PR #23852, opened 2026-09-14
i18n(ja): fix relocation/migration false friend in TOC.md

yahonda:

"Migration Scenarios" in `TOC.md` used a word that means relocating one's
residence or immigrating, not the data-migration sense intended here.

The rest of this corpus already uses the correct term consistently —
e.g. this same file's own "Migration Overview"/"Migration Tools" entries
just above the affected line. This PR unifies the site to match.

(This PR previously also touched `dm/dm-webui-guide.md`, which used the
same word for its own "Migration" section. That page's "Migration"
section turned out to be a literal DM WebUI navigation label — like the
sibling "Task"/"Source"/"Members" labels in the same list, which are
already correctly kept in English — so it needs a different fix (keep
in English, not translate) and is being handled in a separate PR
instead of here.)

- [x] i18n-ja-release-8.5 (TiDB Japanese documentation for TiDB 8.5 versions)

- This PR is translated from:
- Other reference link(s):

### AI agent involvement

- [x] The changes in this PR were primarily made by an AI agent on behalf of the PR author.

- [ ] Delete files
- [ ] Change aliases

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yahonda:
Folded into pingcap/docs#23851, which already covers other TOC.md-only fixes.

